PR #685, opened 2015-01-10
Implement GUIWorkspace.dispose

### LINKED ISSUE #684: Add way to dispose of GUIWorkspace
Currently, there's no way to cleanly dispose of a `GUIWorkspace`. Other `Workspace`s have a [`dispose()`](https://github.com/NetLogo/NetLogo/blob/59e6bf6965844662db6b8a33faea31290ba87757/src/main/org/nlogo/workspace/AbstractWorkspace.java#L138) method that does it, but `GUIWorkspace` doesn't support it. In addition to calling `super.dispose()` (to close the job thread and extensions), `GUIWorkspace.dispose()` should also

- Kill its [`LifeGuard` thread](https://github.com/NetLogo/NetLogo/blob/59e6bf6965844662db6b8a33faea31290ba87757/src/main/org/nlogo/window/GUIWorkspace.java#L100)
- Stop its [`PeriodicUpdater`](https://github.com/NetLogo/NetLogo/blob/59e6bf6965844662db6b8a33faea31290ba87757/src/main/org/nlogo/window/GUIWorkspace.java#L85)
- Stop its [repaint `Timer`](https://github.com/NetLogo/NetLogo/blob/59e6bf6965844662db6b8a33faea31290ba87757/src/main/org/nlogo/window/GUIWorkspace.java#L98)

There may be other things, but this is what I've managed to discover so far.

qiemem:
Fixes #684 

[Review on Reviewable](https://reviewable.io/reviews/netlogo/netlogo/685)

### DIFF
--- a/src/main/org/nlogo/window/GUIWorkspace.java
+++ b/src/main/org/nlogo/window/GUIWorkspace.java
@@ -63,6 +63,10 @@ public abstract strictfp class GUIWorkspace // can't be both abstract and strict
   // for grid snap
   private boolean snapOn = false;
 
+  private PeriodicUpdater periodicUpdater;
+  private javax.swing.Timer repaintTimer;
+  private Lifeguard lifeguard;
+
   public GUIWorkspace(final org.nlogo.agent.World world,
                       KioskLevel kioskLevel, java.awt.Frame frame,
                       java.awt.Component linkParent,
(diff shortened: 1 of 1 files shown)


SethTisue:
have you verified that this results in GUIWorkspaces actually being garbage collected?

frankduncan:
Yeah, at least running 1500 models works now whereas before it would crash after 200 or so, according to Bryan's tests.

qiemem:
> have you verified that this results in GUIWorkspaces actually being garbage collected?

Yes. I also (just to super duper sure) opened and closed models a bunch, taking heap dumps along the way to further verify. I confirmed that, in fact, there was only 1 instance of `LiteWorkspace` and all the `GUIWorkspace` anonymous inner classes after a gc.

> 1500 models works

Over 5000 thank you very ;)

PR #690, opened 2015-01-20
Fix memory leak in code generation process

mrerrormessage:
When compiling code, the instruction generator is getting attached to the instructions that it generates. These instructions live for the duration of the model, and thus the instruction generator also lives for the duration of the model (even though it isn't actually needed). Attached are some heapdumps that I ran after opening wolf-sheep headlessly 100 times in NetLogo. Note that a corresponding (but smaller) savings of perhaps 5% of total memory can be expected for NetLogo users who do not open multiple models at once.

Before:
before1 12 30 58 pm
before2 12 34 46 pm

After:
after2 at 12 44 08 pm
after2 at 12 42 22 pm

[Review on Reviewable](https://reviewable.io/reviews/netlogo/netlogo/690)


SethTisue:
nice catch!

mrerrormessage:
Thanks! @qiemem and I looked at this together and we were able to find a nice reduction in memory use.

PR #692, opened 2015-01-23
Fix #678: hsb primitives should use real world ranges

### LINKED ISSUE #678: `hsb` primitive strangely expects each argument to be between 0 and 255
a user reports:

>I’m not sure if this is a bug in the doc, the code, or my understanding.
>according to this web page
>http://rgb.to/color/358/aqua-blue
>the HSB color 193, 70, 53 is equivalent to the RGB color 40, 113, 134
>as i understand it the hsb primitive should convert the former to the latter. But it reports  [46 38 53].

seems wrong to me too. I tried it in NetLogo 4.0, so I bet it's always been broken and nobody ever noticed.

frankduncan:
@SethTisue Just check over, please.  I'll be adding tests next week, but I need to get this in for a build for Uri tonight.

[Review on Reviewable](https://reviewable.io/reviews/netlogo/netlogo/692)


src/main/org/nlogo/api/VersionHistory.scala
```diff
@@ -195,6 +195,16 @@ object VersionHistory {
     remove3d(version) == "NetLogo 4.2pre3" ||
     remove3d(version) == "NetLogo 4.2pre4"
 
+  def olderThan52(version: String) =
```
SethTisue: This doesn't seem fully correct to me. Suppose NetLogo 5.0.6 and/or NetLogo 5.1.1 were to be released.

src/main/org/nlogo/prim/etc/_approximatehsb.scala
```diff
@@ -0,0 +1,34 @@
...
+    Syntax.reporterSyntax(Array(Syntax.NumberType, Syntax.NumberType, Syntax.NumberType),
+                          Syntax.NumberType)
+  override def report(context: Context): java.lang.Double =
+    report_1(context,
+        argEvalDoubleValue(context, 0),
+        argEvalDoubleValue(context, 1),
+        argEvalDoubleValue(context, 2));
+
+  def report_1(context: Context, h: Double, s: Double, b: Double): java.lang.Double =
+    api.Color.getClosestColorNumberByHSB(h.toFloat, s.toFloat, b.toFloat, 360.0f, 100.0f, 100.0f);
```
SethTisue: why `float`? seems a bit odd

src/main/org/nlogo/prim/etc/_extracthsb.scala
```diff
@@ -0,0 +1,99 @@
...
+}
+
+class _extracthsb extends Reporter with Pure {
+  override def syntax =
+    Syntax.reporterSyntax(Array(Syntax.NumberType | Syntax.ListType),
+                          Syntax.ListType)
+  override def report(context: Context): LogoList =
+    report_1(context, args(0).report(context))
+
+  def report_1(context: Context, obj: Object): LogoList =
```
SethTisue: style nit: `AnyRef`

src/main/org/nlogo/prim/etc/_extracthsb.scala
```diff
@@ -0,0 +1,99 @@
...
+    obj match {
+      case rgbList: LogoList =>
+        try {
```
SethTisue: style nit: `try` doesn't require curly braces

src/main/org/nlogo/prim/etc/_extracthsb.scala
```diff
@@ -0,0 +1,99 @@
...
+          case e: ClassCastException =>
+            throw new org.nlogo.nvm.EngineException(context, this, displayName + " an rgb list must contain only numbers")
+        }
+      case color: java.lang.Double =>
+        extracthsb.fromColor(360.0f, 100.0f, 100.0f, color)
+      case _ => throw new org.nlogo.nvm.ArgumentTypeException(context, this, 1, Syntax.ListType | Syntax.NumberType, obj)
+    }
+
+  def report_2(context: Context, rgbList: LogoList): LogoList =
+    try {
```
SethTisue: ditto on curlies

src/main/org/nlogo/prim/etc/_extracthsb.scala
```diff
@@ -0,0 +1,99 @@
+// (C) Uri Wilensky. https://github.com/NetLogo/NetLogo
+
+package org.nlogo.prim.etc
+
+import org.nlogo.api, api.{ Syntax, LogoListBuilder, LogoList }
+import org.nlogo.nvm.{ Context, Pure, Reporter }
+
+object extracthsb {
```
SethTisue: better moved to api.Color perhaps?

src/main/org/nlogo/prim/etc/_extracthsb.scala
```diff
@@ -0,0 +1,99 @@
...
+    hsbList.toLogoList
+  }
+
+  def fromColor(hMax: Float, sMax: Float, bMax: Float, color: Double): LogoList = {
+    val ccolor = if (color < 0 || color >= 140) api.Color.modulateDouble(color) else color
+    api.Color.getHSBListByARGB(api.Color.getARGBbyPremodulatedColorNumber(ccolor), hMax, sMax, bMax)
+  }
+}
+
+class _extracthsb extends Reporter with Pure {
```
SethTisue: could use a comment saying what it is, why it exists

SethTisue:
LGTM, but I'd like to see the tests too once they exist.

SethTisue:
there's no reason this needs to be merged for you to make a one-off build, is there? can't you just do the build on this branch?

src/main/org/nlogo/api/VersionHistory.scala
frankduncan: 5.2 is 5.1.1, we just named it this because of the book.  Just like 5.1 was 5.0.6, named such because of the new nw extension.  5.2 is effectively 5.0.7, at least from core netlogo's point of view.

src/main/org/nlogo/prim/etc/_approximatehsb.scala
frankduncan: The Color class depends on floats when talking about hsb, and awt.Color deals with floats when converting things to hsb.

frankduncan:
Yeah, I think I will.  Thanks for the feedback.

src/main/org/nlogo/prim/etc/_extracthsb.scala
frankduncan: I didn't have a good grasp of what belongs in Color, and what belongs here.  I just kind of followed whatever _extracthsb.java did.  Is there a good decision procedure for this stuff?

SethTisue: The goal is to keep language-independent engine code from code that is specific to the NetLogo language. If the code would be useful for writing agent-based models in another JVM language besides NetLogo, then it shouldn't be in the `prim` package. The code in a `prim` class should marshal arguments, check argument types, and report errors, but it shouldn't perform the actual job of the primitive (unless it's something that every language has, so e.g. it would be silly to require that `_plus` call separate code to add the numbers; we might as well just inline it). The actual meat should be in `agent` or a package `agent` depends on such as `api`.

PR #699, opened 2015-02-03
Add hooks for a hypothetical extension with some sort of custom logging

TheBizzle:
The title says it all... but this:

The Custom Logging extension needs these changes in order to work in 5.x.  The `signedjars` branch (now just a tag) had these commits in some harebrained form, but they were of less-than-nice code quality, and were somewhat entangled with other logging stuff that I was doing at the time.  Consequently, rather than cherry-picking them over, I just reimplemented them in order to create this PR.  Most of it is just mindless cruft, anyway....

I've tested this with the latest version of the Custom Logging extension (which you can download [here](https://github.com/NetLogo/Custom-Logging-Extension/releases/tag/1.0)), and everything appears to be behaving as expected.

[Review on Reviewable](https://reviewable.io/reviews/netlogo/netlogo/699)

### DIFF
--- a/src/main/org/nlogo/api/Context.scala
+++ b/src/main/org/nlogo/api/Context.scala
@@ -50,4 +50,18 @@ trait Context {
    */
   def getRNG: org.nlogo.util.MersenneTwisterFast
 
+  /**
+   * Logs a customized message (primarily for use by the Custom Logging extension)
+   *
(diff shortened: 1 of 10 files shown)


TheBizzle:
I've assigned it to the 5.2 milestone, since @cbradyatinquire has bugged me about this on various people's behalves.

Issue #612, opened 2014-05-15
rename “agent monitor” to “inspector” or “agent inspector”?


SethTisue:
to match the `inspect` primitive, and to prevent confusion with regular monitors

SethTisue:
@mrerrormessage curious if this change might actually happen (I saw you referring to "agent inspectors" on another ticket just now)

mrerrormessage:
@SethTisue there are no plans to officially or formally change the name, I just tend to use "inspector" as a shorthand since (as this ticket alludes to) "monitor" is a bit confusing.

PR #718, opened 2015-02-26
Add stop-inspecting to the language.

### LINKED ISSUE #354: Allow programmatic closing of agent monitors
Uri writes:

> We have programmatic control of opening inspectors, but there is no way to close them programmaticaly. It'd be nice if there was.

That would be the complement to [`inspect`](http://ccl.northwestern.edu/netlogo/docs/dictionary.html#inspect). Not sure what a good name would be. `stop-inspecting`? `close-inspector`? `close-monitor`?

### LINKED ISSUE #716: Add primitive for closing inspection windows
Various CCLers (including Uri) have voiced support for a primitive that can close inspection windows (given that we can programmatically open them with `inspect`).  Here is my proposal for such a primitive:

> `stop-inspecting <agent>`: Closes any inspection window that is open for `<agent>`.  If no inspection window is open for `<agent>`, nothing happens.

I believe that Uri would like this primitive included in 5.2 (though I might be wrong).

frankduncan:
Fixes #354 

Would you mind taking a look @nicolaspayette ?  Just making sure it does what you want with the model referenced in: NetLogo/IABM-Models#4

[Review on Reviewable](https://reviewable.io/reviews/netlogo/netlogo/718)

### DIFF
--- a/docs/dictionary.html
+++ b/docs/dictionary.html
@@ -129,16 +129,17 @@ Go figure! - ST 12/2/04
       "#show-turtle">show-turtle</a> (<a href="#show-turtle">st</a>)
       <a href="#sprout">sprout</a> <a href=
       "#sprout-breeds">sprout-<i>&lt;breeds&gt;</i></a> <a href=
-      "#stamp">stamp</a> <a href="#stamperase">stamp-erase</a> <a href=
-      "#subject">subject</a> <a href=
+      "#stamp">stamp</a> <a href="#stamperase">stamp-erase</a>
+      <a href="#stop-inspecting">stop-inspecting</a>
+      <a href="#subject">subject</a> <a href=
       "#subtract-headings">subtract-headings</a> <a href="#tie">tie</a>
       <a href="#towards">towards</a> <a href="#towardsxy">towardsxy</a>
       <a href="#turtle">turtle</a> <a href="#turtle-set">turtle-set</a>
       <a href="#turtles">turtles</a> <a href="#turtles-at">turtles-at</a>
       <a href="#turtles-here">turtles-here</a> <a href=
       "#turtles-on">turtles-on</a> <a href="#turtles-own">turtles-own</a>
-      <a href="#untie">untie</a> <a href="#uphill">uphill</a> <a href=
-      "#uphill4">uphill4</a>
+      <a href="#untie">untie</a>
+      <a href="#uphill">uphill</a> <a href="#uphill4">uphill4</a>
       <!-- ======================================== -->
     <h3>
       <a name="patchgroup" id="patchgroup">Patch-related</a>
@@ -166,8 +167,9 @@ Go figure! - ST 12/2/04
       "#random-pxcor">random-pxcor</a> <a href=
       "#random-pycor">random-pycor</a> <a href="#self">self</a> <a href=
       "#sprout">sprout</a> <a href=
-      "#sprout-breeds">sprout-<i>&lt;breeds&gt;</i></a> <a href=
-      "#subject">subject</a> <a href="#turtles-here">turtles-here</a>
+      "#sprout-breeds">sprout-<i>&lt;breeds&gt;</i></a>
+      <a href="#stop-inspecting">stop-inspecting</a>
+      <a href="#subject">subject</a> <a href="#turtles-here">turtles-here</a>
       <!-- ======================================== -->
     <h3>
       <a name="agentsetgroup" id="agentsetgroup">Agentset</a>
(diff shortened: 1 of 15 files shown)


docs/dictionary.html
```diff
@@ -7939,6 +7941,25 @@
...
+        stop-inspecting <i>agent</i>
+      </h4>
+      <p>
+        Closes the agent monitor for the given agent (turtle or patch).
+        In the case that no agent monitor is open, stop-inspecting does
+        nothing.
+      <pre>
+stop-inspecting patch 2 4
+;; the agent monitor for that patch closes
+stop-inspecting one-of sheep
```
nicolaspayette: Not sure I have a better suggestion from the top of my head, but that's a very unlikely use case...

src/main/org/nlogo/app/AgentMonitorManager.scala
```diff
@@ -153,6 +153,17 @@ with org.nlogo.window.Event.LinkParent
         new Runnable() { def run() { frame.requestFocus() }})
   }
 
+  def stopInspecting(agent: Agent) {
+    if(agent != null) {
```
nicolaspayette: Or `Option(agent).foreach`?  ;-)

frankduncan: I actually had that thought when I wrote it, but I was just mirroring the code above

docs/dictionary.html
frankduncan: It's actually just kind of a goofy line so that it mirrors the doc for inspect.  Made me giggle

src/main/org/nlogo/app/AgentMonitorManager.scala
nicolaspayette: Ah, I hadn't realized. Consistency may be better in that case.

nicolaspayette:
Update https://github.com/NetLogo/NetLogo/blob/stop-inspecting/src/test/org/nlogo/compiler/TestAllSyntaxes.scala ?

docs/dictionary.html
nicolaspayette: It _is_ funny. :-) A suggestion, though: `ask turtles [ stop-inspecting self ]`

frankduncan: Ooo, that's a stellar idea

nicolaspayette:
This is a minor issue, but the current proposal leaves a small hole: there is no way to uninspect a dead turtles. (And turtles dying while they're inspected is not uncommon. Think Wolf-Sheep.)

frankduncan:
I actually ran into that.  The problem is that the agent that gets passed is Nobody.  turtle <dead-id> returns nobody, not a dead turtle.  In fact, if you hold it in a variable and use stop-inspecting, it will work like you'd like.

SethTisue:
I think the workaround Frank suggests is a bug. Any user-visible way of distinguishing an api.Nobody from an agent.Turtle with an id of -1 is a bug and the loophole should be closed. In NetLogo semantics a dead turtle no longer exists; end of story; no exceptions should ever be allowed to leak through.

SethTisue:
Oops, our comments crossed. Anyway, so there is no loophole, but one should not be created.

UriCCL:
good point.
this does happen a fair bit.

at the minimum, stop-inspecting should not error when given a dead agent.

is there something cleverer we can do?

—Uri

Uri Wilensky

frankduncan:
I can definitely make it stop erroring on Nobody.  maybe we should update inspect windows to close when their watched agent dies?

SethTisue:
fwiw, it was an intentional design choice that they don't close, with consensus at devel meeting, and a fair amount of Esther's effort went into making it work correctly. This shouldn't be reversed lightly.

iirc the main design consideration was that you might want to see the final state of the turtle to understand when, how, and why it died.

frankduncan:
Ah, makes sense.  Should we add a programmitic way to close all open inspectors, then?  Or just leave dead ones hanging around?

nicolaspayette:
> Should we add a programmitic way to close all open inspectors

Not sure it's the best possible design, but I'd certainly use that in Traffic Grid Goal if it was available.

frankduncan:
Making it so "stop-inspecting" on nobody actually broke netlogo in weird ways (just catching that exception and ignoring).  But then I thought about it, and I came to the personal conclusion that we fail on nobody in other commands, so this should be in line with them.

However, uninspect-all has been added.

frankduncan:
Talked to Uri, the following is the resolution:

stop-inspecting <agent>
stop-inspecting-dead-agents (to clean up dead inspectors)

I'm going to go ahead and add a menu item for that second one as well.

frankduncan:
Any interested parties should take a look at this.  I'll merge later this week if there's no objections.

nicolaspayette:
Pushed two very minor doc updates. Aside from that, LGTM.

frankduncan:
Maybe a change to "Close Monitors for Dead Agents"?

nicolaspayette:
> Maybe a change to "Close Monitors for Dead Agents"?

Sounds good to me! And then I'd rename "Close All Agent Monitors" to "Close Monitors for All Agents", for consistency.

frankduncan:
Actually not, since close all agent monitors also closes ones that don't have an agent assigned to them yet.

nicolaspayette:
> that don't have an agent assigned to them yet.

I hadn't realized this was even possible! So yeah, good point!

And once they're renamed to inspectors (#612), it will become:

- Close All Inspectors
- Close Inspectors for Dead Agents

...which doesn't look too bad.

PR #719, opened 2015-02-26
Fix Depend via separating global events

frankduncan:
@SethTisue would you mind reviewing this (more philosophically than technically) to see if I made a major netlogo faux pas?

[Review on Reviewable](https://reviewable.io/reviews/netlogo/netlogo/719)


SethTisue:
Will do when I return. But see my comments on netlogo-devel. I am suspicious of this change. It increases coupling, and (worse) expands future opportunities for increased coupling.

frankduncan:
Do you think a callback would be better.  If the editor has a jumpToDefinition callback, it can enable the menu and call that upon clicking.  That callback raises the event.

I'm going to run into this again with code completion, as I need the running app to tell me all the tokens it knows about, and I'll probably have to handle in the same way.

SethTisue:
Yeah I was going to suggest a callback. (More next week.)

frankduncan:
To be honest, a callback would be better for my eventual plan of being able to run multiple Apps in one jvm, but I look forward to hearing your thoughts in general.

frankduncan:
Superceded by #723

PR #724, opened 2015-03-03
Fix #721.  Add netlogo-web? primitive

### LINKED ISSUE #721: Add `netlogo-web?` primitive
It seems likely that there won't be another release of JVM NetLogo before NetLogo Web sees its public release.  Consequently, Uri and I would like to see a `netlogo-web?` primitive introduced in 5.2, so that people can use it in much the same way that the existing `netlogo-applet?` primitive is used.

As for the behavior of this primitive, it is very simple.  As one would expect, `netlogo-web?` always returns `false` in JVM NetLogo.

frankduncan:
@TheBizzle Just a sanity check that this is what you want, please.

[Review on Reviewable](https://reviewable.io/reviews/netlogo/netlogo/724)

### DIFF
--- a/docs/dictionary.html
+++ b/docs/dictionary.html
@@ -536,6 +536,7 @@ Go figure! - ST 12/2/04
     <p>
       <a href="#netlogo-applet">netlogo-applet?</a> <a href=
       "#netlogo-version">netlogo-version</a>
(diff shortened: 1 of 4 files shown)


TheBizzle:
LGTM

PR #733, opened 2015-03-19
Fix #657.  Add behaviorspace-experiment-name prim.

### LINKED ISSUE #657: Add support for a behaviorspace-experiment-name primitive
From @alan-isaac:

> Wouldn't it be nice for NetLogo to provide access to the name of the currently running BehaviorSpace experiment (not just the run number)?

This doesn't look very difficult to implement.  `AbstractWorkspace._behaviorSpaceRunNumber` current holds the run number.  I gets initialized by `org.nlogo.lab.gui.Supervisor`, which has a `protocol` member that happens to hold the name of the BehaviorSpace run.

One solution would just be to add another variable to `AbstractWorkspace` and store the name in there.  Another might be to create some small container class BehaviorSpace settings that lives outside of the `org.nlogo.lab` package, and then store one of those in `AbstractWorkspace` for the BehaviorSpace primitives to grab their values out of.

frankduncan:
Sanity check for obvious errors from @TheBizzle please

[Review on Reviewable](https://reviewable.io/reviews/netlogo/netlogo/733)


TheBizzle:
LGTM

PR #698, opened 2015-02-02
upgrade modelruns branch to Scala 2.11 and sbt 0.13

SethTisue:
I haven't done any real QA on this, I just made sure `sbt fast:test` hasn't regressed and that `sbt run` launches the app and you can run a model.

But I think this should be most — perhaps all, knock on wood! — of what is needed.

Note that in the build, I didn't take advantage of any sbt 0.13 features or clean up the various deprecation warnings. You might look at the NetLogo-Headless build, which is much more 0.13-tastic, to see what I did there.

[Review on Reviewable](https://reviewable.io/reviews/netlogo/netlogo/698)

PR #734, opened 2015-03-19
upgrade to Scala 2.11 and sbt 0.13

frankduncan:
Recreating pull #698 due to branch name change


SethTisue:
@mrerrormessage this should probably just be closed?

mrerrormessage:
Good call @SethTisue. hexy has a very similar changeset, and future work on modelruns should be focused on porting those changes to hexy.

PR #743, opened 2015-03-26
Convert EditorAreaErrorLabel to Scala

Idloj:
[Review on Reviewable](https://reviewable.io/reviews/netlogo/netlogo/743)


frankduncan:
Sweet, thanks!  I think the case _ will get caught by scalastyle, which indicates to me that we should add scalastyle back in.  Though I'm not sure I mind case _

SethTisue:
I don't see anything wrong with `case _`

frankduncan:
Oh, whoops, brain fart.  it complains if you case _ in a catch block.

PR #747, opened 2015-03-27
Model Library dialog tweaks following 5.2.0RC4

nicolaspayette:
close https://github.com/NetLogo/models/issues/32

@frankduncan, that seems like a fairly trivial change, but would you like to give it a quick look before merging nonetheless? Thanks!

[Review on Reviewable](https://reviewable.io/reviews/netlogo/netlogo/747)

### DIFF
--- a/src/main/org/nlogo/workspace/ModelsLibrary.java
+++ b/src/main/org/nlogo/workspace/ModelsLibrary.java
@@ -249,12 +249,12 @@ public strictfp class ModelsLibrary {
         {
           orderednames = new String[]{
               "3D", "SAMPLE MODELS", "CURRICULAR MODELS", "CODE EXAMPLES",
-              "HUBNET ACTIVITIES"};
+              "HUBNET ACTIVITIES", "IABM TEXTBOOK", "ALTERNATIVE VISUALIZATIONS"};
         }
       } else {
         orderednames = new String[]{
             "SAMPLE MODELS", "CURRICULAR MODELS", "CODE EXAMPLES",
-            "HUBNET ACTIVITIES"};
+            "HUBNET ACTIVITIES", "IABM TEXTBOOK", "ALTERNATIVE VISUALIZATIONS"};
       }
     } else {
       orderednames = new String[0];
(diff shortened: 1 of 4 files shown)


nicolaspayette:
@frankduncan can you hold off a bit before merging? I'm going to try to fix https://github.com/NetLogo/models/issues/33 here as well.

nicolaspayette:
@frankduncan: ready for you to review and merge when you can.

nicolaspayette:
This should now close NetLogo/models#37 as well.

src/main/org/nlogo/workspace/ModelsLibrary.java
```diff
@@ -200,14 +196,24 @@ private static void scanDirectory(java.io.File directory, Node grandparent, Node
...
       }
     }
+    // Only add nodes when they have at least one model or folder
+    // under them, and they're not the root `models` folder.
+    // See: https://github.com/NetLogo/models/issues/33
+    // (this would still allow a folder with just an empty folder under it, but
+    // I think we can be careful enough to avoid that situation) NP 2015-03-27.
+    if (hasChild && grandparent != null) {
+      grandparent.add(parent);
+    }
```
frankduncan: So, in the case of multiple empty subdirectories, this code will continue to add top level directories, I think.  not that that's a huge problem, but maybe something that comments about that?  Or add an issue in tandem with merging this?  Or maybe I'm misreading it.

nicolaspayette: In the example you provided, everything would get added, so that's not a problem. The only potential problem is that if we have something like:
```
* Empty Folder A
  * Empty Folder B
```

`Empty Folder A` would still get added, because the code doesn't know yet that `Empty Folder B` is empty. `Empty Folder B` would not get added. I don't think that's a situation we're likely to face in real life.

frankduncan: I thought this change was for directories that had only non nlogo files in them.  So in my example, only resources would get added, right?

nicolaspayette: Oh, yes, you're completely right! And that's a more likely situation (though it doesn't occur at the moment). So, yeah, maybe open an issue? You can even assign me. I'd love to rewrite the whole thing in functional style Scala and just recursively prune empty folders once the whole tree is constructed...

nicolaspayette:
#747 has landed...

Issue #726, opened 2015-03-05
Return to "Open" dialog after "No Such File or Directory" error pop-up?


nicolaspayette:
One more in the "Nicolas reports small annoyances" series:

When opening models, I often type path names directly in the dialog box instead of navigating the folder structure with the mouse. But if I make a small typo in the path, I get the error pop-up and the "Open" dialog is closed, causing me to loose what I typed. It would be nice if, in this case, the user was taken back to the Open dialog, with the previous input available for correction.

askadi:
I believe this is relatively simple. @mrerrormessage 
Store the path name entered before check for valid/invalid path.
If invalid when popup closes force an mouse click event on "open/browse" button and populate the field with the path stored. 

If you could assist me with find the code base for this, it would be great. 

Thanks,
Aditya

SethTisue:
does the error popup include a stack trace? if so, that should take you directly to the relevant code

mrerrormessage:
Hi @askadi , thanks for asking about this PR, but we already have an open PR for it on linux - #956 . It might be helpful to keep an eye on that, perhaps you could implement this feature in Windows or Mac?

mrerrormessage:
Now that #956 has been merged, I think this enhancement is only needed for Windows as I don't think Mac allows you to input a raw filename.

askadi:
Alrighty, I was having a nice look at  NetLogo/src/main/org/nlogo/swing/FileDialog.java.
Now, for Linux its fixed. I would look for a solution to fix issue on windows. 

Thanks!

askadi:
@mrerrormessage I would want to clear couple of things.
* As mentioned in the issue you would want previous input to be there for correction, does it mean once the Error popup is closed you would want to open the file again? Which i believe is correct, because user would know that there is mistake and he needs to correct it.
 Or(As implemented in the PR for linux - #956) you don't want the pop up, but File dialog would reopen again ?
Implementing former case involves lot of analysis, which I am looking into, because filepath from FileDialog goes back to FileMenu.java->FileIO.scala and throws exception thru file2String. So far I believe this is correct.

Alternatively,
We can use JOptionPane for error popup with FileDialog.java, when file does not exist and return to FileDialog once popup is closed with the file path previously entered (perhaps using override for setFileName in JFileChooser ).

Please let me know your thoughts. Thanks!

mrerrormessage:
[...]

PR #759, opened 2015-04-16
remember last used directory, fix #720 on Linux

### LINKED ISSUE #720: Should file save and open dialogs re-use last directory?
(On Linux, at least:) **File / Open**, **File / Save** and **File / Save As...** always start from the user's home directory. This is inconvenient when, e.g., successively opening multiple models from a deeply buried folder.

Perhaps we could use the Java preference store (like we did for **Recent Files**) to save the last used directory?

nicolaspayette:
It appears that the problem (#720) was Linux specific, and it was easy enough to fix.

(@frankduncan: I made the change from the `5.x` branch, but it should be easily mergeable in `6.x`.)

One little caveat: I applied the same fix to the `if (!MAC && directoriesOnly)` code branch. As far as I can tell, this only applies to the `user-directory` primitive on Windows. I'd be grateful if someone with a Windows machine could test that. Short of that, it wouldn't be a big deal to remove this line.

That being said, the whole thing is a mess, and uses a bunch of different dialog classes:

OS | Choose what? | Class
---|---|---
Linux | file | `javax.swing.JFileChooser`
Linux | dir | `javax.swing.JFileChooser`
Win | file | `java.awt.FileDialog`
Win | dir | `javax.swing.JFileChooser`
Mac | file | `java.awt.FileDialog`
Mac | dir | `net.roydesign.ui.FolderDialog`

As @fstonedahl says in an old comment:
```
// I suspect my Linux-specific code could replace all the Mac/Windows code below,
// but that code already seems strangely more complicated than expected, so I decided
// not to mess with it.  Probably it's just left-over cruft from a pre-JFileChooser era,
// but I'm not sure.
```

I suspect he's right. I'd even go as far as replacing the whole `FileDialog.java` class for `6.x` by what would end up being only a few lines of Scala code using `JFileChooser`. It would help with #726, get rid of the `net.roydesign.ui` dependency, and possibly other things.

But maybe we shouldn't mess with it. Thoughts? Historical considerations?

### DIFF
--- a/src/main/org/nlogo/swing/FileDialog.java
+++ b/src/main/org/nlogo/swing/FileDialog.java
@@ -98,6 +98,7 @@ public final strictfp class FileDialog {
       if (result != javax.swing.JFileChooser.APPROVE_OPTION) {
         throw new org.nlogo.awt.UserCancelException();
       }
+      currentDirectory = selectedDirectory(chooser);
       return chooser.getSelectedFile().getAbsolutePath();
     }
     if (!MAC && directoriesOnly) {
@@ -109,6 +110,7 @@ public final strictfp class FileDialog {
           != javax.swing.JFileChooser.APPROVE_OPTION) {
         throw new org.nlogo.awt.UserCancelException();
       }
+      currentDirectory = selectedDirectory(chooser);
       return chooser.getSelectedFile().getAbsolutePath();
     }
     java.awt.FileDialog dialog;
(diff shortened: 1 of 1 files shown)


SethTisue:
I've always been prejudiced against JFileChooser since it doesn't have true native look and feel, but the appalling chart Nicolas provides above has me thinking maybe that preference should yield to internal consistency. I don't really know though, it's been a long time since I looked at this stuff.

fstonedahl:
Sorry, I can't add much more to the discussion than what was originally
posted in my comment.

My recollection is that at the time I was about the only CCLer running
NetLogo on Linux... there was some Linux-specific bug that was annoying, so
I went in and fixed it for Linux, but didn't want to break anything that
was currently working on Mac/Windows (especially since I didn't have a Mac
to test it on...)

Cheers,

Idloj:
Why is it still open?

Issue #762, opened 2015-04-17
TestCompileAll should test compiling custom preview commands


nicolaspayette:
(see #760)

nicolaspayette:
I suspect that models with `need-to-manually-make-preview-for-this-model` won't compile. We'll need to handle that somehow, probably by just skipping them.

SethTisue:
there's already two occurrences of `previewCommands.containsSlice("need-to-manually-make-preview-for-this-model")` in the codebase. so elegant.

nicolaspayette:
> there's already two occurrences of `previewCommands.containsSlice("need-to-manually-make-preview-for-this-model")` in the codebase. so elegant.

Still far from ideal, but https://github.com/NetLogo/NetLogo/commit/6f7d6a957ed117e00e3ab4c762b2469e7bbed969 should help a little. :-)

nicolaspayette:
This was resolved by merging  #766.

PR #766, opened 2015-04-21
TestCompileAll improvements

nicolaspayette:
@frankduncan, quick review?

### DIFF
--- a/src/main/org/nlogo/headless/ChecksumsAndPreviews.scala
+++ b/src/main/org/nlogo/headless/ChecksumsAndPreviews.scala
@@ -41,6 +41,8 @@ object ChecksumsAndPreviews {
   }
 
   object Previews {
+    def needsManualPreview(previewCommands: String) =
+      previewCommands contains "need-to-manually-make-preview-for-this-model"
     def okPath(path: String) =
       List("HUBNET", "/GOGO/", "/CODE EXAMPLES/SOUND/")
         .forall(!path.toUpperCase.containsSlice(_))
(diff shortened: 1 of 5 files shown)


src/main/org/nlogo/headless/ChecksumsAndPreviews.scala
```diff
@@ -41,6 +41,8 @@ object ChecksumsAndPreviews {
   }
 
   object Previews {
+    def needsManualPreview(previewCommands: String) =
+      previewCommands contains "need-to-manually-make-preview-for-this-model"
```
SethTisue: surely this is always false? should be `containsSlice`, not `contains`

two cheers for Scala's non-typesafe `contains` method

nicolaspayette:
```
scala> def needsManualPreview(previewCommands: String) =
     | previewCommands contains "need-to-manually-make-preview-for-this-model"
needsManualPreview: (previewCommands: String)Boolean

scala> needsManualPreview("need-to-manually-make-preview-for-this-model")
res1: Boolean = true
```
I think it is `java.lang.String.contains` being called here, not `scala.collection.SeqLike.contains`. `contains` reads clearer to me than `containsSlice`, and is probably a bit faster. Would there be any reason to be using `containsSlice` anyway? (A bit clearer, perhaps?)

SethTisue: hah. you are right, of course.

yes, I would argue that `containsSlice` is clearer, to Scala eyes — it doesn't lead the careful reader who knows Scala well down the garden path I went down. but then I _would_ say that, to make myself look less foolish for going down that garden path :-)

nicolaspayette: This code will hopefully be short-lived anyway: I have something else in store for `6.x` that should look much better to Scala eyes... :-)

nicolaspayette:
This, by the way, should close #762.

Issue #765, opened 2015-04-21
Model library scanner probably shouldn't follow symlinks


nicolaspayette:
For example, creating a symlink going up one folder:
```
nl/models/Sample Models$ ln -s .. _parent
```

results in:

![screenshot from 2015-04-20 20 37 14](https://cloud.githubusercontent.com/assets/908754/7243671/0b03682c-e79e-11e4-8d6a-2ce511962aa9.png)

It bottoms out and (silently) stops at some point; I'm not sure why it doesn't trigger a stack overflow.

That's not currently a problem in the library because this situation doesn't occur.

It _does_, however occur in the NW extension repo, where such a symlink was recently introduced to allow testing from the main repo (https://github.com/NetLogo/NW-Extension/commit/d0b42e68ff1ae56be73bb928f987bbed3d53c501). I found that out while working on `TestCompileAll` improvements (for #762), which uses `ModelsLibrary.getModelPathsAtRoot("extensions")` to test if models in extensions repo compile. The test currently doesn't crash, but ends up testing the same models many many times because of that (taking much longer than needed). I have a simple workaround for `TestCompileAll`, but we should write a proper fix at some point (possibly while taking care of https://github.com/NetLogo/NetLogo/issues/754).

nicolaspayette:
I just realized that `getModelPaths` only seems to be used for tests and generating checksums and previews:

```
$ git grep getModelPaths
src/main/org/nlogo/headless/ChecksumsAndPreviews.scala:20:        ModelsLibrary.getModelPaths(true)
src/main/org/nlogo/workspace/ModelsLibrary.java:21:  public static String[] getModelPaths() {
src/main/org/nlogo/workspace/ModelsLibrary.java:22:    return getModelPaths(false);
src/main/org/nlogo/workspace/ModelsLibrary.java:25:  public static String[] getModelPaths(boolean exclusive) {
src/main/org/nlogo/workspace/ModelsLibrary.java:39:  public static String[] getModelPathsAtRoot(String path) {
src/test/org/nlogo/headless/Dump.scala:45:    for(path <- ModelsLibrary.getModelPaths; if include(path))
src/test/org/nlogo/headless/TestCompileAll.scala:12:  for (path <- ModelsLibrary.getModelPaths ++ ModelsLibrary.getModelPathsAtRoot("extensions")) {
src/test/org/nlogo/workspace/TestLibrary.scala:11:    val duplicateNames = getModelPaths
[...]

PR #767, opened 2015-04-27
Add Preview Commands Editor

### LINKED ISSUE #760: Make model preview commands accessible from the GUI?
It would make model library management easier for us, but could also be useful in other places (e.g., Modeling Commons? Community Models?)

nicolaspayette:
Two main changes:

- a model's preview commands are now represented as a data type instead of just a string; this makes code interacting with previews nicer/safer.

- there is now a graphical preview commands editor (to resolve #760); this will make quality control of the previews in the library much easier for me and hopefully encourage model authors to pay attention to their model's preview.

Some notes:

- I added an option to access the preview commands editor from the Tools menu. It could also go elsewhere; I didn't think about that too deeply.

- I put the GUI code in a new `org.nlogo.app.previewcommands` package. @frankduncan, I know you have architecture changes planned for 6.x. Let me know if I should move it somewhere else.

- There are no new unit tests in the PR as this is mostly GUI stuff and I consider that the new `api.PreviewCommands` data type is implicitly tested through `TestCompileAll`; but I'm open if anyone has testing suggestions.

- I manually integrated the changes from #766, with some specific 6.x exclusions. The `GoGo` exclusion will have to be removed when the `gogo` extension is updated for 6.x

- The PR includes an update to the `models` submodule because 6.x was previously pointing to a `models` commit where some preview commands were not compiling.

The Travis build currently fails because of https://github.com/NetLogo/NetLogo/issues/765. I'll handle this here, so hold off merging for now (but code reviews appreciated).


SethTisue:
maybe the preview commands editing should go in the model's Settings dialog?

nicolaspayette:
> maybe the preview commands editing should go in the model's Settings dialog?

Yeah, that was actually my first impulse; and the fact that you thought about it too means we should seriously consider it.

Reasons I didn't do it:

- It's not obvious _where_ a **Preview Commands** button should go in that dialog. It doesn't really fit in any of the three current sections ("World", "View" and "Tick Counter"). I guess we could create a fourth one, but it's an already busy dialog.

- It was really nice, while working on the feature, to have the <kbd>Ctrl</kbd><kbd>Shift</kbd><kbd>p</kbd> shortcut to bring them up. But it probably doesn't matter for the end user. Or maybe we could keep the shortcut and hide the menu item.

SethTisue:
Maybe an "Advanced..." button that brings up a second dialog for rarely used stuff

nicolaspayette:
I tried adding a button to the **Models Settings** dialog (https://github.com/NetLogo/NetLogo/commit/4938fd4b1c836a362ea085537390868a42267990, in a separate branch).This is the result:

screenshot from 2015-04-29 17 58 05

I don't like it.

It feels out of place. I also think that it's an unnecessary level of indirection. It also raises the question of how saving/canceling should behave: if you click **OK** in the **Preview Commands Editor** but then **Cancel** in the **Models Settings**, should changes to the preview commands be kept or not?

The more I think about it, the more I feel like the **Tools** menu is where it belongs.

@frankduncan, over to you for review.

nicolaspayette:
This work was done off the `6.x` (modelruns) branch. I need to rebase it on `hexy`.

nicolaspayette:
Working on the Hexy port in a different branch (https://github.com/NetLogo/NetLogo/pull/886). Closing this one.

PR #774, opened 2015-05-06
add Online Models dialog

nicolaspayette:
(This is emphatically just a prototype. This pull request should not get merged.)

## What is it?

A tool to browse GitHub for NetLogo models and load them directly in NetLogo. The idea stems from a chat with @frankduncan so he gets credit for it (but flaws in execution are mine alone).

Here is a screenshot:

screenshot from 2015-05-05 21 24 52

- The left panel is a tree, just like in the model library dialog. Top nodes are repos, and you have their directory structure under them, showing only folders with NetLogo files.

- The right panel shows the readme if a repository node is selected or the info tab content if a model is selected. We could conceivably generate and add preview images.

To use it, bring it up from the **File** menu (or <kbd>Ctrl</kbd>+<kbd>g</kbd>) enter your GitHub username/password and click **(Re)Load from GitHub**.

It shows every GitHub repo that _you_ have access to, i.e. all public repos + your private repos.

Click **Open** to load the selected model (double-clicking doesn't work yet).

## What's currently wrong with it?

- The code is really messy, mixing up GUI and GitHub api querying stuff.

- There is almost no error handling. If something goes wrong with a request, things just won't load.

- Support files (e.g., data, images, .nls files, etc.) are not downloaded.

- Of course, if a model needs an unbundled extension, it doesn't get downloaded either.

- Nothing is currently cached on disk between sessions (this is bad for speed and api rate limit).

- Everything (i.e., every model file, once the info tab is displayed) stays in memory for the whole session.

- The user currently needs to log into GitHub using username/password (the password is not stored). It's probably possible to make life easier for the user using OAuth tokens or something; I'm not very familiar with this stuff. There is no way around the fact that GitHub requires authentification, though. Anonymous API queries are limited to 60/hour (https://developer.github.com/v3/#rate-limiting) which is not nearly enough.

- Need to find a workaround for https://github.com/NetLogo/NetLogo/commit/d7b654f9e282251db5b1a66f5767a50a5ea9222b.

Whether or not an improved version of this should get integrated into NetLogo is an open question. We could also make it a plugin at first.

That being said, I already had a lot of fun just browsing the tree and looking at people's models!

Comments welcome.

### DIFF
--- a/build.sbt
+++ b/build.sbt
@@ -75,6 +75,8 @@ NativeLibs.nativeLibsTask
 
 moduleConfigurations += ModuleConfiguration("javax.media", JavaNet2Repository)
 
+resolvers += "Typesafe Repo" at "http://repo.typesafe.com/typesafe/releases/"
+
 libraryDependencies ++= Seq(
   "log4j" % "log4j" % "1.2.17",
   "javax.media" % "jmf" % "2.1.1e",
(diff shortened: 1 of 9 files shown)


SethTisue:
whoa, really really cool. it would be really great if something like this happened.

nicolaspayette:
Re-targeting (_very_ tentatively) for `hexy`.

Idloj:
Something like this could be used for extensions.

PR #778, opened 2015-05-08
Stop tracking intellij project files

qiemem:
Since Intellij has had [built-in support for sbt since 2013](http://blog.jetbrains.com/scala/2013/11/18/built-in-sbt-support-in-intellij-idea-13/), there's no reason to include Intellij project files in the repo. I'm not even sure if the project files that are in there even work.

This PR can probably apply to both 5.x and 6.x.


mrerrormessage:
:+1: to removing these files.

SethTisue:
:+1:

SethTisue:
@Idloj points out that NetLogo.iml didn't get removed — presumably just an oversight?

PR #779, opened 2015-05-12
Update AutoConverter1.scala

Idloj:


mrerrormessage:
I would say retarget this to hexy, but it's too small to standalone. Perhaps combine it with the scala conversion PR?

Idloj:
Can this be combined with #781 instead?

mrerrormessage:
Ideally, all three PRs would be combined into one.

PR #781, opened 2015-05-20
Tweak the code a bit

Idloj:
I have used https://github.com/HairyFotr/linter and didn't include it in the `build.sbt` because it had too many false positives.


mrerrormessage:
Needs to be retargeted to hexy, combine with scala conversion PR, if possible.

PR #783, opened 2015-05-21
LevelSpace zooming support

mrerrormessage:
Add the ability to hook into InterfacePanelLite/AppletPanel.

* Make ZoomMenu public
* Decouple AppletPanel constructor to allow alternate InterfacePanelLite
  to be used


mrerrormessage:
Included with the codeblock PR onto hexy.

PR #787, opened 2015-06-02
Allow model saving

mrerrormessage:
This is needed for LevelSpace. Contains https://github.com/NetLogo/NetLogo/pull/783 (I'll close that one if desired).


mrerrormessage:
Included with the codeblocks PR onto Hexy.

PR #790, opened 2015-06-08
Make the Add button more straight-forward

Idloj:


SethTisue:
can you explain a little more what this is about?

Idloj:
Instead of doing a special thing that inherits JToggleButton and using a noneButton and a ButtonGroup, use just a plain JToggleButton

mrerrormessage:
This PR can't be accepted onto 6.x, as 6.x is not taking any new PRs at the moment. It appears that the rationale behind the existing design is listed [here](https://github.com/NetLogo/NetLogo/pull/790/files#diff-18e4f20afd87c873bd1eb39ea797f315L97). While this doesn't require using a custom subclass of JToggleButton (to my knowledge), we need to make a decision on whether the complexity reduction here is sufficient to eliminate that feature. 

In summary, you *can* re-open this PR against hexy, but we'll need to have a pretty thorough discussion around the behavior change and there's no guarantee that we'll merge it.

Idloj:
I saw the comment, but it probably talks about some specific case. This behavior does happen in current NetLogo case because the toggle button is added to a group with the noneButton. You can easily see that the general behavior isn't what described by creating a simple JFrame with a JToggleButton and click on it.

I *did* test what I changed, not just pushed it, and the behavior didn't change.

mrerrormessage:
It does appear to work the same as the existing behavior, but this change again falls under the category of changes that don't accomplish much ([see this comment on another PR for more explanation](https://github.com/NetLogo/NetLogo/pull/859#issuecomment-144445288)). At the moment, I don't think that it's worth the time to verify and test further.

Idloj:
Refactoring is important.

SethTisue:
Refactoring is important in living code. The code you're touching here is, if not actually dead, in a nursing home and only intermittently aware of its surroundings.

Idloj:
By saying this you basically say that only changes that change the behavior worth investing.
This also means that changing the netlogo file format is questionable, because it requires a lot while providing no change in behavior.

mrerrormessage:
Unnecessary complexity is bad when it causes problems. Tolerating poor internal structure of a GUI class which NetLogo uses a *single* instance of is just the price of working in a legacy codebase. Working code (especially code that is peripheral to the system, like GUI code) is left alone until it needs to be changed or extended. There are *zero* automated tests around the behavior of this class (around all GUI behavior, for that matter) which means GUI refactorings are extra-expensive. Until someone needs this class changed or extended; discussing (let alone coding and testing) refactoring on it is just changing code for the sake of changing code. This class hasn't been changed in any meaningful way in the past four years.

If you're interested in refactoring, converting prims to scala would be a great place to start. Additionally, there are a number of java files in the `agent` package most of which are thoroughly tested and would be great choices for refactoring. Those classes are used frequently and improving their internal structure offers value by making NetLogo easier to change in the future.

mrerrormessage:
The reason for changing the NetLogo file format is to allow future extension. The current format is not extensible. At all. There are multiple features of interest which more or less necessitate a change in the NetLogo file format.

Idloj:
OK, got your point.

Idloj:
When converting prims to Scala, should I put a `@strictfp` annotation on them?

PR #803, opened 2015-06-23
Evaluate `runresult` argument only once

### LINKED ISSUE #801: `runresult` evaluates argument twice
Discovered in https://github.com/NetLogo/models/blob/master/Code%20Examples/User%20Interaction%20Example.nlogo, where the following code:
~~~
  let col runresult
            user-one-of "What color should the turtles be?"
                        ["red + 1" "blue - 2" "green + 3"]
~~~

causes the dialog box to pop twice.

@SethTisue says:

> indeed. it is a bug in _runresult.scala
it does args(0).report(context) and then later does argEvalString(context, 0)

nicolaspayette:
This should fix #801, and also close https://github.com/NetLogo/models/issues/74 once merged.

@frankduncan, review?

### DIFF
--- a/src/main/org/nlogo/prim/etc/_runresult.scala
+++ b/src/main/org/nlogo/prim/etc/_runresult.scala
@@ -21,8 +21,7 @@ class _runresult extends Reporter {
           throw new EngineException(context, this,
             token.name + " doesn't accept further inputs if the first is a string")
         try {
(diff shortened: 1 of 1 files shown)


SethTisue:
LGTM. tempted to ask for test coverage, but it probably doesn't make sense to test that every primitive doesn't double-evaluate any of its arguments

PR #805, opened 2015-06-24
Manage original and zoomed font separately in OutputWidget

### LINKED ISSUE #802: Zoomed-in font size should not be saved for output widget
- Open any model with an output box (e.g. **Profiler Example**, where the output box's font size is 12).
- Zoom in a few times. The font size is increased (e.g., to 20).
- Save the model with no other changes. The increased font size is saved.
- Reset your zoom level to "normal".
- Reload the model you just saved. You get the big font, even if zoom level is normal.

Just zooming in or out should not actually _change_ a model.

nicolaspayette:
This is intended to fix #802, as `fontSize` (called by the `save` method) now returns the original font size.

The general logic is lifted from: https://github.com/NetLogo/NetLogo/blob/5.x/src/main/org/nlogo/widget/NoteWidget.scala#L43

Things are simpler here, though, because changing the font size doesn't change the size of the widget.

There is no test coverage, as it is generally absent for widgets and I'm not even sure what it would look like in this case. Suggestions welcome, though.

@frankduncan, review? @SethTisue, I'd be grateful if you could take a quick look as well.

### DIFF
--- a/src/main/org/nlogo/window/OutputWidget.scala
+++ b/src/main/org/nlogo/window/OutputWidget.scala
@@ -21,8 +21,15 @@ class OutputWidget extends SingleErrorWidget with CommandCenterInterface with
   }, java.awt.BorderLayout.CENTER)
 
   def propertySet = Properties.output
-  def fontSize = outputArea.fontSize
-  def fontSize(fontSize:Int) = outputArea.fontSize(fontSize)
+
+  originalFont = outputArea.getFont
+  def fontSize = originalFont.getSize
+  def fontSize_=(newSize: Int): Unit = {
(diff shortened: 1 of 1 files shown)


SethTisue:
My memory of the zooming stuff is growing dim, so I can't really comment on the code. One thing to test is whether it works when you snap back to the original zoom level all at once, instead of always zooming a step at a time. Loading or saving while zoomed is also a typical failure mode.

nicolaspayette:
- [ ] snap back to the original zoom level all at once, instead of always zooming a step at a time.
- [ ] Loading or saving while zoomed

(that's manual testing, though.)

PR #828, opened 2015-07-14
Put READMEs in packages

Idloj:
Just copied from the Package Guide(I didn't make README for "More primitives.")


SethTisue:
These should be Scaladoc, using the package object. For an example of how this is done, see https://github.com/NetLogo/NetLogo/blob/5.x/src/main/org/nlogo/plot/package.scala

If this PR is accepted, https://github.com/NetLogo/NetLogo/wiki/Package-guide should be removed — otherwise we'll have redundant doc that will be sure to get out of sync.

Idloj:
Unfortunately, description in `package.scala` doesn't show up in github.
Maybe `./sbt all` can put the content of the `README.md`s into `package.scala`s so they are available in the Scaladoc?

SethTisue:
I think it's better to just put them in package.scala and be done with it — it's not worth the trouble to do anything more elaborate. There is no standard in the open source world that every directory in the source tree should have a README file.

PR #833, opened 2015-07-16
(WIP) Support 3D rendering in headless mode

nicolaspayette:
I started tinkering on this a couple of weeks ago after a chat with Seth made me realize that the 3D view currently cannot be exported from headless mode.

I would very much like to have that working in order to automatically generate the preview images for 3D models. It would also improve regression testing for NetLogo 3D.

My efforts so far have failed, and then I had to move on to other stuff. I'll give it a another try when/if time permits, but I wanted to at least have the work in progress out there in the meanwhile.

### DIFF
--- a/src/main/org/nlogo/api/RendererInterface.scala
+++ b/src/main/org/nlogo/api/RendererInterface.scala
@@ -2,17 +2,20 @@
 
 package org.nlogo.api
 
-trait RendererInterface {
+trait HeadlessRendererInterface {
   def trailDrawer: TrailDrawerInterface
   def changeTopology(wrapX: Boolean, wrapY: Boolean)
+  def resetCache(patchSize: Double)
+  def exportView(settings: ViewSettings): java.awt.image.BufferedImage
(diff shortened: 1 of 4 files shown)


nicolaspayette:
Old comment made by @frankduncan on gitter (pasting here so I don't forget):

> My guess is nothing came out
> Because 3d export needs a framebuffer to read from
> And that requires the gpu to be initialized via glut
> Which means you have to actually open a 1x1 pixel window
> (I use xfvb on travis to accomplish this on travis)

mrerrormessage:
@nicolaspayette , I should have asked you this question a while back, but do you consider this PR to still be useful and important? It will need some modifications to make it into hexy, but seems doable at this point.

nicolaspayette:
Useful, yes. Important? Not especially. I'd like to resume working on it
one day, but that won't happen for a while. Feel free to close for now.

--
Nicolas
(mobile)

mrerrormessage:
Closing as this is pointing to 5.x

PR #834, opened 2015-07-21
Add link to the GoGo Upgrade wiki page

Idloj:


SethTisue:
review by, I dunno, @cbradyatinquire?

mrerrormessage:
The link just above this (to gogo-rxtx) refers to the version of the gogo extension prior to serial, while the transition guide is on how to go from serial to HID. This addition would be most clear if it appeared at the end of the paragraph ending L62.

Idloj:
Like this?

docs/transition.html
```diff
@@ -61,6 +61,8 @@
       gogo-serial extension.  Change your model to use gogo-serial as opposed to
       gogo.
     <p>
+      More details [here](https://github.com/NetLogo/NetLogo/wiki/GoGo-Upgrade).
```
Idloj: Like this?

mrerrormessage: It doesn't need a separate `<p>` tag. It can go just after the prior sentence ends.

mrerrormessage:
Looks good, thanks for this change!

PR #838, opened 2015-07-24
Add tooltips when widget labels are truncated

Idloj:


frankduncan:
May be more conventional to have setToolTipText take an Option[String].  We generally don't like to use nulls because that can possibly lead to NPEs down the line, and Option takes care of that at compile time.

Idloj:
setToolTipText is a swing api, I must use `null`, that's how it works

frankduncan:
Ah, I see!  Hmmm, would it be better to filter these through a function in Widget, since it already has the displayName?  Maybe not.

Idloj:
The widgets can be categorized into general two groups: a) the widgets which draw themselves alone on theur JPanel b) the widgets which use some sub-components.
The technique for each group is different and then there are also sliders which have SliderPainter and its subclasses.

N.B. once SliderPainter is converted to scala, some logic can be lifted up to it.

PR #844, opened 2015-08-31
Add ability to Save to NetLogo Web

mrerrormessage:
The `net-logo-web.html` file is generated from changes to Galapagos on [this branch](https://github.com/NetLogo/Galapagos/pull/319). This set of changes doesn't do anything to check whether the model is exportable, but upon opening the exported file it displays an informative error message, like this one:
screen shot 2015-08-31 at 10 20 20 am

There's a lot of potential room for enhancement:
* Opening the browser to view the exported model immediately after saving it.
* The model could be pulled from a remote URL if an internet connection was available. Note that the model is modified from a Galapagos export so that we don't need to do a full HTML parse to generate the NLW export, so it couldn't be pulled *directly* from netlogoweb.org, but it could be pulled from ccl.northwestern.edu, for example.
* We could check whether the model uses extensions or unimplemented primitives before exporting.

However, given that this is a minor release, none of those features were requested, and all of them introduce substantial additional complexity, I thought it best to defer them until we knew we wanted them for sure.


frankduncan:
This looks good to me.  I'm going to merge in tomorrow when I take care of fixing other release problems, the make the 5.2.1-M1

PR #856, opened 2015-09-25
Use sbt to build all extensions

mrerrormessage:


SethTisue:
* about project/Testing.scala, I overhauled that stuff in https://github.com/NetLogo/NetLogo-Headless/blob/master/tests.sbt, you might want to base it off that
* while you're upgrading stuff, you might go all the way to sbt 0.13.9

mrerrormessage:
@SethTisue , thanks for the heads-up, I'll definitely integrate the fixes from that branch, I don't think I had notice that testing was overhauled there. I'll also bump to 0.13.9.

PR #859, opened 2015-09-28
Tweak the button widget appearance. close #807 close #808

### LINKED ISSUE #807: Active forever icon should have transparent background, not black

### LINKED ISSUE #808: All icons on the button widget should be inversed, not darkened
Basically, forever icon and agent icon.
I don't see why active forever button is green and not white like the text, because in inactive mode they are the same color.
The agent icons can be hardly seen, and the turtles legs are gone.

Idloj:


mrerrormessage:
Thanks for submitting this. I think this could be a useful addition, but it will require some work. When inactive, buttons look the same in both. When activated, the changes are as follows:

Forever buttons look like this right now:
screen shot 2015-09-30 at 8 05 19 am
screen shot 2015-09-30 at 8 21 20 am
screen shot 2015-09-30 at 8 21 16 am

After this PR, they look like this:
screen shot 2015-09-30 at 8 02 40 am
screen shot 2015-09-30 at 8 05 29 am
screen shot 2015-09-30 at 8 18 55 am

I think that the color scheme changes here need a little work.

* Having a bright green forever arrow, as at present, helps to signal to the user "something is happening here" which is important, as forever buttons are the core of how models work.
* I find the inverted-color turtles look less like turtles than the almost-invisible turtles that are on the button at present. Unlike links and patches, which don't have an inherent color, turtles as displayed on the icon are green and it's confusing to see them in pink-purple.
* The forever link button is definitely clearer.

There are several ways that this could be fixed, the most straightforward being adding separate images for the on and off state of each icon. An alternative method would be to lighten (as opposed to inverting) the color on each. It doesn't matter to me which way you choose to make the change, as long as the turtles end up green. I'm also planning to leave a few comments on the code for style, but it looks fine for the most part.

src/main/org/nlogo/window/ButtonWidget.scala
```diff
@@ -62,8 +61,8 @@ object ButtonWidget {
   // encapsulates what used to be a bunch of 4 way if statements.
   // ButtonWidget now has a single ButtonType object that handles all this logic for it.
   case class ButtonType(name: String, agentClass:Class[_ <: Agent],
-                        img:Option[ImageIcon], darkImg:Option[ImageIcon]){
-    def img(dark:Boolean): Option[ImageIcon] = if(dark) darkImg else img
+                        img:Option[ImageIcon], inverseImg:Option[ImageIcon]){
+    def img(inverse:Boolean): Option[ImageIcon] = if(inverse) inverseImg else img
```
mrerrormessage: Please change `if(` to `if (`.

src/main/org/nlogo/window/ButtonWidget.scala
```diff
@@ -62,8 +61,8 @@ object ButtonWidget {
   // encapsulates what used to be a bunch of 4 way if statements.
   // ButtonWidget now has a single ButtonType object that handles all this logic for it.
   case class ButtonType(name: String, agentClass:Class[_ <: Agent],
-                        img:Option[ImageIcon], darkImg:Option[ImageIcon]){
-    def img(dark:Boolean): Option[ImageIcon] = if(dark) darkImg else img
+                        img:Option[ImageIcon], inverseImg:Option[ImageIcon]){
```
mrerrormessage: Instead of renaming `darkImg` to `inverseImg`, let's rename it to something like `activeImg`. If you would like, you can also rename `img` to `inactiveImg`.

src/main/org/nlogo/window/ButtonWidget.scala
```diff
@@ -374,15 +373,22 @@ class ButtonWidget(random:MersenneTwisterFast) extends JobWidget(random)
     def paintButtonRectangle(g: Graphics) {
       g.setColor(getPaintColor)
       g.fillRect(0, 0, getWidth(), getHeight())
-      def renderImages(g: Graphics, dark: Boolean) {
+      def renderImages(g: Graphics, inverse: Boolean) {
         def maybePaintForeverImage() {
           if (forever) {
-            val image = if (dark) FOREVER_GRAPHIC_DARK else FOREVER_GRAPHIC
+            val image = new ImageIcon(java.awt.Toolkit.getDefaultToolkit.createImage(
```
mrerrormessage: I see using a filter here as a disadvantage. Granted, it allows us to only use one forever icon, but the cost of that is loss of flexibility between enabled/disabled modes as well as the additional complexity of using an `RGBImageFilter` that swaps foreground and background colors.

Idloj: What exactly do you mean by "loss of flexibility between enabled/disabled modes" ?

Idloj:
The reason I changed the green forever icon was to have consistency between active and inactive state. In the inactive state the text and the icon were in the same color, while in the active state they were in completely different colors. As for signaling the user that work is being done, I think the dramatic change in colors in the entire button is sufficient. You know, you can't miss it when the button change from light gray to black.

src/main/org/nlogo/window/ButtonWidget.scala
mrerrormessage: I mean that by swapping out the whole image, we can have a completely different image for on and off. For instance, we could use pause and play images, or on and off images. We aren't limited to a single image that reverses it's colors.

Idloj: I understand now.
Still, in the current state, having 4 images doubled doesn't seem right to me.
The pause and play can still be used instead of one forever icon, but it doesn't mean all other icons should be sort of duplicated.

mrerrormessage: I'm not necessarily saying that all images need to be doubled. I imagine that links and patches probably won't be. However, this code only affects the forever image on a forever button. So the choice is one extra image vs an anonymous RGBImageFilter. The image is both simpler and more flexible.

Idloj: OK, I can change this.

mrerrormessage:
The issue that was being resolved here (as I understand it) is that it was difficult to see the agent icon on forever buttons when active. 

While it's true that it's hard to miss when a button changes from light gray to black, sometimes users first see a model when a forever button is already active and/or they are unfamiliar with the NetLogo user interface. Additionally, a button changing color does not indicate that there is an action actively being performed. Bright colors (green in this case) signal "something is going on here."

A digression: It's certainly fine to entertain the idea that such an icon might not be needed at all (it isn't in NetLogo Web, for instance), but that wasn't the issue originally addressed. While NetLogo could definitely use major work on the user interface, changing a single piece of it in a way that doesn't markedly improve quality is just change for the sake of change. We're interested in actively addressing user issues, extending NetLogo in a way that makes it more powerful, or altering NetLogo to make it easier to use. Changes in PRs that materially advance one or more of these goals will almost always be welcome. Other changes will be looked at as unimportant at best. I would lump the changing of the forever icon color into the "other changes" category, since it doesn't make NetLogo more powerful, address a user issue, or make NetLogo easier to use. Even though it doesn't make NetLogo any harder to use, it just isn't worth the small effort and risk because it really isn't valuable.

SethTisue:
agree. we're spending more time reviewing this can it can possibly be worth

Idloj:
I will try to make a new PR soon taking into account what you've said ( I will rename the forever images on the way(in both code and files) and make the active image's background transparent(to be independent of the active background color, on the way. Any objections to this change? )

Idloj:
From some experiments I can summarize that nothing makes the icons look prominent(talking about lightening, inverting obviously does the work). Maybe special icon for the active state like forever has will do. Or we can leave it as it is, so users can barely see the icons.

mrerrormessage:
Have you tried increasing the brightness of the icons via HSB? That seems like it would help them stand out.

Idloj:
Unfortunately there is no HSBImageFilter, but maybe I can change the color model, I will see.

What I'm currently doing is increase all the RGB values if one of them exceed 255 subtracting all of them so the maximum component is 255.

(This is the code if you want:
var red = (rgb >> 16) & 0xff
    var green = (rgb >> 8) & 0xff
    var blue = (rgb) & 0xff
    red = (value * red).toInt
    green = (value * green).toInt
    blue = (value * blue).toInt
    val difference = Math.max(red, Math.max(green, blue)) - 255
    if (difference > 0) {
      red -= difference
      green -= difference
      blue -= difference
    }
    (rgb & OpaqueAlpha) | (red << 16) | (green << 8) | blue
```
)

PR #861, opened 2015-09-28
Add tooltips when widget labels are truncated

Idloj:


src/main/org/nlogo/window/SliderPainter.java
```diff
@@ -9,8 +9,6 @@
 
   abstract public java.awt.Dimension getMaximumSize();
 
-  abstract public void setToolTipText(String text);
-
```
mrerrormessage: Why was this removed? Was the old behavior causing problems?

src/main/org/nlogo/widget/Switch.scala
```diff
@@ -93,10 +93,14 @@ abstract class Switch extends MultiErrorWidget with MouseWheelListener
...
     g.setColor(getForeground)
+    val shortString = org.nlogo.awt.Fonts.shortenStringToFit(
+      displayName, getWidth - 3 * BORDERX - controlLabelWidth, fontMetrics)
     g.drawString(
-      org.nlogo.awt.Fonts.shortenStringToFit(displayName, getWidth - 3 * BORDERX - controlLabelWidth, fontMetrics),
+      shortString,
       controlLabelWidth + 2 * BORDERX,
       (getHeight - fontMetrics.getHeight - (2 * BORDERY)) / 2 + stringAscent + 1)
+
+    setToolTipText(if(shortString != displayName) displayName else null)
```
mrerrormessage: Please add a space between `if` and the open parens in each file you've changed, per [the scala style guide](http://docs.scala-lang.org/style/control-structures.html).

src/main/org/nlogo/window/SliderPainter.java
Idloj: When this is present, the full variable name shows when you put your mouse over the handler and channel. I think it's more appropriate to show the variable name only when you actually hover the variable name, not when you change the value.

This also makes the tooltip move around every time you move your mouse between the various components(namely handler, channel and the name).

mrerrormessage:
This looks good. As soon as the spacing of `if`s matches my comment, I think this can be merged.

screen shot 2015-09-29 at 2 46 45 pm

Idloj:
Updated

Idloj:
You want me to test it on Linux? What tests should be done?

I tested it as I implemented it.

mrerrormessage:
Tested on windows. This looks great, but should be squashed into a single commit. Then it can be merged in.

Idloj:
`git squash` and `git push -f` ?

mrerrormessage:
See [gitready](http://gitready.com/advanced/2009/02/10/squashing-commits-with-rebase.html) for information about squashing commits with rebase.

Idloj:
I found a stackoverflow answer that gave a neat solution:
```
git reset --soft HEAD~<number of commits, 10 in this case>
git commit
```

Issue #740, opened 2015-03-23
`*-link-neighbor?` primitives wrong for undirected link breeds


qiemem:
To reproduce:

```
undirected-link-breed [ undirected-links undirected-link ]

to test
  ca
  crt 2
  ask turtle 0 [
    create-undirected-link-with turtle 1
    print link-neighbor? turtle 1
    print out-link-neighbor? turtle 1
    print in-link-neighbor? turtle 1
  ]
end
```

When run, this print:

```
true
true
false
```

But it should print

```
true
false
false
```

The docs do indeed specify that `in-link-neighbor?` and `out-link-neighbor?` only return true for directed links.

qiemem:
`*-link-neighbor? foo` should be equivalent to `member? foo *-link-neighbors`.

frankduncan:
So, some confusion on what should be the expected functionality.  The docs say

"Reports true if there is a directed link going from turtle to the caller."

I can read this one of two ways.  One, return true iff there is a directed link going from caller to callee, false otherwise.  Two, return true if there is a directed link going from caller to callee, with the assumption that you are in a directed links model.

The difference is whether we throw an error in the case of a directed link primitive being used on a situation with undirected links (as we do elsewhere in netlogo).  I'm going to update the docs either way we decide, but I don't know what will be better for modelers.  I'm thinking that we should throw an exception since I can't think of a time when you're using undirected links in your model, but want to use directed link primitives.

@SethTisue @UriCCL please weigh in.

qiemem:
`link-neighbor?` errors in the analogous case (the analagous case being the model is not using breeded links and the default breed is directed).

Furthermore, it seems like `*-link-neighbor? foo` should be exactly the same as `member? foo *-link-neighbors`, which also errors (with "LINKS is an undirected breed.")

qiemem:
Further clarification:

This bug is about the situation in which you're using link breeds. In that situation, you can't have "directed links model" (since you can have both directed and undirected in the same model with breeds). So, it should never error.

The only time an error makes sense is if you're not using link breeds, but are using undirected links.

frankduncan:
Hmm, I see.  So then, should directed links of a directed breed even be considered?  Or put differently, should generic link reporters work on breeded links?

frankduncan:
Based on conversation in hipchat, all directed links, breeded or no, should be considered for out-link-neighbor and in-link-neighbor

qiemem:
[...]

PR #866, opened 2015-10-01
Convert prim/_breed* to Scala

Idloj:


src/main/org/nlogo/prim/_breedsingular.scala
```diff
@@ -0,0 +1,29 @@
...
+  override def toString: String = s"${super.toString}:$breedName"
+
+  override def report(context: Context) = report_1(context, argEvalDoubleValue(context, 0))
+
+  def report_1(context: Context, idDouble: Double): AnyRef = {
+    val id = validLong(idDouble)
+    if (id != idDouble)
+      throw new EngineException(context, this, s"$idDouble is not an integer")
+    val turtle = world.getTurtle(id)
+    if (turtle == null) return Nobody$.MODULE$
```
mrerrormessage: This can be converted to just `Nobody`, now that it's in scala.

mrerrormessage: This is causing the Travis build to fail, since `scalac` is choking trying to compile this.

Idloj: I noticed the Travis failure, and I don't know what's happening because compiling goes fine for me.

Idloj: So the change needed is `Nobody$.MODULE$` => `Nobody.MODULE` and `import ... Nobody$` => `import ... Nobody`, is that correct?

mrerrormessage: Change `Nobody$.MODULE$` => `Nobody` and `import ... Nobody$` => `import .. Nobody`. Scala `object`s are referenced as `ObjectName$.MODULE$` in java, but just as `ObjectName` in scala. I'm not sure why it fails on travis but not on your machine, but perhaps performing `sbt clean` would cause it to fail on your machine.

src/main/org/nlogo/prim/_breedhere.scala
```diff
@@ -0,0 +1,30 @@
...
+
+  def report_1(context: Context): AgentSet = {
+    val patch = context.agent match {
+      case t: Turtle => t.getPatchHere
+      case p: Patch  => p
+    }
+    val breed = world.getBreed(breedName)
+    val agentSet = new ArrayAgentSet(classOf[Turtle], patch.turtleCount, false, world)
+    val itr = patch.turtlesHere.iterator
+    while (itr.hasNext) { val turtle = itr.next()
```
mrerrormessage: `val turtle ...` should be on its own line.

Idloj: I made it like that to be similar to foreach, match, etc.
I could get used to it in less than a minute.
It's like:
```
while (itr.hasNext) { turtle =>
  ...
}
```

mrerrormessage:
I think this looks ready for merge after a few corrections. I want to see every conversion of while with a temporary val look like:
```Scala
while (itr.hasNext) {
  val foo = itr.next()
  // ...
}
```
If the while doesn't use a temporary val, no change is needed.

Other than that, looks great, thanks for working on this!

src/main/org/nlogo/prim/_breedhere.scala
mrerrormessage: It makes sense to me that you might want to write it like this, and I agree that it isn't hard to get used to. The problem is that it is not idiomatic scala. Although anyone *can* get used to it, they *have to* get used to it, and it can be a little surprising. `grep`-ing through both our codebase and the scala source, this pattern isn't used unless the `while` is on a single line, for instance:
```Scala
while (i.hasNext) { val a = i.next(); f(a) }
```
The person has to remember one idiom to use in one section of the code and a different idiom in a different section.

If you would like to convert any of the while loops to the above single-line format, the `val` can stay on the same line, otherwise it needs its own line.

mrerrormessage:
As soon as Travis finished, this can be merged.

src/main/org/nlogo/prim/_breedhere.scala
SethTisue: agree. never seen anyone write `while` loops like this

Idloj: I can tell you what you told me once (about tree-style imports that are very uncommon):
> someday the rest of the world will catch up

SethTisue:
thanks @Idloj!

Idloj:
How do we continue from here?

mrerrormessage:
I would pick another group of related primitives and convert them over again. The three groups that stick out immediately are patch primitives, link primitives, or `with` and its optimized forms. The optimized `with` forms are:
* src/main/org/nlogo/prim/_anyotherwith.java
* src/main/org/nlogo/prim/_anywith.java
* src/main/org/nlogo/prim/_countotherwith.java
* src/main/org/nlogo/prim/_countwith.java
* src/main/org/nlogo/prim/_createlinkswith.java
* src/main/org/nlogo/prim/_createlinkwith.java
* src/main/org/nlogo/prim/_linkwith.java
* src/main/org/nlogo/prim/_oneofwith.java
* src/main/org/nlogo/prim/_otherwith.java
* src/main/org/nlogo/prim/_with.java

I expect that the patch primitives would be the easiest. The link primitives have a few outstanding bugs (like #739 and #740). Those would be easier to fix if the link primitives were in scala. The optimized forms of with primitives being in scala would support some optimizations we want to make regarding the efficiency of agentsets. What you would prefer to work on is really up to you!

PR #868, opened 2015-10-08
Show line numbers in the editor. closes #217

### LINKED ISSUE #217: Add line numbers to editor
Suggested by Justin Rising.

Use case: remote co-workers trying to discuss code via phone/skype. Line numbers would make referring to specific parts of the code much easier.

Idloj:
Note that it doesn't work with zooming, thus if you zoom in and back out it will ruin the alignment between the line numbers and the actual code lines.

Other than that, I think it's quite cool.

A possible enhancement would be adding a checkbox to show/hide the bar.
This is a very simple thing to add, so just tell me if it's desirable.

### DIFF
--- a/dist/docs/interface.html.mustache
+++ b/dist/docs/interface.html.mustache
@@ -392,6 +392,16 @@
             &nbsp;
           <td class="nameCenter">
         <p>
+            Show Line Numbers
+          <td class="descriptionRight">
+        <p>
+            Toggle whether the line numbers are shown in code tabs.
+          <tr>
+        <td class="emptyLeft">
+        <p>
+            &nbsp;
+          <td class="nameCenter">
+        <p>
             Shift Left /
             <br>
             Shift Right
@@ -716,13 +726,13 @@
         <p>
             Opens the NetLogo Users Group site in a web browser.
           <tr>
-        <td style='vertical-align: top; border-top:none;border-left:double windowtext 4.5pt; border-bottom:double windowtext 4.5pt;border-right:solid windowtext .5pt; padding:0in 5.4pt 0in 5.4pt'>
+        <td style="emptyLeft">
         <p>
             &nbsp;
-          <td style='vertical-align: top; border-top:none;border-left: none;border-bottom:double windowtext 4.5pt;border-right:solid windowtext .5pt; padding:0in 5.4pt 0in 5.4pt'>
+          <td style="nameCenter">
         <p>
             Donate
-          <td style='vertical-align: top; border-top:none;border-left: none;border-bottom:double windowtext 4.5pt;border-right:double windowtext 4.5pt; padding:0in 5.4pt 0in 5.4pt'>
+          <td style="descriptionRight">
         <p>
             Opens the NetLogo donation page in a web browser.
           </table>
--- a/dist/i18n/GUI_Strings_en.txt
+++ b/dist/i18n/GUI_Strings_en.txt
@@ -64,6 +64,7 @@ menu.edit.redo = Redo
 menu.edit.selectAll = Select All
 menu.edit.find = Find...
 menu.edit.findNext = Find Next
+menu.edit.showLineNumbers = Show Line Numbers
 menu.edit.shiftLeft = Shift Left
 menu.edit.shiftRight = Shift Right
 menu.edit.comment = Comment
--- a/src/main/org/nlogo/app/App.scala
+++ b/src/main/org/nlogo/app/App.scala
@@ -381,7 +381,7 @@ class App extends
     pico.addComponent(tabs.interfaceTab.getInterfacePanel)
     frame.getContentPane.add(tabs, java.awt.BorderLayout.CENTER)
 
-    frame.addLinkComponent(new CompilerManager(workspace, tabs.proceduresTab))
+    frame.addLinkComponent(new CompilerManager(workspace, tabs.codeTab))
     frame.addLinkComponent(listenerManager)
 
     org.nlogo.util.Exceptions.setHandler(this)
@@ -875,7 +875,7 @@ class App extends
    * Returns the contents of the Code tab.
    * @return contents of Code tab
    */
-  def getProcedures: String = dispatchThreadOrBust(tabs.proceduresTab.innerSource)
+  def getProcedures: String = dispatchThreadOrBust(tabs.codeTab.innerSource)
 
   /**
    * Replaces the contents of the Code tab.
@@ -883,7 +883,7 @@ class App extends
    * @param source new contents
    * @see #compile
    */
-  def setProcedures(source:String) { dispatchThreadOrBust(tabs.proceduresTab.innerSource(source)) }
+  def setProcedures(source:String) { dispatchThreadOrBust(tabs.codeTab.innerSource(source)) }
 
   /**
    * Recompiles the model.  Useful after calling
--- a/src/main/org/nlogo/app/CodeTab.scala
+++ b/src/main/org/nlogo/app/CodeTab.scala
@@ -13,7 +13,7 @@ import java.awt.print.PageFormat
 import javax.swing.{JButton, ImageIcon, AbstractAction, Action, ScrollPaneConstants, JScrollPane, BorderFactory, JPanel}
 import org.nlogo.api.I18N
 
-class ProceduresTab(val workspace: AbstractWorkspace) extends JPanel
+class CodeTab(val workspace: AbstractWorkspace) extends JPanel
   with org.nlogo.window.ProceduresInterface
   with ProceduresMenuTarget
   with Events.SwitchedTabsEvent.Handler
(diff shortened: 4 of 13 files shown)


mrerrormessage:
Thanks for the PR! From what you've said, it sounds like this is off to a good start. I would like to see a few changes before considering this for review.

* This absolutely must work with zooming. I understand zooming is really wonky. If you would like we can setup a time to go through and try to sort out how this ought to interact with the Zoomer.
* Line numbers should start disabled and be enabled with a checkbox. I would prefer the checkbox to go either in the "Edit" menu, next to "Snap to Grid" or to rename the "Zoom" menu to "View" and put it there.
* If possible, this should remember whether the user has chosen to use line numbers using [java Preferences](http://docs.oracle.com/javase/7/docs/api/java/util/prefs/Preferences.html) which are persisted between runs. For examples of how these are used in NetLogo, you can look at `I18N`, `RecentFiles`, and `App`.

Idloj:
I also need to sort out dependencies

Idloj:
The only thing to do now is Prefrences to store the last state

Idloj:
I think this is ready for review!

mrerrormessage:
Overall, this is tremendous. Thanks so much for the hard work you put into this!

A few notes:

* The image at the top of `docs/interface.html` needs to be be changed to reflect the changed menus.
* The line numbers nmemonic `L` doesn't work for me on mac or windows. I'm fine with merging this whether or not it works, I just wanted to know if it worked for you. If it doesn't, I'd prefer that it be removed in order to prevent future developer confusion.
* I notice that in windows the numbers have a different background color than the rest of the code tab, while on the mac they share a background color. Do you know what could be causing this? I prefer the distinct background color and would like to see it everywhere. What are you thoughts? If you would like, I might be able to help get the distinct color working on the mac.

windows:
screen shot 2015-11-06 at 6 59 14 pm
mac:
screen shot 2015-11-06 at 6 57 09 pm

SethTisue:
this is awesome!

perhaps it merits a checkbox alongside the "Indent automatically" checkbox? putting it a menu isn't bad, but many users probably won't notice it there. not sure what I think about this.

nbd, but on my Mac, the line numbers seem offset a few pixels vertically from the lines they correspond to. it's visible in Robert's screenshot too, actually

Idloj:
@mrerrormessage
I don't have a Mac, so I can't capture a new image for the docs.
I'll remove the mnemonic (another problem with the mnemonic is that it doesn't work with i18n, this includes the menus mnemonic so it needs to be tackled, too).
The background color on Linux is distinct, too. I don't know the cause of the behavior on Mac, so yeah, It will be great if you can help me with this.

@SethTisue
The offset appears also in Windows, and I guess it's related to the margins. I will to solve it without hardcoding the margins.
I really think it's better in a menu, and I reckon it should be discribed in the realease notes as a UI change. I think it will be sufficient.

Idloj:
May I move "Hide Command Center" to the new View menu?

SethTisue:
I feel strongly it's not a good idea to have new View menu at all, because there's already something in NetLogo called the "view", and this menu has nothing to do with it

mrerrormessage:
@SethTisue , I agree that a "View" menu could be confusing with respect to NetLogo's view, but options like these are typically under a menu called "View", so calling it something else could also be confusing. Do you have any suggestions for what to call it? Some ideas: "Appearance", "Interface" (also a loaded word in NetLogo), and "Presentation". 

@Idloj , I agree that "Show / Hide Command Center" fits with the "View" category, but I think we should decide on whether to move it once we have a name for the menu. By the way, I hadn't forgotten about this PR and was planning to work on the mac side of it sometime next week.

I'm actually wondering whether the existing menus are conceptually the right way to organize NetLogo's functionality. The existing menu structure lumps a lot of stuff into categories which are somewhat arbitrary. I think I might prefer to see a layout that was a bit more centered around the tabs which NetLogo uses. For instance "Interface" / "Run" (for the interface tab), "Code" / "Editor" (for the code tab), "Tools" (for things that don't go into a tab). This would also dovetail nicely with plugins which open a new tab also being able to provide a menu. If we reorganize menus in this way (and I do think we should discuss it more), it should *not* be done in this PR. @SethTisue , what are your thoughts on changing the menu structure?

Idloj:
If you guys are going to discuss a new menu structure, I would recommend to keep in mind the design rule that says that a design should be [*task/goal-oriented*](https://www.youtube.com/watch?feature=player_detailpage&v=yIedljapuz0#t=1559)

Idloj:
@mrerrormessage, I think I've found what causes Mac to show line numbers with the same background as the text. The component used to show the line numbers is JTextPane, which I disable in the constructor. It seems that Mac doesn't change the background to a different color when the component is disabled (other platforms do). The easiest thing to do would be to explicitly define a background color. Sounds good?

mrerrormessage:
Sounds good to me. As I mentioned, I can test/develop on this sometime
early next week.

Idloj:
@mrerrormessage, if you need a list of color defaults to test from on Mac, [here they are](http://alvinalexander.com/java/java-uimanager-color-keys-list)

SethTisue:
sorry, I don't have time to get deeper into this. I agree that any change to the menu structure, including any addition or renaming of any menus at all, should be a separate PR

SethTisue:
and it's probably a discussion Uri would want to be part of. when we reorganized and renamed the tabs, he was very involved.

Idloj:
@mrerrormessage, It seems that Windows doesn't show mnemonics by default (http://superuser.com/questions/16952/how-to-enable-underscore-shortcut-mnemonics-for-menu-items).
So maybe Windows does show mnemonics. Did you open the menu by keyboard(Alt+V)? Did you try and hold Alt while the menu is open?

Idloj:
Brought the mnemonic back

Idloj:
Does the offset on Mac increases as the lines, or is it a constant offset?

Idloj:
Just checked the mnemonic on Windows and it works fine (the menu must be shown with Alt, and even if not, the mnemonic doesn't show but you can still type the letter and make it work).

Idloj:
iirc, you can access the menu item in Mac by typing its first letter.

mrerrormessage:
OK, I've put the mac fix in a branch `mac-line-numbers`. I'll plan to merge it the first week in December, assuming we can figure out the menu situation.

Idloj:
Unfortunately, there seems to be a bug that slipped my attention while testing.
- Open "Wolf Sheep Predation"
- Without scrolling to the end, try to change one of the lines

The line numbers bar will jump to the end until you scroll and fix it.

Idloj:
@SethTisue, I've put a line that *might* fix the vertical offset. Please let me know if it fixes the problem.

Idloj:
@SethTisue, I've put a line that *might* fix the vertical offset. Please let me know if it fixes the problem.

SethTisue:
sorry, don't have time either this week or next

Idloj:
nbd, this little test can still be done by anyone with a Mac

mrerrormessage:
I'll test it on Friday

Idloj:
I've fixed the alignment on Windows and hopefully on Mac too.

Idloj:
In regard to menu structure, I do find it a bit odd that you have, for example, "Snap to grid" and "Uncomment" in the same menu (next to each other, as a matter of fact).

mrerrormessage:
I agree that having "Snap to Grid" and "Uncomment" next to each other in the same menu is a bit odd at first glance, but I think upon closer inspection it actually is a terrific example of goal-oriented design. The goal in this case is to edit the model. You might think of the model as being primarily about the code tab, but the interface tab is an equally important part of the model. Both of those items are really about editing the model. 

The way they are presented to the user could definitely be improved though. At the moment, there is a menu dividing line separating them, which is a good first step. If we wanted to change the edit menu based on the active tab, I think we could add/remove the appropriate items for each tab, but I'm not sure about adding/removing items based on the tab. I would, however, accept a pull request which grayed out the "Comment/Uncomment" and "Shift Left/Right" items if any tab other than a code tab is active (like the "Find..." menu item on the interface tab).

mrerrormessage:
OK, this looks good to me, but we still haven't sorted out the menu issue. Hate to delay this PR any further, but I absolutely agree with Seth that renaming the "Zoom" menu to "View" needs to be vetted, which won't happen until January. At this point, the "Line Numbers" Menu Item can be moved to the "Edit" or "Tools" menus or the PR can wait until the menu situation is figured out.

Idloj:
> At this point, the "Line Numbers" Menu Item can be moved to the "Edit" or "Tools" menus or the PR can wait until the menu situation is figured out.

Is one way preferred over the other?

mrerrormessage:
I would probably prefer "Edit".

Idloj:
In fact, I meant "Move it to Edit" versus "Wait until the menu situation is figured out".

mrerrormessage:
Sorry it's taken me so long to get back on this. Let's move it to "Edit". I want to get this working on hexy, as well as to reduce the number of outstanding hexy branches we have right now.

Idloj:
"Show Line Numbers" is now on the Edit menu.
Documentation updated.

mrerrormessage:
So close. It seems like

* Line numbers are enabled by default, we want them to be opt-in
* Line numbers aren't synced up with preferences on startup (see screenshot)
screen shot 2016-01-12 at 1 18 54 pm

Let me know if you aren't seeing these issues; if they're mac-only I'll look at them.

Idloj:
Fixed. You'll probably need to delete the `line_numbers` entry in `prefs.xml` in order for the fix to take effect.

mrerrormessage:
Looks good. Thanks for all the work on this!

SethTisue:
congrats @Idloj :tada:

PR #870, opened 2015-10-14
Major overhaul of `ExtensionManager`

mrerrormessage:
This is designed to make ExtensionManager flexible enough to handle what we're going to throw at it, as well as to trim down the API as much as possible before we start adding it back in. A number of methods were removed from the api as no longer needed. Most were easy to remove now that we're no longer supporting applets, though one did require a small tweak in the sound extension. Additionally:

* Removes the "Save as Applet" option from the File menu.
* Allows loading "extensions" from in-memory classes for tests via the `InMemoryExtensionLoader`.
* Removes `AppletSaver`.

Further work, to be done in one or more later pull requests, at a date TBD:

* I would really like to have an `ExtensionLoader` which loads extensions from zip files instead of needing them to be unzipped by the user.
* We had talked about automatic downloading of extensions in the future, and I anticipate that that will build on the `ExtensionLoader` system in some way.


SethTisue:
if this is the no-more-applet PR, then the `test/applet` directory can go, and maybe other stuff too — see 033b8b51155f360c145420a2c68fd8fb871b6cad

mrerrormessage:
Needs to be updated and incorporate the fix from #908.

PR #873, opened 2015-10-27
Convert prim/_patch* to Scala

Idloj:
The benchdumps diff is here: https://gist.github.com/Idloj/923fd979f26c5dc79198


mrerrormessage:
I will review this by Tuesday of next week. What I've glanced over so far looks good, but I want to look more carefully at the benchdumps, as they are a bit more complicated than rename changes. Thanks for working on this!

mrerrormessage:
I see only one problem here: `patchcol` and `patchrow` both use `foreach`. They should use the same sorts of while loops that are used elsewhere. It isn't obvious from the dumps (I don't think they're used in any models we dump), but if you look at `javap -c` for each of them you can see that there are some very substantial changes.

Other than that, looks terrific. Looking forward to merging this in soon!

SethTisue:
thanks @Idloj!

PR #879, opened 2015-11-17
Search field is focused when you open the Models Library dialog

Idloj:
for the first time. Subsequent openings open in the last state (probably a model selected). If the search field should be focused on every opening of the Models Library, just let me know.

### DIFF
--- a/netlogo-gui/src/main/app/ModelsLibraryDialog.java
+++ b/netlogo-gui/src/main/app/ModelsLibraryDialog.java
@@ -7,19 +7,28 @@ package org.nlogo.app;
 // instead of sifting through all the files at that time cause that's
 // super slow. ev 3/26/09
 
-import org.nlogo.core.I18N;
-import org.nlogo.swing.BrowserLauncher;
-import org.nlogo.workspace.ModelsLibrary;
-
+import java.awt.Toolkit;
+import java.awt.event.ActionEvent;
+import java.awt.event.KeyAdapter;
+import java.awt.event.KeyEvent;
+import java.net.URI;
 import java.util.Enumeration;
 import java.util.Iterator;
 import java.util.LinkedList;
 import java.util.List;
-
-import java.net.URI;
-
+import javax.swing.AbstractAction;
+import javax.swing.Action;
+import javax.swing.JComponent;
 import javax.swing.JOptionPane;
+import javax.swing.JTextField;
+import javax.swing.KeyStroke;
+import javax.swing.event.AncestorEvent;
+import javax.swing.event.AncestorListener;
 import javax.swing.event.HyperlinkEvent;
+
+import org.nlogo.core.I18N;
+import org.nlogo.swing.BrowserLauncher;
+import org.nlogo.workspace.ModelsLibrary;
 
 strictfp class ModelsLibraryDialog
     extends javax.swing.JDialog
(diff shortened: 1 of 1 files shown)


src/main/org/nlogo/app/ModelsLibraryDialog.java
```diff
@@ -32,11 +34,6 @@ public static String open(java.awt.Frame parent)
       me = new ModelsLibraryDialog(parent);
       me.tree.setSelectionRow(0);
     }
-    // not sure I really understand why this should be necessary
-    // (it isn't on Windows), but at least on Macs, this ensures
-    // that the tree gets the keyboard focus when the dialog opens
-    // - ST 12/17/04
-    me.setFocusable(false);
```
mrerrormessage: @SethTisue , I see that this comment is well over ten years old, but do you remember whether it was particularly important to anyone that the file tree get keyboard focus on dialog open? It seems like the search box is a much more natural place to put the focus, but perhaps that didn't exist in early iterations of this class?

mrerrormessage:
This PR is really a design change. I happen to agree with it, but I want to make sure that I understand the rationale for the prior design (as much as possible) before actually merging this.

src/main/org/nlogo/app/ModelsLibraryDialog.java
SethTisue: I'm afraid I don't remember anything about this

I do like using the arrow keys to navigate the tree, but I'm fine w/ the change.

nicolaspayette:
As a guy that spends a lot of time opening models, I would welcome such a change.

qiemem:
How about the following:

Search box focused by default (as is done here), but then up/down arrows focus model tree (and perform their action there).

Additional features that should probably be added (these are just basic UI things that are common):

- Tab switches focus between tree and search box
- cmd/ctrl-f focuses search box, with the text selected

SethTisue:
> Tab switches focus between tree and search box

That would be great, and would remove any slight hesitancy I might have about the default-focus-location change.

Idloj:
@qiemem, I hope to do these soon, but I fear I don't understand "cmd/ctrl-f focuses search box, with the text selected".

qiemem:
> I fear I don't understand "cmd/ctrl-f focuses search box, with the text selected".

In most application, hitting ctrl-f (in Windows/Linux) brings up a search bar with the text of the search bar selected. Selecting the full text of the search bar is nice because it lets you quickly replace the previous thing you searched for with something new. In OS X, the hotkey is cmd-f instead of ctrl-f. There should be a way to handle the cmd vs ctrl thing automatically, as other NetLogo hotkeys change depending on OS.

I'm suggesting that we replicate this feature in the models library selection dialog. So hitting ctrl-f (on Windows/Linux) or cmd-f (on OS X) focus on the search box (so that when you type, it types in the search box) and selects all the text in the search box (to make it easy to replace your previous search).

Idloj:
> Tab switches focus between tree and search box
Currently, Tab cycles through all the components, so if you want to switch back-and-forth you would press Tab and Shift+Tab. If this is changed so Tab cycles just through the tree and the search box, how would you move to another component (e.g., "Go to User Community Models web page") using keyboard?

qiemem:
Tab and shift tab don't do anything for me at the moment.

mrerrormessage:
@Idloj , it looks like this PR has fallen behind the current hexy branch and can't be merged in without some work. Are you interested in continuing to work on it? If not, I'll go ahead and close it.

Idloj:
I can make it adhere to the current directory structure if that's what you mean

Idloj:
I'm going to squash this into one commit, OK?

mrerrormessage:
sounds good

Idloj:
@mrerrormessage Is there anything I need to do that stops the PR from being merged?

mrerrormessage:
This looks good and ready to merge as soon as it merges cleanly with `hexy`. You can merge hexy into it or rebase it on top of hexy, but other than that it looks good to merge.

Idloj:
Rebased

PR #886, opened 2015-11-23
Add Preview Commands Editor to Hexy

nicolaspayette:
Work in progress. Supersedes #767.


nicolaspayette:
@mrerrormessage, I think this is ready for review. (No rush, though... I know you're busy with 5.2.2...)

- I placed the `previewcommands` package under `app`. It could be argued, however, that it should be a top level package like `mc.gui` or `sdm.gui`. I don't mind moving it if you think I should.

- As a rule, preview commands should run in a freshly opened workspace where `random-seed 0` has been called _before_ opening the model (in case it affects the `startup` procedure).

  - The Modelling Commons image generator used to violate that rule. It is now fixed.

  - `Checksummer.initModelForChecksumming` still violates that rule. I'd like to use the`new `PreviewCommandsRunner` for all the checksumming stuff, but I'm saving that for a future PR.

  - The `__make-preview` primitive violates that rule: it just runs the commands in the current workspace. That's hard to fix because the `prim` package cannot depend on `workspace` to instantiate a new one. I'd be tempted to just ditch the primitive since it is pretty much obsoleted by the preview commands editor, but opinions on the matter are welcome.

- Error handling in the `mc/ImageGenerator.java` is ugly because Java doesn't know what type of exceptions to expect from `scala.util.Try.get()`. I used the pattern described in http://stackoverflow.com/a/20007787/487946.

- `WorkspaceFactory` has been moved to the `workspace` package because it needs to return `AbstractWorkspaceScala` instead of `nvm.Workspace` (because `previewCommands` cannot be set from `nvm.Workspace`). Consequently, `mc` now depends on `workspace` instead of `nvm`.

- As I was saying on #767: " There are no new unit tests in the PR as this is mostly GUI stuff and I consider that the new `api.PreviewCommands` data type is implicitly tested through `TestCompileAll`; but I'm open if anyone has testing suggestions."

- It's possible that some merge conflicts are created when you will merge `5.x` into `hexy`. Let me know if you want me to sort those out when you get there.

This should close #760 once merged.

mrerrormessage:
> I placed the previewcommands package under app. It could be argued, however, that it should be a top level package like mc.gui or sdm.gui. I don't mind moving it if you think I should.

This seems fine by me.

> The `__make-preview` primitive violates that rule: it just runs the commands in the current workspace. That's hard to fix because the prim package cannot depend on workspace to instantiate a new one. I'd be tempted to just ditch the primitive since it is pretty much obsoleted by the preview commands editor, but opinions on the matter are welcome.

I'm fine with the approach that drops it entirely. Alternatively, could you get the desired outcome by setting the rng seed and then using `magicOpen` to reopen the current model? Magic opening doesn't depend on the workspace package, but it seems like it would "restart" the workspace, although I don't know the internals well enough to say for sure. 

> Error handling in the mc/ImageGenerator.java is ugly because Java doesn't know what type of exceptions to expect from `scala.util.Try.get()`.

`ImageGenerator` seems like a pretty simple class, even more so after this change. Would you mind converting it to scala? I'm happy to do the conversion if you would prefer. I think it would be a 25-line scala file and it could avoid the awkwardness of trying to error check a scala `Try` from java.

Everything else in your comment seems fine to me. I have no problem with `mc` depending on `workspace`. Long-term, I would love to see `mc` and `lab` (the two packages which depend on `WorkspaceFactory`) turned into some-sort of plugin and as it is nothing depends on them. I'm going to be reviewing the code and I'll leave any separate comments there. Thanks for this PR, looking forward to the `CheckSummer` PR in the future.

src/main/org/nlogo/swing/ModalProgressTask.scala
```diff
@@ -63,8 +63,8 @@ object ModalProgressTask {
   extends Thread("ModalProgressTask#Boss") {
     override def run() {
       try {
-        while (!dialog.isVisible)
-        Thread.sleep(50)
+        while (!dialog.isVisible) Thread.sleep(50)
```
mrerrormessage: I would prefer if this was styled
```scala
while (!dialog.isVisible) {
  Thread.sleep(50)
}
//...
```
I would be fine with omitting the braces, but `Thread.sleep(50)` needs its own line.

mrerrormessage:
Code looks good, just one minor comment.

src/main/org/nlogo/swing/ModalProgressTask.scala
nicolaspayette: 76c1610

nicolaspayette:
> I'm fine with the approach that drops it entirely. Alternatively, could you get the desired outcome by setting the rng seed and then using magicOpen to reopen the current model?

Maybe it could work. But if you don't mind me dropping the primitive, I will: I don't want people to use it anyway, so I would rather remove it than fix it.

> ImageGenerator seems like a pretty simple class, even more so after this change. Would you mind converting it to scala?

I'll do it!

mrerrormessage:
This looks awesome. I assume it's ready to merge? I'll merge it on Monday unless you want to do more work on it.

nicolaspayette:
:+1:

PR #898, opened 2015-12-08
Convert prim/_link* to Scala

Idloj:
Benchdumps diff: https://gist.github.com/Idloj/79012c33bd3cfa78d1ae

This diff is actually interesting: it is -2 lines.


src/main/org/nlogo/prim/_linkbreedvariableof.scala
```diff
@@ -0,0 +1,41 @@
...
+class _linkbreedvariableof(name: String) extends Reporter {
+  override def syntax: Syntax =
+    Syntax.reporterSyntax(
+      Array[Int](Syntax.LinksetType | Syntax.LinkType),
+      Syntax.WildcardType)
+
+  override def toString: String = s"${super.toString}:$name"
+
+  override def report(context: Context): AnyRef = args(0).report(context) match {
+      case agent: Agent =>
```
SethTisue: incorrect indentation here

src/main/org/nlogo/prim/_linkbreedvariableof.scala
```diff
@@ -0,0 +1,41 @@
...
+        if (agent.id == -1)
+          throw new EngineException(context, this,
+            I18N.errors.getN("org.nlogo.$common.thatAgentIsDead", agent.classDisplayName))
+        try {
```
SethTisue: just a style note, you don't need curly braces with `try` if there's just a single expression inside. (but I don't insist you change it if you prefer the braces.)

src/main/org/nlogo/prim/_linkvariableof.scala
```diff
@@ -0,0 +1,92 @@
...
+  override def syntax: Syntax =
+    Syntax.reporterSyntax(
+      Array[Int](Syntax.LinksetType | Syntax.LinkType),
+      Syntax.WildcardType)
+
+  override def toString =
+      s"${super.toString()}:${if (world != null) world.linksOwnNameAt(vn) else vn}"
+
+  override def report(context: Context): AnyRef = args(0).report(context) match {
+      case agent: Agent =>
```
SethTisue: indentation again

src/main/org/nlogo/prim/_linkvariableof.scala
```diff
@@ -0,0 +1,92 @@
...
+          case ex: AgentException => throw new EngineException(context, this, ex.getMessage)
+        }
+        result.toLogoList
+      case agentOrSet =>
+        throw new ArgumentTypeException(context, this, 0,
+          Syntax.LinksetType | Syntax.LinkType, agentOrSet)
+    }
+
+  def report_1(context: Context, agentOrSet: AnyRef): AnyRef = agentOrSet match {
+      case agent: Agent =>
```
SethTisue: and here

SethTisue:
LGTM other than the indentation problems, which should be fixed before merge

src/main/org/nlogo/prim/_linkbreedvariableof.scala
Idloj: I did this because the `catch` has curly braces, and it looked sot of awkward to have the `try` without and the `catch` with.

mrerrormessage:
Looks terrific. Thanks for doing this!

PR #899, opened 2015-12-08
use proper setter for font size when loading Output widget, fix #896

### LINKED ISSUE #896: 5.2.1 regression: weirdly large font size in output area
Matrix Example with font size of output area set to 12. left is 5.2, right is 5.2.1. 5.2.2-RC2 is similar to 5.2.1

screen shot 2015-12-04 at 11 01 07 pm

nicolaspayette:
(see #896 for context)


mrerrormessage:
Works great, thanks for taking care of this!

PR #901, opened 2015-12-14
make sure `git submodule update` runs from base directory

nicolaspayette:
to allow the `extensions` task to be run from the `models` build.

### DIFF
--- a/project/Extensions.scala
+++ b/project/Extensions.scala
@@ -43,7 +43,7 @@ object Extensions {
       val (packagedNetLogoJar, packagedNetLogoTestJar) = extensionDeps.value
       val s = streams.value
       val scala = scalaInstance.value
-      "git submodule --quiet update --init" ! s.log
+      ("git -C " + base + " submodule --quiet update --init") ! s.log
       val dirs = extensionDirs(baseDirectory.value)
       dirs.flatMap{ dir =>
         cacheBuild(s.cacheDirectory, dir, Set(base / "NetLogo.jar", base / "NetLogoLite.jar"))(


SethTisue:
> to allow the extensions task to be run from the models build

this seems bizarre to me. what is this about?

nicolaspayette:
Some tests in the `models` repo use the NetLogo compiler. Those tests could not be applied to models that use extensions because the `models` build only depended on `NetLogo.jar` and the extensions jars were not available for them. I'm trying to fix that by having the `models` build depend on the full NetLogo project. My (wip) branch for that is here: https://github.com/NetLogo/models/tree/depend-on-hexy-branch.

This will also have the side benefit of allowing `models` to depend on the current state of the NetLogo project without having to wait for some up-to-date `NetLogo.jar` to be uploaded to a devel server.

I'm not quite there yet, but getting close. @mrerrormessage has been helping me working out the kinks.

SethTisue:
holy circular dependencies Batman!

nicolaspayette:
I know it looks weird! :-)

But the way I see it, it's not that NetLogo and the models mutually depend on each other: it's that the model _tests_ depend on both NetLogo and the models. And the tests happen to be in the same repo as the models.

Also, while `models` is a git submodule in the NetLogo repo, the NetLogo project is an sbt dependency in the model tests, so there is no infinite regress happening here.

SethTisue:
> it's that the model tests depend on both NetLogo and the models

yeah, that makes sense. (and not worth making a third repo over)

mrerrormessage:
Looks good. I'm re-running one of the tests and when it passes I'll go ahead and merge this.

PR #907, opened 2015-12-28
Mark the birth version of primitives

Idloj:
There are some issues with this (breeds related prims are not marked, for example), but it's a start.
The process was to diff each and every `tokens.txt` file since NetLogo 1.0 (except NetLogo 4.0, which its "Other" download doesn't include the NetLogo.jar. I assumed that new prims weren't added between 4.0 and 4.0.1), and mark the prims accordingly.
 - Most prims in `3d.html.mustache` that are shared with the normal prims are not marked due to the lack of change in `tokens.txt`.
 - I marked primitives according to their exact form, so for example, I didn't mark `any?` until it had the question mark.
 - Some prims were introduced in early versions, removed later, and then added back in. I did want to mark them as when they added back, but this didn't happen and instead they are marked as when they were first introduced.

Please eyeball it and provide bullets for what needs to be changed. Feedback will be very welcome!

### DIFF
--- a/dist/docs/3d.html.mustache
+++ b/dist/docs/3d.html.mustache
@@ -606,7 +606,7 @@ if (distancexyz 0 0 0) &lt; 10
       </div>
     <div class="dict_entry" id="link-pitch">
       <h3>
-        <a>link-pitch<span class="since">4.0</span></a>
+        <a>link-pitch<span class="since">4.0.2</span></a>
       </h3>
       <h4>
         <span class="prim_example">link-pitch</span>
--- a/dist/docs/3d.html.mustache
+++ b/dist/docs/3d.html.mustache
@@ -606,7 +606,7 @@ if (distancexyz 0 0 0) &lt; 10
       </div>
     <div class="dict_entry" id="link-pitch">
       <h3>
-        <a>link-pitch</a>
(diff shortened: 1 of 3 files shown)


mrerrormessage:
This looks terrific! Thanks for working on this. There are a few primitives that aren't tagged: `create-temporary-plot-pen`, `clear-plot`, `at-points` (3D), `left` (3D), `load-shapes-3d`, `right` (3D) were the ones I found. Keywords and variable names aren't tagged with versions in this, but most of them haven't changed since introduction (I don't think), so that's fine for now. 

You will also need to change `dist/project/PrimIndex.scala`. Replace `.text` on line 41 with `.ownText`. That script scrapes the dictionary and generates the index used for dynamic help. At the moment, it's dragging in the version number, but that's easy to fix. Also, if you're interested in more documentation fixes, I just opened #911 which is about reskinning the dictionary page. Thanks again for working on this!

mrerrormessage:
I would really like to get this in. If you aren't planning to make the changes I listed above, would you mind if I made them? I think this would be really huge for the NetLogo documentation.

Idloj:
Yes, please

mrerrormessage:
OK, closing this PR but I plan to port it over to hexy and put it up at some future point. Thanks for doing the bulk of the work on this!

PR #908, opened 2015-12-31
Bugfix: close url classloader of extension on extension manager reset.

romainreuillon:
In OpenMOLE some users are complaining about a file descriptor leak when using extensions with NetLogo Headless. After investigation it seems that the URL class loader of the extension manager are never closed.

From oracle documentation, is seems that closing this class closes also files opened by the classloader:
https://docs.oracle.com/javase/7/docs/api/java/net/URLClassLoader.html#close%28%29

This patch is an attempt to fix this issue.

### DIFF
--- a/src/main/org/nlogo/workspace/ExtensionManager.java
+++ b/src/main/org/nlogo/workspace/ExtensionManager.java
@@ -9,6 +9,7 @@ import org.nlogo.api.ErrorSource;
 import org.nlogo.api.ExtensionException;
 import org.nlogo.api.Primitive;
 
+import java.io.IOException;
 import java.util.ArrayList;
 import java.util.HashMap;
 import java.util.Iterator;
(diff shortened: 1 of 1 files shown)


mrerrormessage:
I'm looking forward to merging in this PR. But first, the call to `close()` needs to be changed to a guarded reflective call. It can't be called directly because we compile against Java 6. NetLogo compiles against Java 6 and runs against Java 8 yet in order to avoid making backward incompatible changes for extensions. If we were to compile against Java 8, then extensions will also have to compile against Java 8. At the moment, we're planning to compile against Java 8 in the next major release, I'll port this fix over to the branch for that release.

mrerrormessage:
@romainreuillon , I would really like to get this into the next release of NetLogo. Do you plan to make the changes I described above? I'm happy to offer more information if needed. If I don't hear anything on this PR in the next 24 hours or so, I'll go ahead and make the changes necessary.

romainreuillon:
Great! Go head with the changes, I am under a lot of work right now and it is non trivial to install Java 6 on ubuntu nowadays. I think it will be better tested if you implement the changes. However from what we tested this doesn't solve the unclosed jar problem described in issue #909.

mrerrormessage:
Addressed in https://github.com/NetLogo/NetLogo/commit/705eec69a1945446e6daf8fc7f70d53172d5f2e8 . Thanks for starting this PR!

PR #913, opened 2016-01-12
Mathematica relink

mrerrormessage:
Bundles Mathematica with installation.

### DIFF
--- a/dist/build.sbt
+++ b/dist/build.sbt
@@ -5,7 +5,7 @@ import java.util.jar.Attributes.Name.{ CLASS_PATH => JAR_CLASS_PATH }
 
 import java.io.File
 
-import DistSettings.{ aggregateJDKParser, buildNetLogo, buildVariables,
+import DistSettings.{ aggregateOnlyFiles, aggregateJDKParser, buildNetLogo, buildVariables,
   mapToParser, netLogoRoot, netLogoVersion, netLogoLongVersion, numericOnlyVersion,
   packageAppParser, platformMap, settings, subApplicationMap, webTarget }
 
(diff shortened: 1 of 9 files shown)


dist/build.sbt
```diff
@@ -106,6 +108,23 @@ lazy val dist = project.in(file("."))
...
         }
     },
+    packagedMathematicaLink := {
+      val mathematicaLinkDir = netLogoRoot.value / "Mathematica-Link"
+      IO.createDirectory(target.value / "Mathematica Link")
+      Seq(
+        mathematicaLinkDir / "NetLogo-Mathematica Tutorial.nb",
+        mathematicaLinkDir / "NetLogo-Mathematica Tutorial.pdf",
+        mathematicaLinkDir / "NetLogo.m",
+        mathematicaLinkDir / "target" / "scala-2.10" / "mathematica-link.jar")
```
SethTisue: 2.10?

mrerrormessage: `sbt` puts it there. What's the best practice for this situation? I would rather not specify the scala version in the `build.sbt`, since that's irrelevant to the project. I don't really care where the compiled artifact ends up, but I would prefer it be the same even when the scala version used by NetLogo changes.

SethTisue: `crossPaths := false`?

mrerrormessage: yep, that got it.

SethTisue:
trouble in Mathematica land? http://stackoverflow.com/questions/43209522/recommended-hotpatch-for-mathematica-link-to-netlogo-6-didnt-work , http://stackoverflow.com/questions/43289271/unable-to-use-link-between-netlogo-6-0-1-and-mathematica-11-1-on-a-64-bit-proces

PR #921, opened 2016-01-20
Fix non-exhaustive pattern match

qiemem:
To reproduce the problem, make sure you've built at least one extension in hexy. I'll use the `csv` extension, but it doesn't matter.

Open a new model, enter `extensions [ csv ]` in the code tab. Switch to the interface tab, and type "csv:". When you hit the ":", an exception will occur.


mrerrormessage:
Looks great, thanks!

PR #739, opened 2016-01-26
`link-neighbor?` wrong for directed, breeded links

mrerrormessage:
To reproduce:

```
directed-link-breed [ directed-links directed-link ]

to test
  crt 2
  ask turtle 0 [
    create-directed-link-to turtle 1
    print link-neighbor? turtle 1
  ]
end
```

Running `test` prints `true` but should print `false`. According to the docs:

According to the docs:

> Reports true if there is an undirected link between turtle and the caller.

Note that it should print `true` if link is an undirected link breed. If the link breed is directed, unbreeded, it errors with:

> LINKS is a directed breed.

which is arguably correct since you can't have breeded and unbreeded links in the same model.

### DIFF
--- a/src/main/org/nlogo/prim/_linkneighbor.java
+++ b/src/main/org/nlogo/prim/_linkneighbor.java
@@ -3,6 +3,7 @@
 package org.nlogo.prim;
 
 import org.nlogo.agent.AgentSet;
+import org.nlogo.agent.Link;
 import org.nlogo.agent.LinkManager;
 import org.nlogo.agent.Turtle;
 import org.nlogo.api.LogoException;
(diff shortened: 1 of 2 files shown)


mrerrormessage:
From @qiemem : The following appears to be returning false:
```scala
undirected-link-breed [ undirected-links undirected-link ]
to test
  crt 2
  ask turtle 0 [
    create-undirected-link-with turtle 1
    print link-neighbor? turtle 1
  ]
end
```

mrerrormessage:
@qiemem , I'm getting your snippet above to print `true`. Maybe you ran it in the wrong NetLogo window or needed to do a `clean` before running? I've added tests to cover each case, regardless, so this should be good to go.

qiemem:
Hmm, ya, getting `true` now as well... Sorry about that!

mrerrormessage:
Not a problem, good to have the extra tests!

PR #922, opened 2016-01-27
Enables codeblocks, symbols, and lots of other awesome stuff for LevelSpace

qiemem:


src/main/org/nlogo/api/Argument.scala
```diff
@@ -155,4 +155,23 @@ trait Argument {
   @throws(classOf[LogoException])
   def getCommandTask: CommandTask
 
+  /**
+   * Returns the argument as a <code>org.nlogo.api.CommandTask</code>.
+   *
+   * @throws ExtensionException if the argument is not a <code>CommandTask</code>
```
mrerrormessage: Give the full path to this class (`org.nlogo.api....`), since scaladoc isn't smart enough to figure out what you mean. This should be done for each of the `@throws` annotations added here.

src/main/org/nlogo/compiler/CompilerMain.scala
```diff
@@ -29,7 +29,7 @@ private object CompilerMain {
     for(procedure <- structureResults.procedures.values.asScala) {
       procedure.topLevel = subprogram
       val tokens =
-        new IdentifierParser(program, oldProcedures, structureResults.procedures, false)
+        new IdentifierParser(program, oldProcedures, structureResults.procedures, forgiving = true)
```
mrerrormessage: After this change, is there any time we want to run this with `forgiving = false`? If not, let's remove the argument.

src/main/org/nlogo/compiler/ExpressionParser.scala
```diff
@@ -557,6 +568,50 @@ private class ExpressionParser(procedure: Procedure,
...
+    else if(compatible(goalType,Syntax.CodeBlockType)) {
+      //new CodeBlock(block.tokens.mkString(" "),tokens.head.startPos,block.tokens.last.endPos,tokens.head.fileName)
+      import scala.collection.JavaConverters._
+
+      // Because we don't parse the inside of the code block, precisely because we don't want to define
+      // legality of code in terms of the standard NetLogo requirements, we have to do a little sanity
+      // checking here to make sure that at the very least, parenthesis and brackets are matched up
+      // without being mixed and matched.  FD 8/19/2015
+
+      def check(remaining: Seq[Token], stack: Seq[Token] = Seq()) {
```
mrerrormessage: I would strongly prefer that this be annotated with `@tailrec`.

src/main/org/nlogo/compiler/IdentifierParser.scala
```diff
@@ -104,7 +98,7 @@ private class IdentifierParser(program:Program,
     else if(forgiving)
       new _unknownidentifier
     else
-      exception("Nothing named " + varName + " has been defined",
+      exception(I18N.errors.getN("compiler.LocalsVisitor.notDefined", varName),
```
mrerrormessage: Hooray for I18N :+1:

src/main/org/nlogo/nvm/LetBinding.java
```diff
@@ -4,9 +4,9 @@
 
 import org.nlogo.api.Let;
 
-strictfp class LetBinding {
-  final Let let;
-  Object value;
+strictfp public class LetBinding {
+  public final Let let;
+  public Object value;
```
mrerrormessage: @TheBizzle , what were you saying about grad students going around making all the fields public....

In this case I would prefer public accessor methods for anything outside of nvm, but I'm fine the with the fields being package-private.

src/main/org/nlogo/prim/_constcodeblock.java
```diff
@@ -0,0 +1,38 @@
...
+
+  @Override
+  public Object report(Context context) {
+    return value;
+  }
+
+  /*
+  public List<Token> report_1(Context context) {
+    return value;
+  }*/
```
mrerrormessage: This commented-out method should be removed. Also, I would prefer if this primitive were implemented in scala (although it's simple enough that I'm fine with merging it without it being in scala).

src/main/org/nlogo/nvm/LetBinding.java
TheBizzle: > @TheBizzle , what were you saying about grad students going around making all the fields public?

The great irony here is that @qiemem himself was the one saying that.  :laughing:

src/test/org/nlogo/compiler/AutoConverter2Tests.scala
```diff
@@ -1,40 +0,0 @@
-// (C) Uri Wilensky. https://github.com/NetLogo/NetLogo
-
-package org.nlogo.compiler
-
-import org.scalatest.FunSuite
-import org.nlogo.nvm.DummyWorkspace
-
-class AutoConverter2Tests extends FunSuite {
```
mrerrormessage: :+1: :+1: :+1:

src/main/org/nlogo/nvm/LetBinding.java
mrerrormessage: My mistake! Well, it's funny how things go sometimes.

mrerrormessage:
OK, so I left a lot of comments in the code. Overall this looks great and I'm really excited about this feature! Before this is ready to merge, it needs a few (2 or 3) language tests for codeblocks and symbols. They don't need to be big, they probably just need to be an underscore primitive that prints the symbol/codeblock that was provided.

I'm happy to help fix stuff, just let me know what you would like me to tackle.

src/main/org/nlogo/nvm/LetBinding.java
qiemem: Well, that's embarrassing :P Could I just change this to a scala case class, making `value` a `var`? That gives us customizable accessors if we want them, with convenient syntax.

src/main/org/nlogo/prim/_constcodeblock.java
qiemem: I'll change it. This is Frank territory, though I should've caught that.

src/main/org/nlogo/nvm/LetBinding.java
mrerrormessage: I think that's a great idea.

src/main/org/nlogo/compiler/CompilerMain.scala
qiemem: Good call.

Here's one place where it was still false: https://github.com/NetLogo/NetLogo/blob/cd253418c13860c1ee1c5874be66480fc2019982/src/main/org/nlogo/compiler/CompilerMain.scala#L32

Pretty sure this is incorrect though, and would've prevented the use of symbols in reporters given at the command line.

src/main/org/nlogo/compiler/IdentifierParser.scala
qiemem: Alas, I'm deleting this conditional branch since `forgiving` is getting tossed...

src/test/org/nlogo/headless/TestCompiler.scala
```diff
@@ -147,8 +167,7 @@ class TestCompiler extends FunSuite with OneInstancePerTest with BeforeAndAfterE
   /// isReporter
 
   val reporters = Seq("3", "[]", "[", "((5))", "timer", "glob1")
-  val nonReporters = Seq("", ";", " ; ", "ca", "((ca))",
-                         "5984783478344387487348734", "gkhjfghkjfhjkg")
+  val nonReporters = Seq("", ";", " ; ", "ca", "((ca))", "5984783478344387487348734")
```
qiemem: This may be a particularly controversial consequence of removing the `forgiving` parameter, though its a natural result of allowing symbols in the language.

mrerrormessage: I think this is correct. From the point of view of the compiler, an unbound symbol shouldn't be identified as a reporter or a non-reporter.

mrerrormessage:
This looks good. It can be merged as soon as tests are added around the new features and all tests pass on Travis. Are you working on the language tests?

Travis only has one failing test: `TestProfiler`. It should be easy to fix, as it's another `isReporter` test, just deleting it should be fine.

src/main/org/nlogo/nvm/LetBinding.java
SethTisue: fwiw, I made them fields on purpose for maximum performance — this is extremely low-level, extremely inner-loop stuff.

(but did I benchmark it both ways? no. for all I know HotSpot would figure it out and the performance would be the same.)

qiemem:
:tada:

PR #925, opened 2016-01-31
Add Link to StackOverflow Q & A. Close #887

### LINKED ISSUE #887: Add Link to StackOverflow Q & A to "Help" menu
* Menu entry for NetLogo Q & A on stackOverflow

Idloj:
Of course I can change the text and order of the menuitem, let me know if there's a need for anything.
(I didn't actually run it because I can't run the 5.x branch, but I suppose it'll work fine)

### DIFF
--- a/dist/i18n/GUI_Strings_en.txt
+++ b/dist/i18n/GUI_Strings_en.txt
@@ -103,6 +103,7 @@ menu.help.lookUpInDictionary(F1) = Look Up In Dictionary (F1)
 menu.help.netLogoUserManual = NetLogo User Manual
 menu.help.netLogoDictionary = NetLogo Dictionary
 menu.help.netLogoUsersGroup = NetLogo Users Group
+menu.help.netLogoStackOverflow = NetLogo Q & A - Stack Overflow
 menu.help.donate = Donate
 
 tabs.run = Interface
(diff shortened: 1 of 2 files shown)


dist/i18n/GUI_Strings_en.txt
```diff
@@ -103,6 +103,7 @@ menu.help.lookUpInDictionary(F1) = Look Up In Dictionary (F1)
 menu.help.netLogoUserManual = NetLogo User Manual
 menu.help.netLogoDictionary = NetLogo Dictionary
 menu.help.netLogoUsersGroup = NetLogo Users Group
+menu.help.netLogoStackOverflow = NetLogo Q & A - StackOverflow
```
SethTisue: there should be a space between the words Stack and Overflow

Idloj: Can I omit "Stack Overflow" (I mean, "NetLogo Users Group" doesn't mention Yahoo, does it?).

SethTisue: I think it's better to leave it in. Informally, we call netlogo-users netlogo-users and the fact that it's on yahoo is incidental, but in the Stack Overflow case, it's not really its own thing, it's just "Stack Overflow". But not everybody knows what that is, so the "NetLogo Q&A" should be there too IMO.

mrerrormessage:
Looks good, thanks for the PR!

SethTisue:
:tada:

PR #927, opened 2016-02-09
upgrade ScalaTest and ScalaCheck

SethTisue:
no special reason, just keeping current


mrerrormessage:
Since project structure changed so much, this is addressed in #954 . Thanks for the PR, I'm excited about scala 2.11.8 coming soon.

Issue #940, opened 2016-02-24
Replace-And-Find deletes any text selected when opened


mrerrormessage:
Steps to reproduce:
- Select text in the code tab
- Open find/replace with ctrl/cmd-f
- enter anything in find
- click find & replace
- will delete all text selected

To fix we just need to deselect any text initially selected when the user opens the find/replace tool

Idloj:
It will also be nice to use the selected text as the initial value for "Find". (This is actually pretty standard in Find/Replace)

mrerrormessage:
Terrific idea! Let's do that.

mrerrormessage:
Closed with #946 .

PR #946, opened 2016-03-01
Replace and Find honouring selection

Akarshit:
Fix for #940 
This keeps the user selection as it is if nothing is found.
Now if the Replace and Find is selected for the first time it finds the text and shows it to the user. 
If Replace and Find is pressed again then the text is replaced and next occurrence is found.
This also handles the special case in which the text to be found is already selected. In this case the selected text is replaced in the first go.

The feature in which the selected text is used as initial value for find field will be done is a separate PR.

### DIFF
--- a/src/main/org/nlogo/app/FindDialog.java
+++ b/src/main/org/nlogo/app/FindDialog.java
@@ -226,7 +226,12 @@ public strictfp class FindDialog
         notFoundLabel.setVisible(false);
       }
     } else if (e.getSource().equals(replaceAndFindButton)) {
-      replace(replaceBox.getText());
+      // Stores if the Replace and Find is selected for the first time.
+      // Also checks if the text in find field and selected text are same or not.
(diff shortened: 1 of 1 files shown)


mrerrormessage:
Hi @Akarshit , thanks for the pull request! Can you combine the selected-text-used-as-initial-value fix into this PR? I think that once that is done a lot of the logic you have for the replace-and-find button won't be necessary.

Akarshit:
@mrerrormessage I have added the feature. But this does not help is simplifying the logic of Find and Replace as we have to handle those case in which the user changes the text of Find field.
For example if "turtles" was selected in the code then the user opened the Find dialog, the Find field would contain "turtles" by default. But if now he manually changes the Find field to something else (say "tortoise"), the replace and find will have to find the text first and later replace it.
The conditions in Replace and Find are handling these special cases.

mrerrormessage:
Nice catch! This looks good and can be merged. As soon as you sign the CLA on this or #947, it should clear for merge on both builds and I'll go ahead and do the merge.

Issue #831, opened 2015-07-14
Make auto-formatting more discoverable in the UI


Idloj:

SethTisue:
Wouldn't that just be duplicating the functionality that already exists on the Edit menu? I might be misunderstanding.

Idloj:
What functionallity? I'm talking about a button that format your entire code automatically(like in most IDEs)

SethTisue:
Oh, forgetful me, it's not on the Edit menu. But it's hooked up the Tab key. It already does that if you select your code and hit the Tab key.

Maybe it should be made more discoverable in the UI?

Idloj:
agree

Idloj:
I think providing this functionality through the Tab key is not a good choice, better be in something like Ctrl+F or whatever. I didn't see any IDE which does through Tab, and for a very good reason. The Tab is used for indenting. Not to mention the great inconsistency between automatically indentation enabled/disabled.

mrerrormessage:
This can be done with an entry in the edit menu somewhere around comment / uncomment, indent / outdent. We already have a shortcut key for this (tab) and we could add a ctrl-tab mnemonic if that would help make it clearer.

Idloj:
I suggest that if no text is selected it will format the entire code.

Idloj:
Summary of current behavior:

Shift+Tab always shifts the current line (or current lines of selected text) to the left by 2 spaces.
Shift Left/Right in the edit menu always shifts the current line (or current lines of selected text) to the left/right by 1 space.

Tab key behavior:

 | Selected text | No selected text
--- | --- | ---
Auto-indent enabled | Format selected text | Nothing
Auto-indent disabled | Replace selected text by 2 spaces | Adds 2 spaces at cursor position

Idloj:
I think it'll do much good to decide on a consistent behavior for the Tab key, regard less of auto-indentation state, to prevent [mode errors](https://en.wikipedia.org/wiki/Mode_%28computer_interface%29#Mode_errors).
My thought splits between two approaches:
 - Tab will just add 2 spaces at cursor position when no text is selected, while indenting the lines of selected text if there is some.
 - Tab will indent the current lines (or current lines of selected text), which makes Shift Left/Right redundant (or making Tab and Shift+Tab the shortcuts for Shift Left/Right).

I think I'm more inclined to the second option.

Anyway, I think Shift Left/Right should shift by 2, not 1.

mrerrormessage:
[...]

PR #947, opened 2016-03-02
Added the format shortcut to Edit menu

Akarshit:
Fix for #831 
The shortcut is still kept to be `Tab` as `Ctrl + Tab` switches the tab to Interface Tab (atleast on Linux)
Had to overload functions in `Menu.scala` as the default function was ORing the mask for menu item by default.


Idloj:
I think the appropriate name will be "Format" rather than "Indent" (which is more of what "Shift right" does)

Akarshit:
@Idloj 
Done, thanks.

mrerrormessage:
This looks good, I'll need to test and then I'll merge.

mrerrormessage:
Tested and looks good. This is ready for merge, just need you to sign the CLA above.

Akarshit:
@mrerrormessage Done.

Issue #813, opened 2015-07-06
Open/Save dialog should show only .nlogo files for NetLogo and only .nlogo3d files for NetLogo 3D


Idloj:

nicolaspayette:
Agreed. Maybe this could be fixed as part of https://github.com/NetLogo/NetLogo/pull/759. (I think this may be a Linux specific bug as well.)

Idloj:
Why would it be Linux specific, is it coded anywhere?

nicolaspayette:
That's just a hunch. I haven't actually verified. But:
- Linux is the only platform to use `JFileChooser`;
- #720 was Linux specific;
- No one else has reported this before (Mac/Windows versions have a lot more eyeballs on them).

The guilty code is likely in https://github.com/NetLogo/NetLogo/blob/5.x/src/main/org/nlogo/swing/FileDialog.java.

If you're going to go in and change anything, I would suggest doing it from the https://github.com/NetLogo/NetLogo/pull/759/ branch.

Idloj:
The intent to do this is actually in the comment:
>     // Incidentally, I'd like to make NetLogo properly support filename filters too...
    // But I'm guessing this late in the "ready-to-release" cycle, people might not be
    // too pleased with me hacking much and possibly introducing new bugs.  :-)
    //  So I'm just making a change that affects Linux users...
    //  ~Forrest (6/8/2009)

mrerrormessage:
Closed in #953

PR #949, opened 2016-03-03
Added extension filter for linux File chooser

Akarshit:
Fix for #813

### DIFF
--- a/src/main/org/nlogo/swing/FileDialog.java
+++ b/src/main/org/nlogo/swing/FileDialog.java
@@ -70,11 +70,12 @@ public final strictfp class FileDialog {
     // but I'm not sure.
     //  ~Forrest (6/8/2009)
     if (LINUX) {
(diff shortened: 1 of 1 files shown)


src/main/org/nlogo/swing/FileDialog.java
```diff
@@ -69,14 +68,26 @@ private static String show(java.awt.Frame parentFrame, String title,
...
+            return true;
+          }
+          return f.getName().endsWith(ext);
+        }
+
+        @Override
+        public String getDescription() {
+          return "NetLogo files (*" + ext + ")";
+        }
+	  };
```
mrerrormessage: This line is indented with a tab, should be two-spaces instead.

mrerrormessage:
Looks good. I would prefer to have the two commits squashed into one and then this can be merged.

Akarshit:
Will issue another

Issue #814, opened 2015-07-06
Add an asterisk in front of the title to indicate if the model was modified since last saved


Idloj:

nicolaspayette:
You mean to indicate if the model was modified since last saved? (Don't hesitate to give more details when reporting issues or making suggestions. What's obvious to you may not be to the person reading your report.)

Idloj:
Precisely

mrerrormessage:
Closed in #860, #955 .

PR #860, opened 2015-09-28
Add an asterisk to the title to indicate dirtiness. closes #814

Idloj:

PR #952, opened 2016-03-04
Added * after title to indicate modified model

Akarshit:
Fix for #814 
Added Document change listeners and added `*` accordingly.

### DIFF
--- a/project/autogen/events.txt
+++ b/project/autogen/events.txt
@@ -35,6 +35,7 @@ window.LoadBegin
 window.LoadEnd
 window.LoadSection - String version - Object section - String[] lines - String text
 window.ModelSaved - String modelPath
+window.ModelModified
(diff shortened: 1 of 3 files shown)


mrerrormessage:
I appreciate the PR, but this needs to be done similarly to #860. The way you currently have it done will work for changes in the editor, but not to other parts of the model. Lots of parts of a NetLogo model other than code areas can change. To name a few:

* Widget attributes and values
* Info tab
* BehaviorSpace
* AggregateManager

We would accept backporting #860. The reason the fix from 860 isn't already present on 5.x is that it was made to our "next major version" branch (hexy).

Akarshit:
@mrerrormessage OK got it.
So i have to create a PR just like #860 ?

mrerrormessage:
Yep, that's it.

PR #953, opened 2016-03-04
Added extension filter for linux File chooser

Akarshit:
Fix for #813


mrerrormessage:
Looks good.

In the future when squashing you can just force push up to your branch with the altered commit(s) and GitHub will change the PR for you automatically.

Akarshit:
@mrerrormessage Thanks, will keep that in mind in future.

PR #956, opened 2016-03-07
Open dialog reopens after wrong directory path

Akarshit:
Fix for #726 
This is a fix for Linux ( I only have Linux) but if this looks fine, I can do a fix for other OS.


mrerrormessage:
This looks great, thanks!

Issue #450, opened 2013-10-05
error while undoing


turtlemaniac:
NetLogo is unable to supply you with more details about this error.  Please report the problem
at https://github.com/NetLogo/NetLogo/issues, or to bugs@ccl.northwestern.edu, and paste the
contents of this window into your report.

javax.swing.undo.CannotUndoException
 at javax.swing.text.GapContent$InsertUndo.undo(GapContent.java:856)
 at javax.swing.undo.CompoundEdit.undo(CompoundEdit.java:46)
 at javax.swing.text.AbstractDocument$DefaultDocumentEvent.undo(AbstractDocument.java:2788)
 at javax.swing.undo.CompoundEdit.undo(CompoundEdit.java:46)
 at javax.swing.undo.UndoManager.undoTo(UndoManager.java:320)
 at javax.swing.undo.UndoManager.undo(UndoManager.java:398)
 at org.nlogo.editor.UndoManager.undo(UndoManager.java:63)
 at org.nlogo.editor.UndoManager$UndoAction.actionPerformed(UndoManager.java:139)
 at javax.swing.SwingUtilities.notifyAction(SwingUtilities.java:1639)
 at javax.swing.JComponent.processKeyBinding(JComponent.java:2859)
 at javax.swing.JComponent.processKeyBindings(JComponent.java:2894)
 at javax.swing.JComponent.processKeyEvent(JComponent.java:2822)
 at java.awt.Component.processEvent(Component.java:6074)
 at java.awt.Container.processEvent(Container.java:2039)
 at java.awt.Component.dispatchEventImpl(Component.java:4660)
 at java.awt.Container.dispatchEventImpl(Container.java:2097)
 at java.awt.Component.dispatchEvent(Component.java:4488)
 at java.awt.KeyboardFocusManager.redispatchEvent(KeyboardFocusManager.java:1856)
 at java.awt.DefaultKeyboardFocusManager.dispatchKeyEvent(DefaultKeyboardFocusManager.java:722)
 at java.awt.DefaultKeyboardFocusManager.preDispatchKeyEvent(DefaultKeyboardFocusManager.java:1000)
 at java.awt.DefaultKeyboardFocusManager.typeAheadAssertions(DefaultKeyboardFocusManager.java:865)
 at java.awt.DefaultKeyboardFocusManager.dispatchEvent(DefaultKeyboardFocusManager.java:686)
 at java.awt.Component.dispatchEventImpl(Component.java:4532)
[...]

PR #957, opened 2016-03-07
Undo only if the edit can be undone

Akarshit:
Fix for #450.
Prevents the original problem from occurring instead of catching the Exception.


mrerrormessage:
Hi Akarshit, thanks for opening all of these PRs. Due to some circumstances outside my control, I may not be able to review them until late next week. Just wanted to let you know that I have seen them and I will review them when possible. Thanks again!

Akarshit:
@mrerrormessage No problem.

mrerrormessage:
Looks good. Thanks!

Issue #364, opened 2013-05-29
some file-* primitives may double-evaluate their inputs when an error occurs


SethTisue:
by inspection of the source code, the affected prims are `file-delete` and `file-exists?`; note the repeated calls to `argEvalString`

mrerrormessage:
Closed in #959

PR #959, opened 2016-03-07
Inputs evaluated once in case of error

Akarshit:
Fix for #364


mrerrormessage:
This looks good. Ordinarily I would want to see a test for changes like this, but in this case I don't think it's possible to generate a failing test without doing some really hacky stuff (I tried). Thanks for the PR!

Issue #961, opened 2016-03-07
Plot pens 'edit' and 'delete' buttons are not aligned well


Idloj:
![netlogo-plot](https://cloud.githubusercontent.com/assets/8083458/13581348/225fd69a-e4af-11e5-8719-a3485a588bd8.png)

mrerrormessage:
Fixed in #962

PR #962, opened 2016-03-08
Set layout for buttons and added margins

Akarshit:
Fix for #961 
I have used Group Layout so i think this should be fixed for all platforms and resolutions.
The buttons now look like below.

screenshot from 2016-03-08 21 46 10
In the second commit a different issue is addressed. When editing the `Pen Update Command` column the right and bottom grid used to disappear like below

screenshot from 2016-03-08 21 53 32

To prevent this a margin of 1 pixel is added and the row height is increased to make the buttons appear in center(vertically).


mrerrormessage:

PR #964, opened 2016-03-08
Removed control strip and added ticks to ViewUpdatePanel

Akarshit:
Fix for #930 
* Basically deleted every line that contained the word  `controlStrip`.
* Added the ticks counter below the speed slider.
* Deleted the spaces before "ticks"
* Added spaces before normal-speed to center(though this should be done in a better way but that would be a different PR)

The interface looks like this now:
screenshot from 2016-03-09 00 42 10

### DIFF
--- a/src/main/org/nlogo/app/InterfacePanel.java
+++ b/src/main/org/nlogo/app/InterfacePanel.java
@@ -258,8 +258,7 @@ strictfp class InterfacePanel
     if (viewWidget instanceof org.nlogo.window.ViewWidget &&
         !type.equals("GRAPHICS-WINDOW") &&
         VersionHistory.olderThan13pre1(modelVersion)) {
-      y += ((org.nlogo.window.ViewWidget) viewWidget).getExtraHeight() +
-          ((org.nlogo.window.ViewWidget) viewWidget).controlStrip.getHeight();
+      y += ((org.nlogo.window.ViewWidget) viewWidget).getExtraHeight();
     }
     if (type.equals("GRAPHICS-WINDOW")) {
       // the graphics widget (and the command center) are special cases because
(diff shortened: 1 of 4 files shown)


Idloj:
Don't you want to remove ViewControlStrip altogether?

Akarshit:
@Idloj I didn't remove it in case the control need to be added elsewhere in future. Should i remove it?

Idloj:
I dunno. @mrerrormessage ?

mrerrormessage:
I appreciate the PR, but this is actually a really big change. I'm going to close this PR because this can't be done on the 5.x series and the following aspects will need to be considered first:

* The centering of the ticks label can't be done using spaces. It needs to be done using a box or a border layout (or perhaps some more exotic swing layout which allows for centering). This isn't just a best practice but a necessity because both the ticks label can be changed (to "steps" or even "", for instance) *and* the number of ticks can get quite large, potentially having many more digits.
* The swing layout will need to be adjusted for Mac OS X and Windows Swing. The Linux screenshot looks good, and I would expect windows to be similar. I think Mac will likely require some adjustment, however.
* We need to figure out how to do this consistently with NetLogo 3D. At the moment, NetLogo 3D displays ticks to the left of the speed slider in the bar above the 3D view, but we want the layout to be as consistent as possible. I would prefer 3D have either (a) all controls on the interface tab, or (b) all controls on the 3D view, or (c) the 3D view as part of the interface tab (just like the 2D view is). (c) would be a major change in its own right, whether we choose (a) or (b) should be discussed.

I *do* really appreciate this work, even though this PR can't get merged right now. I'm creating a new hexy branch [for code related to removing the control strip](https://github.com/NetLogo/NetLogo/tree/wip-remove-control-strip). I would merge the code in this PR into that branch, where we could cooperatively work on it as we adjust for the above three items. I can work on the Mac (and Windows, if needed) changes on the branch while we also figure out the other two items.

Akarshit:
@mrerrormessage Yes i also thought about (a) and (b). But i have not centered the ticks using spaces, it's the labels `normal speed`, 'slower`, `faster` that were positioned using spaces. I just added two extra spaces before `normal speed` to center it as it was quite out of place.
But yes i agree with your points.

PR #963, opened 2016-03-09
post-merge hexy: Accessing latter arguments in a mutlivariate foreach sets all preceding arguments to that value

mrerrormessage:
Confused by the title? Good, because this is really weird:

```
observer> (foreach [1 2 3] [4 5 6] [ print ?2 print ?1 ])
4
4
5
5
6
6
```

To be clear, the expected output is:

```
4
1
5
2
6
3
```

Note that `?2` is accessed before `?1`. Let's flip them:

```
observer> (foreach [1 2 3] [4 5 6] [ print ?1 print ?2 ])
1
4
2
5
3
6
```

Using 3 lists makes the behavior clearer:

```
observer> (foreach [1] [2] [3] [ type ?2 type ?1 type ?3 type ?1 type ?2 type ?1 ])
223333
```

So, accessing `?2` set `?1` to the value of `?2`, but then accessing `?3` set *both* `?1` and `?2` to the value of `?3`.


mrerrormessage:
See PR #966. This was opened against the wrong branch.

PR #966, opened 2016-03-09
Fixes command task variables (#963)

mrerrormessage:
See #963 for description.


netlogo-gui/src/main/nvm/Procedure.scala
```diff
@@ -46,8 +46,8 @@ class Procedure(
 
   def getTaskFormal(n: Int, token: Token): Let = {
     while (taskFormals.size < n)
-      taskFormals += new Let("?" + n)
-    taskFormals.last
```
SethTisue: haha who wrote this crap? oh... me?

SethTisue:
should this be backported to 5.x?

mrerrormessage:
The bug doesn't exist in 5.x, Bryan's example works as expected. I didn't trace down the *entire* reason why this broke, but I believe it was because under hexy tasks are compiled as `_commandtask[...]` and `_reportertask[...]` instead of `_task[_commandtask[...]]` and `_task[_reportertask[...]]`.

PR #969, opened 2016-03-09
fix for grepping javac path for openjdk or ppa:webupd8team/java packa…

tirnak:
grep arguments changed to be able to get "/usr/lib/jvm/java-8-oracle/bin/javac" path, which is default for ubuntu's openjdk or webupd8team's oracle JDK.
And surely, dot is to be escaped.


mrerrormessage:
Thanks for the PR!

Issue #895, opened 2015-12-05
NaN and Infinity can be produced by multiplication


qiemem:
`2 ^ 1024` errors correctly, but:

```
observer> show (2 ^ 1023) * 2
observer: Infinity
observer> show ((2 ^ 1023) * 2) / ((2 ^ 1023) * 2)
observer: NaN
```

It looks like `_mult` doesn't call `isValidDouble`. That said, we should just use Java 8's new `isFinite` instead. We definitely need to thoroughly benchmark any of these changes since they touch basic math operations.

SethTisue:
`validDouble` should be changed to call `isFinite`, you mean. `validDouble` doesn't return a boolean, it actually errors if the input isn't valid

SethTisue:
there are a bunch of other prims (including the bytecode-generated version of `_plus`) that ought to call `validDouble` or `newValidDouble` but don't.

qiemem:
Ah, yes, you're right.

mrerrormessage:
The fix here is to ensure that all prims which generate a new number call `validDouble` or `newValidDouble` on their result before returning it. The following prims are the ones which can return a number:

* `count`
* `countother`
* `countotherwith`
* `countwith`
* `minus`
* `nsum`
* `nsum4`
* `patchvariabledouble`
* `plus`
* `random`
* `randomconst`
* `sum`
* `turtlevariabledouble`
* `unaryminus`
* `randomorrandomfloat`
* `abs`
* `acos`
* `approximatehsb`
* `approximatehsbold`
* `approximatergb`
* `asin`
* `atan`
* `behaviorspacerunnumber`
* `ceil`
* `cos`
* `distance`
* `distancenowrap`
* `distancexy`
* `distancexynowrap`
* `div`
* `dx`
* `dy`
* `exp`
* `floor`
* `int`
* `length`
* `linkheading`
* `linklength`
* `ln`
* `log`
* `max`
* `maxpxcor`
* `maxpycor`
* `mean`
* `median`
* `min`
* `minpxcor`
* `minpycor`
* `mod`
* `monitorprecision`
* `mult`
* `nanotime`
* `newseed`
* `patchsize`
* `position`
* `pow`
* `precision`
* `processors`
* `randomexponential`
* `randomfloat`
* `randomgamma`
* `randomnormal`
* `randompoisson`
* `randompxcor`
* `randompycor`
* `readfromstring`
* `remainder`
* `round`
* `scalecolor`
* `sin`
[...]

PR #970, opened 2016-03-10
Added checks for validDouble

Akarshit:
Fix for #895 
There is a validDouble check in `dz` but not in `dy` and `dx`. Any specific reason?
The list is as follows now: (crossed out items have validDouble checks)
* count
* `countother`
* `countotherwith`
* `countwith`
* ~~`minus`~~
* ~~`nsum`~~ (already had the check)
* ~~`nsum4`~~ (already had the check)
* `patchvariabledouble`
* ~~`plus`~~
* `random`
* `randomconst`
* ~~`sum`~~ (already had the check)
* `turtlevariabledouble`
* `unaryminus`
* `randomorrandomfloat`
* `abs`
* ~~`acos`~~ (already had the check)
* ~~`approximatehsb`~~
* ~~`approximatehsbold`~~
* ~~`approximatergb`~~ (already had the check)
* ~~`asin`~~  (already had the check)
* ~~`atan`~~ Review this
* `behaviorspacerunnumber`
* ~~`ceil`~~
* ~~`cos`~~
* `distance`
* `distancenowrap`
* ~~`distancexy`~~
* ~~`distancexynowrap`~~
* ~~`div`~~
* `dx`
* `dy`
* ~~`exp`~~
* `floor`
* `int`
* `length`
* `linkheading`
* `linklength`
* ~~`ln`~~  (already had the check)
* ~~`log`~~  (already had the check)
* `max`
* `maxpxcor`
* `maxpycor`
* ~~`mean`~~  (already had the check)
* ~~`median`~~  (already had the check)
* `min`
* `minpxcor`
* `minpycor`
* `mod`
* `monitorprecision`
* ~~`mult`~~
* `nanotime`
* `newseed`
* `patchsize`
* `position`
* ~~`pow`~~  (already had the check)
* `precision`
* `processors`
* ~~`randomexponential`~~  (already had the check)
* ~~`randomfloat`~~
* ~~`randomgamma`~~  (already had the check)
* ~~`randomnormal`~~  (already had the check)
* `randompoisson`
* `randompxcor`
* `randompycor`
* `readfromstring`
* ~~`remainder`~~  (already had the check)
* ~~`round`~~
* ~~`scalecolor`~~  (already had the check)
* `sin`
* `sqrt`
* ~~`standarddeviation`~~  (already had the check)
* `subtractheadings`
* ~~`tan`~~  (already had the check)
* `ticks`
* `timer`
* ~~`towards`~~  (already had the check)
* ~~`towardsnowrap`~~  (already had the check)
* ~~`towardsxy`~~  (already had the check)
* ~~`towardsxynowrap`~~  (already had the check)
* ~~`variance`~~  (already had the check)
* `worldheight`
* `worldwidth`
* `wrapcolor`
* `fileread`
* `mousexcor`
* `mouseycor`
* `hubnetinqsize`
* `hubnetoutqsize`
* `plotxmax`
* `plotxmin`
* `plotymax`
* `plotymin`
* ~~`distancexyz`~~  (already had the check)
* ~~`distancexyznowrap`~~  (already had the check)
* `dz`
* ~~`linkpitch`~~
* `maxpzcor`
* `minpzcor`
* `oheading`
* `opitch`
* `oroll`
* `oxcor`
* `oycor`
* `ozcor`
* `randompzcor`
* `randomzcor`
* ~~`towardspitch`~~  (already had the check)
* ~~`towardspitchnowrap`~~  (already had the check)
* ~~`towardspitchxyz`~~  (already had the check)
* ~~`towardspitchxyznowrap`~~  (already had the check)
* `worlddepth`

### DIFF
--- a/src/main/org/nlogo/prim/etc/_approximatehsb.scala
+++ b/src/main/org/nlogo/prim/etc/_approximatehsb.scala
@@ -16,7 +16,7 @@ class _approximatehsb extends Reporter with Pure {
         argEvalDoubleValue(context, 2));
 
   def report_1(context: Context, h: Double, s: Double, b: Double): java.lang.Double =
-    api.Color.getClosestColorNumberByHSB(h.toFloat, s.toFloat, b.toFloat, 360.0f, 100.0f, 100.0f);
+    validDouble(api.Color.getClosestColorNumberByHSB(h.toFloat, s.toFloat, b.toFloat, 360.0f, 100.0f, 100.0f))
 }
 
 class _approximatehsbold extends Reporter with Pure {
@@ -30,5 +30,5 @@ class _approximatehsbold extends Reporter with Pure {
         argEvalDoubleValue(context, 2));
 
   def report_1(context: Context, h: Double, s: Double, b: Double): java.lang.Double =
-    api.Color.getClosestColorNumberByHSB(h.toFloat, s.toFloat, b.toFloat, 255.0f, 255.0f, 255.0f);
+    validDouble(api.Color.getClosestColorNumberByHSB(h.toFloat, s.toFloat, b.toFloat, 255.0f, 255.0f, 255.0f))
 }
--- a/src/main/org/nlogo/prim/_minus.java
+++ b/src/main/org/nlogo/prim/_minus.java
@@ -25,7 +25,7 @@ public final strictfp class _minus extends Reporter implements Pure {
         argEvalDoubleValue(context, 1));
   }
 
-  public double report_1(Context context, double arg0, double arg1) {
-    return arg0 - arg1;
+  public double report_1(Context context, double arg0, double arg1) throws LogoException {
+    return validDouble(arg0 - arg1);
   }
 }
--- a/src/main/org/nlogo/prim/_plus.java
+++ b/src/main/org/nlogo/prim/_plus.java
@@ -24,7 +24,7 @@ public final strictfp class _plus extends Reporter implements Pure {
             argEvalDoubleValue(context, 1));
   }
 
-  public double report_1(Context context, double d1, double d2) {
-    return d1 + d2;
+  public double report_1(Context context, double d1, double d2) throws LogoException {
+    return validDouble(d1 + d2);
   }
 }
--- a/src/main/org/nlogo/prim/etc/_atan.java
+++ b/src/main/org/nlogo/prim/etc/_atan.java
@@ -34,7 +34,7 @@ public final strictfp class _atan extends Reporter implements Pure {
     if (d2 == 0) {
       return d1 > 0 ? 90 : 270;
     }
-    return (StrictMath.toDegrees(StrictMath.atan2(d1, d2))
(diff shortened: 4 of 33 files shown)


src/main/org/nlogo/prim/etc/_approximatehsb.scala
```diff
@@ -2,7 +2,7 @@
 
 package org.nlogo.prim.etc
 
-import org.nlogo.api, api.{ Syntax, LogoListBuilder }
+import org.nlogo.api, api.{LogoException, Syntax, LogoListBuilder}
```
Idloj: No need to import this since you don't use it anywhere

mrerrormessage:
This looks good, but I would like to see a few more things since this deals with NetLogo primitives.
1. Nice catch on the `mean` primitive. I'm going to review that PR separately, but I appreciate that it has been split out from the rest.
2. Additional tests. To test this I added the following to the bottom of test/reporters/Numbers.txt:
```
ValidateResults1
  (2 ^ 1023) * 2 => ERROR math operation produced a number too large for NetLogo

ValidateResults2
  ((2 ^ 1023) * 2) / ((2 ^ 1023) * 2) => ERROR math operation produced a number too large for NetLogo
```
To run this test, I do `./sbt testOnly org.nlogo.headless.TestReporters`.
3. Dumping generated code. To do this run `./sbt dump bench`. This tracks the bytecode generated for a handful of models so that we don't accidentally change a primitive to a really unoptimized form without noticing. It allows us to review each change to a primitive at the same level as the JVM which helps to keep everything optimized.

Once 2 and 3 have been addressed, I think this will be ready to merge.

SethTisue:
I think those `./sbt` commands won't work as written, you need double quotes to keep sbt from parsing it as separate commands.

Akarshit:
OK working on the things mentioned.

Akarshit:
@mrerrormessage does this look right?

mrerrormessage:
Yep, that looks good. Thanks again!

SethTisue:
hi, sorry, I'm late on this, but I'm skeptical that all of these checks are really needed, and I worry about the performance impact on very commonly used prims. e.g. in `_plus` and `_minus_`, is it even possible to get an invalid result if both inputs are valid...?

mrerrormessage:
@SethTisue , it is possible to get an invalid result from `_plus` or `_minus` if both inputs are valid. For instance, `1.7976931348623157E308 + 1.7976931348623157E308`. I do think we could discuss the merits of putting these checks into every prim. 

I'll go through and benchmark these changes sometime in the next day or two. If the benchmarks show a prohibitive slowdown, there are several fixes I can think of. Perhaps we just check multiplication and division; perhaps we write a compiler phase (or add something to the generator) which validates the results of numeric expressions at the outermost layer so that `1 * 2 * 3 * 4 * 5` is only checked once instead of 4 times. The latter would be my preferred solution.

SethTisue:
> validates the results of numeric expressions at the outermost layer

that sounds like best possible solution, yeah. but also sounds like work :-)

Issue #971, opened 2016-03-10
mean give overflow even for elements within range


Akarshit:
like for `show mean list (2 ^ 1023) (2 ^ 1023)` the output should be 2 ^ 1023 but instead an error comes that calculated number is too large.
The problem here is with the algo. used in mean which sums all the elements of the list(cause of overflow) and then divides it by the size of list.

mrerrormessage:
I'm closing this as you have a proposed fix in #972. In the future, if you have a bug with code to fix it, just open a PR and we'll discuss and tag appropriately on the PR. That keeps the discussion in fewer places. Issues should be opened for the following cases (and preferably only the following cases): 
1. See an issue that you would like to fix, but can't or won't right now. For example, if someone is working on a NetLogo model and discovers a bug in NetLogo, they might open an issue and come back to it when they finish working on their model.
2. Are proposing a change that requires substantial discussion before fixing. For example, if someone was to change the entire layout of NetLogo, or the model file format, or even do something like reorganizing the menu structure, I would prefer an issue first to discuss and work through the idea with the whole team before work was started and/or a PR submitted.

PR #972, opened 2016-03-10
Changed the algo to avoid overflows but lost major performance

Akarshit:
Fix for #971
But this algorithm has major performance loss.
From my testing the time taken by this was 3 time more than the time taken by the previous algorithm. 
If the overflow issues is not a major one perhaps it can be omitted.

### DIFF
--- a/src/main/org/nlogo/prim/etc/_mean.java
+++ b/src/main/org/nlogo/prim/etc/_mean.java
@@ -29,8 +29,9 @@ public final strictfp class _mean extends Reporter implements Pure {
   }
 
   public double report_1(Context context, LogoList list) throws LogoException {
-    double sum = 0;
-    if (list.isEmpty()) {
+    double mean = 0;
+    double size = list.size();
+    if (size == 0.0) {
       throw new EngineException(
         context, this, I18N.errorsJ().get("org.nlogo.prim.etc.$common.emptyList"));
     }
@@ -41,8 +42,8 @@ public final strictfp class _mean extends Reporter implements Pure {
             I18N.errorsJ().getN("org.nlogo.prim._mean.cantFindMeanOfNonNumbers",
                 Dump.logoObject(elt), TypeNames.name(elt)));
       }
-      sum += ((Double) elt).doubleValue();
+      mean += ((Double) elt).doubleValue() / size;
     }
-    return validDouble(sum / list.size());
+    return mean;
   }
 }


src/main/org/nlogo/prim/etc/_mean.java
```diff
@@ -41,8 +41,8 @@ public double report_1(Context context, LogoList list) throws LogoException {
             I18N.errorsJ().getN("org.nlogo.prim._mean.cantFindMeanOfNonNumbers",
                 Dump.logoObject(elt), TypeNames.name(elt)));
       }
-      sum += ((Double) elt).doubleValue();
+      mean += ((Double) elt).doubleValue() / list.size();
```
qiemem: You could try storing `list.size()` in a variable above and dividing by that. Although `list.size()` is fast (it resolves to [`Vector.length`](https://github.com/scala/scala/blob/v2.11.7/src/library/scala/collection/immutable/Vector.scala#L81)), tossing that extra math operation and method calls might reduce the hit (though I'm a little surprised this isn't jitted away).

nicolaspayette:
Performance of the `mean` primitive is crucial in many models.

That being said, you're code repeatedly uses `list.size()`. That operation on a Scala vector (which is the data structure backing the `LogoList`) should be effectively constant time, but it's not free either. I'd suggest trying to store the size in a local variable variable instead and see if it makes a difference.

qiemem:
Ha, great minds, etc etc.

nicolaspayette:
Seems like me and @qiemem had the same thought.

I also suspected that repeatedly dividing a double by an integer also make a difference, so I would cast `list.size()` to double when storing it to a local.

Akarshit:
@qiemem @nicolaspayette
Implementing the changes.

Idloj:
Maybe you can use the new algorithm only if you get a LogoException?

Akarshit:
@ldloj Yup, good idea.

nicolaspayette:
> I am not testing on NetLogo

Why not? For a rough approximation, I'd suggest using the [Profiler extension](https://ccl.northwestern.edu/netlogo/docs/profiler.html).

mrerrormessage:
I would like to see data on performance differences before making any decisions. With no data, I tend to like @Idloj 's idea, but it's hard to know what the actual impact is. @nicolaspayette , do you know of a good model to test against (or could you whip one up quickly)? Ideally we would use the benchmark tool in sbt, but the NetLogo profiler would also be a good way to get some data.

SethTisue:
it would be pretty awesome if somebody made a NetLogo extension that provided easy access to Thyme or ScalaMeter or what have you, so this kind of thing could trivially be tested in the Command Center (or the headless NetLogo shell)

Idloj:
As far as I know, a more efficient way to do this algorithm is by saving a local variable of `1 / list.size()` and then for each item multiply the item by this value. Multiplying Doubles is (much?) more efficient than dividing them.

Idloj:
Just tested what I've said with the following code:

tmp.java:
```
public class tmp {
  public static void main(String[] args) {
    double sum = 0.0;
    for (double i=1.0; i<1000000000.0; i+=1.0)
      sum += i / 10.0;
  }
}
```

tmp2.java:
```
public class tmp2 {
  public static void main(String[] args) {
    double sum = 0.0;
    double mult = 1.0 / 10.0;
    for (double i=1.0; i<1000000000.0; i+=1.0)
      sum += i * mult;
  }
}
```

The results are:
```
$ time java tmp; time java tmp2

real	0m8.338s
user	0m8.300s
sys	0m0.012s

real	0m4.329s
user	0m4.316s
sys	0m0.004s
```

Idloj:
After testing a few more variation, the time is shortened by 0.3s when I declare `mult` as final, and you can get the same times if you replace `sum += i / 10.0;` with `sum += i * (1.0 / 10.0)`.

Idloj:
I don't know why but when I wrote this in Scala it turned out to be much faster.
tmp.scala:
```
object tmp3 extends App {
  private[this] var sum = 0.0
  private[this] var i = 1.0
  while (i<1000000000) {
    sum += i * (1.0 / 10.0)
    i += 1.0
  }
}
```

The result is:
```
$ time scala tmp3

real	0m2.093s
user	0m2.108s
sys	0m0.044s
```

Idloj:
It turned out that the Scala code was way faster because it checked `i<1000000000` (`i` is checked against an int and not a double), yet when I did it on Java the times didn't improve.

Idloj:
Should we enforce (or very highly advise...) the use of `private[this]` on primitives' privates? @SethTisue @mrerrormessage ?

Idloj:
P.S. My suggestion of running the new algorithm only when you get an exception was based on the (reasonable, I think) assumption that the case of overflow is very very rare. That way, we don't lose *anything*, the common case will run at the exact same speed as before and the rare case will run slowly instead of erroring.

SethTisue:
@Idloj on the JVM, you won't get useful results running benchmarks that way; you need to use a proper benchmarking tool. see http://stackoverflow.com/questions/504103/how-do-i-write-a-correct-micro-benchmark-in-java

Idloj:
Sorry, but I still don't understand what's wrong about this way. The warmup and everything will still be the same for both versions. I ran everything a few times for caching of the JVM files or whatever.

SethTisue:
I'm not sure what to suggest except that you read that SO stuff more slowly and carefully, and with more willingness to believe that the people who wrote it (and who wrote tools like JMH, Caliper, and the others I mentioned) know what they're talking about, and that there's good reasons they put a lot of work into making those tools so that people can do accurate benchmarks on the JVM.

Idloj:
OK, I'll look at those.
I want to clarify, though, because it turns out that it wasn't clear and I was misunderstood, that I was not protesting or anything, but really asking. I didn't think for a moment that I understand benchmarking or the JVM better than those people. I was just saying how I see the situation in the hope to be illuminated and to understand why I am wrong.

SethTisue:
ok cool, got it. sorry I don't have time to explain

mrerrormessage:
I ran the JMH benchmarks for the two methods. Whether @Idloj 's method is faster depends on the data structure behind the iteration (but I didn't come across an instance where it was slower than the alternative, sometimes it was the same). I was initially worried that for long lists `1 / list.length` would round to an incorrect floating-point value, but that won't happen unless the list is quite large (on the order of 10^300). At this point it doesn't matter whether we use multiplication or not. I'm happy to supply the benchmarks I used on request.

Based on my benchmarking, it seems you *could* get a substantial increase in performance by using `logoList.toVector` and using an indexed loop, since in my testing indexed looping `for (int i = 0;i++)...)` on a vector is about twice as fast as iterator-based looping on that vector. It would be even more beneficial here because the vector methods on LogoList are actually proxied through an inner `Iterator` class in LogoList (so you have two method calls for every one access). Why don't you change it to use indexed looping and see if the performance improves?

Akarshit:
If the algo for the mean is changed it would have to be copied to `_satandarddeviation` also.

mrerrormessage:
Yep, it should be changed in both places. In general we should probably make indexed calls on vectors instead of iterator-based calls.

mrerrormessage:
I'm trying to close out the 5.x milestone on GitHub - @Akarshit , do you have any interest in moving this up to hexy and continuing work there?

Akarshit:
@mrerrormessage So you want me to close this PR and repoen it in hexy. Right?

mrerrormessage:
Yes, if you are still interested in working on this PR, please close and reopen it in hexy.

PR #974, opened 2016-03-12
Median not overflowing unnecessarily

Akarshit:
Similar to #972 
But the fix is easy in this case.
Also removed the `validDouble()` check as `Double / 2 + Double / 2` should be(?) a valid Double.

### DIFF
--- a/src/main/org/nlogo/prim/etc/_median.java
+++ b/src/main/org/nlogo/prim/etc/_median.java
@@ -43,8 +43,7 @@ public final strictfp class _median
     }
     Double middle1 = nums.get(medianPos - 1);
     Double middle2 = nums.get(medianPos);
-    return newValidDouble
-        ((middle1.doubleValue() + middle2.doubleValue()) / 2);
+    return newValidDouble(middle1.doubleValue() / 2 + middle2.doubleValue() / 2);
   }
 
   @Override


mrerrormessage:
While it's true that the median should never be invalid when the all input doubles are valid, we still need to keep this check in. The if one of the numbers involved in the median calculation is a `NaN` or `Infinity` the number returned will also be a `NaN` or `Infinity`. There are ways to get doubles into NetLogo which are beyond our control (extensions) and I want to keep this check so that our double primitives verify that the numbers are correct before returning them.

Akarshit:
@mrerrormessage I would add the check again to this PR only if you could open it.

mrerrormessage:
it isn't needed - `newValidDouble` performs the check and is in the existing code.

Akarshit:
But the change from `(middle1.doubleValue() + middle2.doubleValue()) / 2` to `middle1.doubleValue() / 2 + middle2.doubleValue() / 2` is still needed.

mrerrormessage:
Right, sorry I glossed right over that change. Reopening.

Akarshit:
@mrerrormessage Added the `validDouble` check again.
I will have to run the tests on this one and push those dumps too?

mrerrormessage:
Yes, that's correct. The dumps may not change for this one as I'm not sure how commonly `median` is used, but it's still a best practice when changing primitives.

Issue #960, opened 2016-03-07
Cut, Copy and Paste in the Edit menu should not be active all the time


Akarshit:
Cut/ Copy should only be active when some text is selected.
Paste should only be active when clipboard in not empty.

mrerrormessage:
Fixed in #975

PR #975, opened 2016-03-12
Cut/Copy/Paste only active when needed

Akarshit:
Fix for #960 
This also deactivates the options in the right click popup menu.


mrerrormessage:
I'm fine with the substance of this PR, but the listener that's added to the Edit should be added in `EditMenu` instead of in `App`.

Akarshit:
@mrerrormessage Absolutely correct, the listener should have been in the EditMenu class only.
I have made the changes.

mrerrormessage:
Awesome, thanks!

Issue #43, opened 2011-12-15
Linux: shell script to run app can't be run from a symlink


SethTisue:
the problem line is:

```
cd "`dirname "$0"`"
```

it's there to ensure that when NetLogo starts, the NetLogo installation directory is the current working directory, since the app expects that. but if there's a symlink to the script, `$0` points to the symlink, rather than to the target, so we end up in the wrong directory

reported by Fred Drueck

Idloj:
I know this isn't exactly a solution, but maybe providing a `.desktop` file can help, so the user doesn't even care about the `.sh` file. And it's just nice to provide a `.desktop` file!

SethTisue:
I assume a PR adding a `.desktop` file would be accepted.

Idloj:
BTW, this one will be nicer if #213 is resolved

mrerrormessage:
As of 5.3, we no longer distribute shell scripts for NetLogo applications on Linux. This is related to #979 which should bring back only the netlogo-headless script and should take symlink resolution into consideration.

PR #976, opened 2016-03-14
symlink are resolved once

Akarshit:
Fix for #43
The symlink to the script is resolve once.
Can't use `readlink -f` as `-f` is not available on OS X. But still `readlink` will resolve the symlink once.

### DIFF
--- a/dist/hubnet.sh
+++ b/dist/hubnet.sh
@@ -1,8 +1,9 @@
 #!/bin/sh
-cd "`dirname "$0"`"             # the copious quoting is for handling paths with spaces
-# -Xmx1024m                     use up to 1GB RAM (edit to increase)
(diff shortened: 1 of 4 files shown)


mrerrormessage:
Thanks for the PR.

The files that you've modified actually aren't distributed as part of the NetLogo release anymore. I'm going to close this (and #43 as well) since we don't use these scripts (I should have deleted them from the repo).

This is all very closely related to #979 and I would welcome a PR which brings back the `netlogo-headless.sh` script to Linux as in this PR. The modified script would need to take into account the changed directory structure on Linux (since the scripts you modified here were written, the location of the jars has changed). Note that the process of building for distribution is actually pretty hairy. Please ping or comment on #979 if you plan to work on it and have any questions.

PR #977, opened 2016-03-15
Nasty stack trace removed

H-Theking:


SethTisue:
this looks like it might fix it for the GUIWorkspace case, but doesn't do anything about the background-run, HeadlessWorkspace case

mrerrormessage:
@Harvey-Theking , do you have the information you need to complete this PR? Are you still interested in completing it?

Harvey-Theking:
I'm currently writing exams, so I'm not looking at any code for the moment.

mrerrormessage:
I understand. I'll check in again in about 2 weeks.

netlogo-gui/src/main/window/WorldViewSettings.java
```diff
@@ -473,6 +474,15 @@ public void setDimensions(org.nlogo.core.WorldDimensions d) {
 
   public void setDimensions(int minPxcor, int maxPxcor,
                             int minPycor, int maxPycor) {
+   if(maxPxcor - minPxcor < 0 || maxPycor - minPycor < 0 || maxPxcor < 0 || maxPycor < 0) {	
+      JOptionPane.showMessageDialog(gWidget, "The world dimensions set are invalid.\nVerify that "
```
mrerrormessage: This string should be done through `I18N.guiJ`.

mrerrormessage:
OK, this still doesn't address the main consideration in a few ways:

* It should **not** be a requirement that the max-pxcor/pycors are positive, just that they are larger than their minimum counterparts
* This doesn't produce a useful error when run headlessly. Example:
```
netlogo/runMain org.nlogo.headless.Main --model <path to nlogo file> --experiment <experiment name> --table "foo.csv" --min-pxcor 1 --max-pxcor 0 --min-pycor 1 --max-pycor 0
```
We should be returning an intelligible error for the user. This can be done (and probably should be done) in `Main.scala`, where the values are parsed and a `WorldDimension` is initialized.

If you want assistance in figuring out how to get either of these things working, please let me know.

SethTisue:
> It should not be a requirement that the max-pxcor/pycors are positive, just that they are larger than their minimum counterparts

I believe it's a NetLogo requirement that `patch 0 0` must exist, actually.

mrerrormessage:
Oops! I've edited my comment above to reflect.

netlogo-gui/src/main/window/WorldViewSettings.java
Harvey-Theking: Couldn't find guiJ. Only GUI_Strings and Errors files.

Harvey-Theking:
> We should be returning an intelligible error for the user. 

Should I use the same error as in `WorldViewSettings` or do you want something more specific?
>This can be done (and probably should be done) in Main.scala, where the values are parsed and a WorldDimension is initialized.

I tried working with it sometime. Couldn't figure out how this thing worked. Just ran it with your coordinates `--min-pxcor 1 --max-pxcor 0 --min-pycor 1 --max-pycor 0` and noticed the statements after `val dims =` are only executed if those parameters are set. Doing the checks here will only manage the case where the experiment was saved with correct parameters. If the experiment was saved with invalid parameters (though in headless I think parameters should always be specified), I think the exception will pop up without the check in the `World.java` class (unintelligible I agree). 
So this brings me back to the initial solution I was thinking about: checking for the values right when they are set in the behaviour space(GUIWorkSpace), but just can't figure how to do that. Doing it here, we won't have to bother about the headless `World` check - just the  one in `Main`.

>If you want assistance in figuring out how to get the check working, please let me know.

Its checking the `minPxcor.get.toInt, maxPxcor.get, minPycor.get.toInt, maxPycor.get.toInt` right?
What I can't yet figure out is the type of dim, accessing its attributes and Some() and...whew! Just getting started with functional programming in Scala.

netlogo-gui/src/main/window/WorldViewSettings.java
Harvey-Theking: I could only find the GUI_Strings and Errors files.

mrerrormessage:
> Couldn't find `I18N.guiJ`

Ah - sorry for the misunderstanding here. `org.nlogo.core.I18N` is the scala object used for Internationalization. `guiJ` is how the java retrieves values from the `GUI_Strings` file. So you would use something like `I18N.guiJ().get("your.key.name.here")`.

> Should I use the same error as in WorldViewSettings or do you want something more specific?

The same error as in `WorldViewSettings` is fine.

> I tried working with it sometime. Couldn't figure out how this thing worked. Just ran it with your coordinates --min-pxcor 1 --max-pxcor 0 --min-pycor 1 --max-pycor 0 and noticed the statements after val dims = are only executed if those parameters are set.

The way `Main` works is by parsing all the parameters out and then checking some of them (like `--model`, `--setup-file`, etc). The `while` loop goes through all of the arguments so the checks are only done after all have been parsed. The `if` statement following `val dims =` always has it's condition evaluated, but that condition is false unless *all* the min/max x/y cors have been specified.

> Doing the checks here will only manage the case where the experiment was saved with correct parameters.

That's correct. We should be giving the user feedback if they're running the model in an invalid way. This PR already provides good feedback to the user when saving invalid values in the UI.

Regarding where the values are set in the UI - that's done in `org.nlogo.lab.gui.ProtocolEditable`. That file has a good example of some checks done [here](https://github.com/NetLogo/NetLogo/blob/hexy/netlogo-gui/src/main/lab/gui/ProtocolEditable.scala#L84-L109). `ProtocolEditable` is the UI component, the different values are set in the `metrics` string. In this case, you'll only want to check values when set for min/max-px/ycor, but that's the right general location to look. I would prefer that the existing checks that I linked to above be pulled out into a helper function. Your checks could either be part of that function, or a separate one which is also called on the input values. I'm happy to discuss `ProtocolEditable` more if you have questions.

H-Theking:
Thanks. I'd look into it first just to get used to scala code. If I'm stuck, I'll get back to you.

mrerrormessage:
Hey, haven't heard back from you on this @H-Theking . Are you still interested in finishing it? It's no problem if you aren't, just want to know what the status of this is.

H-Theking:
I am still interested. School suddenly became very tight with continuous block lectures, so I took a break. I'd resume probably after the end of exams, by the end of this month.

Issue #540, opened 2014-02-05
set-plot-pen-color doesn't support rgb lists


qiemem:
Doing `set-plot-pen-color [ 255 128 0 ]` fails to compile saying it wanted a number, not a list.

SethTisue:
my intuition is that this would be relatively easy to fix.

mrerrormessage:
Should just be a matter of modifying the syntax object to accept a list as well as a number, then dispatching on the type of the argument, just like all the other procedures which take colors.

mrerrormessage:
Resolved in #984

PR #984, opened 2016-03-16
_setplotpencolor can now accept an array or value

pollend:
This is a pull request that resolves issue #540

### DIFF
--- a/src/main/org/nlogo/prim/plot/primitives.scala
+++ b/src/main/org/nlogo/prim/plot/primitives.scala
@@ -2,7 +2,7 @@
 
 package org.nlogo.prim.plot
 
-import org.nlogo.api.{ CommandRunnable, I18N, Syntax }
+import org.nlogo.api.{ CommandRunnable, I18N, Syntax, LogoList }
 import org.nlogo.nvm.{ Command, Context, EngineException, Instruction, Reporter }
 import org.nlogo.plot.PlotManager
 
@@ -317,12 +317,24 @@ final class _setplotpenmode extends PlotCommand(Syntax.NumberType) {
   }
 }
 
-final class _setplotpencolor extends PlotCommand(Syntax.NumberType) {
+final class _setplotpencolor extends PlotCommand(Syntax.NumberType | Syntax.ListType) {
   import org.nlogo.api.Color
   override def perform(context: Context) {
-    currentPen(context).color =
-      Color.getARGBbyPremodulatedColorNumber(
-        Color.modulateDouble(argEvalDoubleValue(context, 0)))
+    val obj = args(0).report(context)
+    obj match {
+      case rgbList: LogoList =>
+        try
+          currentPen(context).color = Color.getARGBIntByRGBAList(rgbList)
+        catch {
+          case e: ClassCastException =>
+            throw new org.nlogo.nvm.EngineException(context, this, displayName + " an rgb list must contain only numbers")
+        }
+      case c: java.lang.Double =>
+        currentPen(context).color =
+          Color.getARGBbyPremodulatedColorNumber(
+            Color.modulateDouble(c))
+      case _ => throw new org.nlogo.nvm.ArgumentTypeException(context, this, 0, Syntax.ListType | Syntax.NumberType, obj)
+    }
     context.ip = next
   }
 }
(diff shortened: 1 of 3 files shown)


src/main/org/nlogo/prim/plot/primitives.scala
```diff
@@ -317,12 +317,25 @@ final class _setplotpenmode extends PlotCommand(Syntax.NumberType) {
...
 }
 
-final class _setplotpencolor extends PlotCommand(Syntax.NumberType) {
+final class _setplotpencolor extends PlotCommand(Syntax.NumberType | Syntax.ListType) {
   import org.nlogo.api.Color
   override def perform(context: Context) {
-    currentPen(context).color =
-      Color.getARGBbyPremodulatedColorNumber(
-        Color.modulateDouble(argEvalDoubleValue(context, 0)))
+    var obj = args(0).report(context);
```
SethTisue: no semicolons please, this is Scala

src/main/org/nlogo/prim/plot/primitives.scala
```diff
@@ -2,7 +2,7 @@
 
 package org.nlogo.prim.plot
 
-import org.nlogo.api.{ CommandRunnable, I18N, Syntax }
+import org.nlogo.api.{ CommandRunnable, I18N, Syntax,LogoList }
```
SethTisue: inconsistent formatting

src/main/org/nlogo/prim/plot/primitives.scala
SethTisue: also this should be a `val` not a `var`

SethTisue: come to think of it, you don't need the `val` either. just write:

```scala
args(0).report(context) match {
  ...
```

src/main/org/nlogo/prim/plot/primitives.scala
```diff
@@ -317,12 +317,25 @@ final class _setplotpenmode extends PlotCommand(Syntax.NumberType) {
...
+        try
+          currentPen(context).color = Color.getARGBIntByRGBAList(rgbList)
+        catch {
+          case e: ClassCastException =>
+            throw new org.nlogo.nvm.EngineException(context, this, displayName + " an rgb list must contain only numbers")
+        }
+      case c: java.lang.Double => 
+        currentPen(context).color =
+          Color.getARGBbyPremodulatedColorNumber(
+          Color.modulateDouble(c))
```
SethTisue: incorrect indentation

SethTisue:
is it possible to add test coverage for this? (genuine question, I don't recall what exists in this area)

also, surely the documentation needs updating as well?

pollend:
https://ccl.northwestern.edu/netlogo/docs/dictionary.html#set-plot-pen-color

 If this is any indication, then yes the documentation needs to be updated.

src/main/org/nlogo/prim/plot/primitives.scala
SethTisue: still not fixed in the new version

SethTisue: I'm suggesting replacing:

```scala
          Color.getARGBbyPremodulatedColorNumber(
          Color.modulateDouble(c))
```

with

```scala
          Color.getARGBbyPremodulatedColorNumber(
            Color.modulateDouble(c))
```

pollend: Ahh, that makes more sense. So if that one call had multiple lines would I keep it in line  Color.modulateDouble(c)

mrerrormessage:
Thanks for the PR! At the moment the code on this branch doesn't compile and there's some trailing whitespace on 331. Additionally, the documentation will need to be updated `dist/docs/dictionary.html.mustache`. The `set-plot-pen-color` example should show "color" in italics instead of the current "number". Once these things have been addressed, I think this looks ready to go.

mrerrormessage:
OK, I ran the tests for this and we have a failing test. In `sbt`, run `testOnly org.nlogo.compiler.TestAllSyntaxes` and fix the test, then this should be good to go.

mrerrormessage:
Looks good, thanks!

PR #988, opened 2016-03-19
Fix for Return to open dialog on error

askadi:


mrerrormessage:
Looks good, thanks so much for the PR!

PR #208, opened 2016-03-22
extensions API: add a `workspace` method to `api.Context`

mrerrormessage:
this is possible now that on master, `api.Workspace` has been split off from `nvm.Workspace`

once this is done, a bunch of typecasts in the sample extensions ought to be removable I think


mrerrormessage:
It's very possible that other changes could be made to extensions based on what's been done here. Most notable (other than the issue at stake):

* Added a few controlling API methods to `api.Workspace`
* Trimmed down `nvm.Workspace` substantially. It turns out that a number of methods on `nvm.Workspace` were either part of `api.Workspace` or simply no longer used
* Loaded Plots using the loader from `org.nlogo.core.model`. This was needed because model loading in `api.Workspace` is done from the `core.Model` data structure instead of from text. This step will need to be repeated over again for the updated model format.

I'm planning to do a build of NetLogo on Thursday for @qiemem and the spring quarter NetLogo class here at the CCL. With that in mind, I would like to merge this in by tomorrow afternoon at the latest, so if there are any comments, please let me know. The release I'm planning to do tomorrow will have version number `6.0.0-M2`, so that will be the first Bintray artifact in which these changes appear.

netlogo-core/src/main/api/Workspace.scala
```diff
@@ -85,4 +85,12 @@ with ViewSettings with Controllable {
...
   def compilerTestingMode: Boolean
   def warningMessage(message: String): Boolean
+
+  @throws(classOf[CompilerException])
+  def evaluateCommands(owner: JobOwner, source: String): Unit
+  @throws(classOf[CompilerException])
+  def evaluateCommands(owner: JobOwner, source: String, waitForCompletion: Boolean): Unit
+
+  @throws(classOf[CompilerException])
+  def evaluateReporter(owner: JobOwner, source: String): AnyRef
```
qiemem: The methods `command` and `report` might be more useful for normal use. They don't require a `JobOwner`, which makes them much easier to use (though creating a `SimpleJobOwner` isn't hard, it does complicate the process unnecessarily).

mrerrormessage: `command` and `report` are found in the `Controllable` trait, which this trait extends. I think it *would* make sense to move all three of these methods into that trait as well so that all controlling API methods were kept in one place.

qiemem: Oh, I totally missed that. Awesome!

mrerrormessage:
Just a note, these changes **did not** go out in `6.0.0-M2`

mrerrormessage:
@qiemem , I'm going to merge this in today as soon as it passes Travis. Would it be helpful to you if I pushed an artifact version of this up to bintray as 6.0.0-M3?

qiemem:
bintray + http://ccl.northwestern.edu/netlogo/6.0beta/ ? That would be great for the class; it would mean students wanting to make extensions wouldn't have to redownload mid-quarter. It'd also let me clean some stuff up in LS.

SethTisue:
> students wanting to make extensions

rly?

mrerrormessage:
@qiemem , I'll see what I can do. Bintray is 5 minutes, 6.0beta is at least an hour. I'm open to it, but it may not get posted before Thursday.

qiemem:
Yup. We offer it as on option for final projects every year, though no one (unsurprisingly) has ever done it. However, this year I've heard from a few saying they're definitely interested in making one.

@arthurhjorth and I are also planning on doing a ccl-research workshop on extension creation, so it would be great for that as well.

Issue #980, opened 2016-03-15
Remember preferred variable type for Input widgets


nicolaspayette:
It has been suggested on cc-feedback that Number is a more commonly desired variable type for an Input widget than the current default of String. While this is probably true, I think the most flexible solution would be to use the Java preference store to remember the previously used type.

This could be handled in pretty much the same way that we now handle BehaviorSpace output type: https://github.com/NetLogo/NetLogo/pull/769.

Issue #682, opened 2015-01-07
issue with handling of empty numeric input boxes


DenoBeno:
Note: initially I thought the issue way with the "carefully", but it turned out that I wasn't careful enough when reporting the issue. In fact the problem was (is) with error checking on the entry field. 
--- Initial report --
I tried to use the entry fields to set up the model. Since the entry field can be empty, I tried to use "carefully" to catch such cases and set default value if nothing is provided. Something like:

to setup-turtles
  Carefully [  Create-turtles N ] [ 
       Set N 100 
      Create-turtles N
    ]
end

I tested this in a simple model which has an entry field called N. On the first try (N undefined), I get this runtime error: 

ASK expected input to be an agent or agentset but got NOBODY instead.
error while observer running ASK
  called by procedure SETUP-TURTLES
  called by procedure SETUP
  called by Button 'setup'

so basically the program comes to runtime-exception checking part and then fails with runtime exception. Nice. :-)

And it gets better: if I run the model with some number and then delete the number, the error changes into "Invalid input for a number: expected a number" exception instead. Which is probably the runtime exception check on the input form. 

hope this helps

SethTisue:
I'm unable to reproduce the problem on my own computer. I don't think you've provided all of the relevant code? The error says "ASK expected input to be an agent or agentset", but `ask` doesn't appear anywhere in the code you've provided.

digitaldust:
I tried too, but the only thing I can reproduce is that if you delete the
number from the input and then try to call setup-turtles, then you have the
expected "Invalid input for a number: expected a number" exception when you
leave the input widget.

2015-01-07 15:09 GMT+01:00 Seth Tisue <notifications@github.com>:

> I'm unable to reproduce the problem on my own computer. I don't think
> you've provided all of the relevant code? The error says "ASK expected
> input to be an agent or agentset", but ask doesn't appear anywhere in the
> code you've provided.
>
> —
> Reply to this email directly or view it on GitHub
> <https://github.com/NetLogo/NetLogo/issues/682#issuecomment-69025902>.
>



-- 
-----------------------------------------------------------------

Simone Gabbriellini, PhD

Post-doctoral Researcher
ANR founded research project "DIFFCERAM"
GEMASS, CNRS & Paris-Sorbonne.

mobile: +39 340 39 75 626
email: simone.gabbriellini@cnrs.fr

DenoBeno:
Sorry, mea culpa. Here is the simplest code I could think of:

to setup
[...]

PR #996, opened 2016-03-27
Added the inputType in Preferences

Akarshit:
Fix for #980

### DIFF
--- a/src/main/org/nlogo/window/InputBox.scala
+++ b/src/main/org/nlogo/window/InputBox.scala
@@ -18,11 +18,19 @@ import javax.swing.text.EditorKit
 import javax.swing.KeyStroke.getKeyStroke
 import javax.swing.{JDialog, JOptionPane, AbstractAction, ScrollPaneConstants, JScrollPane, JButton, JLabel}
 import javax.swing.plaf.basic.BasicButtonUI
+import java.util.prefs.Preferences
 
 abstract class InputBox(textArea:AbstractEditorArea, editDialogTextArea:AbstractEditorArea,
             compiler:CompilerServices, nextComponent:Component)
   extends SingleErrorWidget with Editable with Events.InputBoxLoseFocusEvent.Handler {
 
+  object Prefs {
+    private val prefs = Preferences.userNodeForPackage(InputBox.this.getClass)
+    def inputTypeName = prefs.get("inputTypeName", "Number")
+    def updateTo(inputTypeName: String): Unit = {
+      prefs.put("inputTypeName", inputTypeName)
+    }
+  }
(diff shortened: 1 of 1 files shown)


mrerrormessage:
Thanks for the PR, I'm planning to review this sometime tomorrow.

src/main/org/nlogo/window/InputBox.scala
```diff
@@ -18,11 +18,19 @@ import javax.swing.text.EditorKit
...
 import javax.swing.{JDialog, JOptionPane, AbstractAction, ScrollPaneConstants, JScrollPane, JButton, JLabel}
 import javax.swing.plaf.basic.BasicButtonUI
+import java.util.prefs.Preferences
 
 abstract class InputBox(textArea:AbstractEditorArea, editDialogTextArea:AbstractEditorArea,
             compiler:CompilerServices, nextComponent:Component)
   extends SingleErrorWidget with Editable with Events.InputBoxLoseFocusEvent.Handler {
 
+  object Prefs {
+    private val prefs = Preferences.userNodeForPackage(InputBox.this.getClass)
```
mrerrormessage: `InputBox.this.getClass` is correct, but it would be more idiomatic scala to write `classOf[InputBox]` since the class is known at compile-time. `classOf[X]` is the scala parallel to `X.class` in java.

mrerrormessage:
This adds the enhancement requested by #980 , but it isn't quite ready for merge. The default behavior here is to have the input box start as a number. When a number input box is added without having been set to "string" at some point, an error is displayed to the user "Expected a Number" when closing the edit dialog, before a user even has a chance to enter a number. This same error is not displayed if the input box was of the string type first, even if only briefly. There are two possible fixes here:

* Change the default input box type back to string. Although #980 suggests that the number inputbox is more commonly used, I don't think keeping the default as-is causes any harm.
* Figure out how to make this default to a sensible initial value for each input type so that the first input box created doesn't error, no matter what the type preference is set to.

How you would like to address the problem is up to you. I also had one very minor stylistic comment, but the rest of the PR looks good. Thanks again!

Akarshit:
@mrerrormessage 
Do you think that the second point you mention is same as what has been discussed in #682.
I think if that issue is fixed then this PR can be merged without any changes.

mrerrormessage:
I'm closing out all PRs on the 5.x branch. Are you interested in continuing to work on this PR on hexy?

Akarshit:
Yes I would port the PRs there. Just a little busy this week.

mrerrormessage:
OK, thanks!

mrerrormessage:
Closing as this is 5.x

PR #979, opened 2016-03-28
Bring back `netlogo-headless.sh` in Linux version

mrerrormessage:
Improved packaging in 5.3.0+ has made other `.sh` scripts unnecessary, but that one would still be useful.


mrerrormessage:
This will be covered separately in hexy (probably no PR).

PR #985, opened 2016-03-28
Windows file associations give options when multiple versions of NetLogo installed

mrerrormessage:
Will need to modify the NetLogo wxs. See this post: http://stackoverflow.com/questions/30005173/how-to-retain-file-association-of-previous-version-after-removing-new-version-in


mrerrormessage:
This has already been merged on hexy.

PR #1003, opened 2016-03-30
Fix a couple old-style Femto.get calls

qiemem:
I couldn't find any others via simple grepping, though it's not the easiest thing in the world to grep for, since `Femto` calls are often spread across multiple lines, and the `Array` argument may or may not be constructed inline.

Anyway, you may wonder why anyone would care about `LiteWorkspace` these day. The answer is, as usual, because LevelSpace.


SethTisue:
> Anyway, you may wonder why anyone would care about LiteWorkspace these day

embedding API users

PR #1006, opened 2016-04-04
Preserve extension exception causes

qiemem:


SethTisue:
perhaps this is an opportune time to convert `ExtensionException` to Scala?

netlogo-core/src/main/api/ExtensionException.scala
```diff
@@ -3,9 +3,7 @@
 package org.nlogo.api
 
 /** Wrapper class for exceptions thrown by NetLogo extensions. */
-class ExtensionException(message: String, cause: Exception) extends Exception(message) {
-  if (cause != null)
-    setStackTrace(cause.getStackTrace)
+class ExtensionException(message: String, cause: Exception) extends Exception(message, cause) {
```
mrerrormessage: :red_circle: that was easy!

mrerrormessage:
This looks good, just two notes:

1. I would support converting `EngineException` to scala, as Seth mentioned, provided it didn't bulk up this PR too much. If you end up having to change a lot of Java files in the process, let's keep that as a separate PR.
2. This needs to pass Travis. I don't care whether that's done by making the added test work in 3D, or by disabling it in 3D, that's up to you.

Thanks for the PR! Looking forward to getting this change in.

qiemem:
> perhaps this is an opportune time to convert `EngineException` to Scala?

We decided against this as `EngineException` makes use of null in public fields, and changing that would be more substantial than what we probably want for this PR.

mrerrormessage:
Looks good, thanks!

PR #1010, opened 2016-04-07
Copy only the error message

Akarshit:
The message "NetLogo is unable to supply you with more details about this error.  Please report the problem at https://github.com/NetLogo/NetLogo/issues, or to bugs@ccl.northwestern.edu, and paste the contents of this window into your report." does not get copied when clicking the copy button.

### DIFF
--- a/src/main/org/nlogo/window/RuntimeErrorDialog.java
+++ b/src/main/org/nlogo/window/RuntimeErrorDialog.java
@@ -19,6 +19,7 @@ public strictfp class RuntimeErrorDialog
   private String eventTrace;
   private final String className;
   private String threadName;
+  private String extra;
 
   private String textWithoutDetails;
   private String textWithDetails;
@@ -120,7 +121,8 @@ public strictfp class RuntimeErrorDialog
       copyButton.addActionListener
           (new java.awt.event.ActionListener() {
             public void actionPerformed(java.awt.event.ActionEvent e) {
-              textArea.selectAll();
+              int beginIndex = textArea.getText().indexOf(extra) + extra.length();
+              textArea.select(beginIndex, textArea.getText().length());
               textArea.copy();
               textArea.setCaretPosition(0);
             }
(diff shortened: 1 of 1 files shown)


mrerrormessage:
Thanks for the PR, I'll review tomorrow.

mrerrormessage:
Thanks for the PR! It works as described, but it that isn't always desirable.

Sometimes the first part of an error message contains information that's would be really valuable in figuring out the problem. For instance if I type `runresult "null"` into the command center, I see the following error:
> Nothing named NULL has been defined
> error while observer running RUNRESULT
>   called by Command Center
>
> org.nlogo.nvm.EngineException: Nothing named NULL has been defined
> at org.nlogo.prim.etc._runresult.report(_runresult.scala:34)
> ...
Under this PR, the first part of that (which could be helpful in debugging) isn't copied. The best thing to do would be to truncate the first part of the message only if we know the message is: "NetLogo is unable to supply you with more details about this error...". Does that sound alright to you?

SethTisue:
```scala
scala> "foobar".stripPrefix("foo")
res1: String = bar

scala> "bazbar".stripPrefix("foo")
res2: String = bazaar
```

Akarshit:
Thanks @mrerrormessage for pointing out the missing case and @SethTisue for the idea.
I have changed the commit and the concept now used is similar to what @SethTisue mentioned.
Please mention if i am still neglecting any use case.

src/main/org/nlogo/window/RuntimeErrorDialog.java
```diff
@@ -19,6 +19,7 @@
   private String eventTrace;
   private final String className;
   private String threadName;
+  String extra;
```
mrerrormessage: please make `private`.

mrerrormessage:
Looks good, just one comment and then we can merge this in.

Akarshit:
@mrerrormessage Done.

PR #1014, opened 2016-04-14
Make quick help work beyond first two words

qiemem:


parser-core/src/main/lex/Tokenizer.scala
```diff
@@ -19,7 +19,9 @@ object Tokenizer extends TokenizerInterface {
       Seq(TokenType.Ident, TokenType.Command, TokenType.Keyword, TokenType.Reporter)
     tokenizeString(source)
       .sliding(2)
-      .find(ts => ts.head.start <= position && ts.head.tpe != TokenType.Eof)
```
mrerrormessage: well, I feel :sheep: -ish

mrerrormessage:
Will merge as soon as Travis passes. We can also discuss an M5 to get this and some other small fixes out ASAP, if the TAs or Uri would like.

PR #1012, opened 2016-04-15
Error in my-links (but not in web version)

mrerrormessage:
http://www.netlogoweb.org/web?http://m.modelling4all.org/p/9oJf503_AM446-E7xFl25g.nlogo

runs fine in FireFox but in 5.3.1 it stops with the error

LINKS is a directed breed.
error while object 8 running MY-LINKS
  called by (command task from: procedure -HOUSES-39)
  called by procedure RUN-PROCEDURE
  called by procedure DO-EVERY-INTERNAL
  called by procedure RUN-PROCEDURE
  called by (command task from: procedure START-TICK)
  called by procedure START-TICK
  called by procedure GO-UNTIL
  called by procedure GO
  called by Button 'GO'


mrerrormessage:
PR accidentally opened onto wrong branch, follow issue on #1017. Should be fixed in the next major release.

PR #1020, opened 2016-04-16
Merged the comment uncomment into one.

Akarshit:
In most of the editors there are no separate shortcuts for comment uncomment.
There is a single shortcut that toggles the line. This is implemented in the PR.
If multiple lines are selected then all are commented(one more `;`) if even a single one is not already commented(this is the behavior that i found in eclipse, intelliJ, etc.).
I have changed the language files but would need someone to contribute to them(or should i just google the translation and put it?).

### DIFF
--- a/src/main/org/nlogo/app/EditMenu.scala
+++ b/src/main/org/nlogo/app/EditMenu.scala
@@ -40,8 +40,7 @@ with org.nlogo.window.Events.LoadSectionEvent.Handler
   addMenuItem(I18N.gui("shiftRight"), ']', org.nlogo.editor.Actions.shiftRightAction)
   addMenuItem(I18N.gui("format"), (java.awt.event.KeyEvent.VK_TAB).toChar, org.nlogo.editor.Actions.tabKeyAction, false)
   addSeparator()
-  addMenuItem(I18N.gui("comment"), ';', org.nlogo.editor.Actions.commentAction)
-  addMenuItem(I18N.gui("uncomment"), ';', true, org.nlogo.editor.Actions.uncommentAction)
+  addMenuItem(I18N.gui("comment") + " / " + I18N.gui("uncomment"), ';', org.nlogo.editor.Actions.commentToggleAction)
   addSeparator()
   snapper = addCheckBoxMenuItem(I18N.gui("snapToGrid"), app.workspace.snapOn(), snapAction)
 
--- a/src/main/org/nlogo/editor/Actions.scala
+++ b/src/main/org/nlogo/editor/Actions.scala
@@ -9,8 +9,7 @@ import javax.swing.text.DefaultEditorKit.{CutAction, CopyAction, PasteAction, In
 
 object Actions {
 
-  val commentAction = new CommentAction()
-  val uncommentAction = new UncommentAction()
+  val commentToggleAction = new CommentToggleAction()
   val shiftLeftAction = new ShiftLeftAction()
   val shiftRightAction = new ShiftRightAction()
   val tabKeyAction = new TabKeyAction()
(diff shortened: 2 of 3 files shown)


Idloj:
I think the translation should be left intact, the users would better see the text untranslated than something that doesn't sound right in their language (although Google Translate can be good enough for one word, I still think we should avoid it).

Akarshit:
@Idloj Yup that makes sense.
I have removed the i18n for `toggleComment`.

Idloj:
Come to think of it, if you really want you can look for the phrase "Toggle Comment" in some other software and see how they have translated it.

mrerrormessage:
From a technical perspective, I haven't thoroughly reviewed this PR but I don't have any major problems with it.

From a user interface perspective, we're going to need to discuss this around the CCL and get a bit of consensus before making this change. Anyone else watching this PR who has strong feelings on "Toggle Comment" vs "comment" / "uncomment" please send in any thoughts by this Friday.

Pending the results of our discussion on Friday, I'll give this PR a full review.

qiemem:
I would be in favor of this, unless there are strong objections.

nicolaspayette:
I personally would like this as well.

A lot of our users, however, still use _double_ semi-colons for comments. As far as I know, they serve no purpose, but I suspect it's a deeply ingrained habit. At the moment, you can get `;;` by pressing <kbd>Ctrl</kbd>+<kbd>;</kbd> twice. If the shortcut was a toggle, you couldn't do that any more.

Idloj:
> If the shortcut was a toggle, you couldn't do that any more.

and they will use the new style... (the Coding Guidelines wiki states that "new models should just use one.")

SethTisue:
what happens if some of the lines I select are commented and others not?

Akarshit:
@SethTisue then all the lines get commented.
Those that were already commented get commented again(they get two `;`). This is the behavior that i found in almost all the editors.

dist/i18n/GUI_Strings_es.txt
```diff
@@ -61,8 +61,6 @@ menu.edit.find = Buscar...
 menu.edit.findNext = Buscar siguiente...
 menu.edit.shiftLeft = Des-Sangrar texto
 menu.edit.shiftRight = Sangrar texto
-menu.edit.comment = Comentarios al texto
-menu.edit.uncomment = Descomentar
```
mrerrormessage: Instead of losing these translations altogether, can we do just say "Comment" / "Uncomment". For instance here we could add `menu.edit.toggleComment = Comentarios al texto / Descomentar`, and similarly for the other files.

mrerrormessage:
This looks great! I just had one comment and then I think this is ready for merge.

Akarshit:
@mrerrormessage Done, thanks.

src/main/org/nlogo/app/EditMenu.scala
```diff
@@ -40,8 +40,7 @@ with org.nlogo.window.Events.LoadSectionEvent.Handler
   addMenuItem(I18N.gui("shiftRight"), ']', org.nlogo.editor.Actions.shiftRightAction)
   addMenuItem(I18N.gui("format"), (java.awt.event.KeyEvent.VK_TAB).toChar, org.nlogo.editor.Actions.tabKeyAction, false)
   addSeparator()
-  addMenuItem(I18N.gui("comment"), ';', org.nlogo.editor.Actions.commentAction)
-  addMenuItem(I18N.gui("uncomment"), ';', true, org.nlogo.editor.Actions.uncommentAction)
+  addMenuItem(I18N.gui("comment") + "/" + I18N.gui("uncomment"), ';', org.nlogo.editor.Actions.commentToggleAction)
```
Idloj: Isn't it better to have it the way @mrerrormessage suggested (i.e. a "toggleComment" key) ?
(Anyway, there should be spaces surrounding the slash)

mrerrormessage: I did intend to ask for a "toggleComment" key, and it should definitely have spaces around the slash - sorry if that wasn't clear.

Akarshit:
@mrerrormessage @Idloj Sorry for not including the spaces. I have included them now.

mrerrormessage:

Akarshit:
@mrerrormessage I just noticed that this is not present in `hexy`. Aren't all the features from `Netlogo:5.x` ported to `hexy`?

mrerrormessage:
Oops! This one didn't make it in. I'll port it over today.

PR #1021, opened 2016-04-18
changed xor implementaion to consider one condition

Akarshit:


Idloj:
This line in "Programming Guide" should be updated (found in the section "Syntax", subsection "Compared to other Logos"):
> The and and or reporters are special forms, not ordinary functions, and they "short circuit", that is, they only evaluate their second input if necessary.

The documentation for the xor primitive should be updated, too.

mrerrormessage:
@Idloj , I don't understand your comment. `xor` requires you to evaluate both arguments and is not implemented as a special form, so I don't see any need to update documentation. Is there something I'm missing?

Idloj:
Oh right, silly me.

mrerrormessage:
Looks good. Thanks!

PR #1018, opened 2016-04-18
M4: Recompiling deletes all turtles

mrerrormessage:
To reproduce: Open any model with turtles. Press `setup` then `go`. Switch to the code tab. Make any modification (just put a space at the end of a line or something). Switch back to the interface. All turtles will be gone.

### DIFF
--- a/netlogo-gui/src/test/agent/WorldTests.scala
+++ b/netlogo-gui/src/test/agent/WorldTests.scala
@@ -3,7 +3,9 @@
 package org.nlogo.agent
 
 import org.scalatest.FunSuite
-import org.nlogo.core.WorldDimensions
+import org.nlogo.core.{ Breed, Program, WorldDimensions }
+
+import scala.collection.immutable.ListMap
 
 class WorldTests extends FunSuite with AbstractTestWorld {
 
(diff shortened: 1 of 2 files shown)


mrerrormessage:
Mutable state is the devil. Let me know if you think any further tests would be helpful here.

netlogo-gui/src/test/agent/WorldTests.scala
```diff
@@ -24,6 +26,30 @@ class WorldTests extends FunSuite with AbstractTestWorld {
...
+      createPatches(dimensions)
+      realloc()
+    }
+    w.program(program)
+    w.realloc()
+    w
+  }
+
+  def changeProgram(w: World, program: Program) = {
+    w.rememberOldProgram()
```
qiemem: Does something actually test that this is called during recompile?

qiemem:
I'm not actually sure what's supposed to happen when a breed is removed during recompile, but adding tests for that (if they don't already exist) might be a good idea as well. Otherwise, looks great!

Indeed, mutable state is in fact the devil.

netlogo-gui/src/test/agent/WorldTests.scala
mrerrormessage: That's done by `CompilerManager` and it's actually called immediately following a successful compilation (so subsequent compilations don't have to worry about it). I'll see if I can figure out a way to test it tomorrow morning, but I'm a bit unsure whether I can get something to work since it's heavily entwined with the awt event thread.

qiemem: > I'm a bit unsure whether I can get something to work since it's heavily entwined with the awt event thread.

Ew. Anyway, if it's trouble, don't worry about it.

mrerrormessage: I have a few simple tests working on a branch, but I'm going to make a note to go ahead and merge and come back to those tests later for a few reasons:
1. The tests depend on `CompilerManager` being converted to scala
2. I want to get a release out today
I've made myself a note to get those tests merged in and I think they will be quite helpful on the new file format branch, as `CompilerManager` will likely experience some changes on that branch.

qiemem: Sounds good.

SethTisue:
> I'm not actually sure what's supposed to happen when a breed is removed during recompile

I would expect turtles of that breed to become unbreeded.

qiemem:
> I would expect turtles of that breed to become unbreeded.

That could silently cause problems though. Just removing them might be safer. Not sure.

Removing them appears to be the behavior in 5.3.1.

PR #1033, opened 2016-05-02
Allow defined primitive names to be used as symbols

mrerrormessage:
Posting for review. @qiemem , please note that this still doesn't cover the case where the would-be symbol takes arguments. For instance `turtles` now works, but `turtle 0` doesn't.


qiemem:
> For instance `turtles` now works, but `turtle 0` doesn't.

I would expect `turtle` to be a single symbol, but not `turtle 0`.

mrerrormessage:
That makes sense and should work correctly. The test defining this behavior is [here](https://github.com/NetLogo/NetLogo/pull/1033/files#diff-0fe4e4eec0bfb5e1069f084e3e401fb6R121)

qiemem:
I'm suggesting that:

```
__symbol turtle => "turtle"
```

should pass, which it doesn't currently appear to.

mrerrormessage:
Just changed. `__symbol turtle` now works.

qiemem:
Woohoo! lgtm then.

PR #780, opened 2016-05-02
Unbundle qtj extension

mrerrormessage:
Apple is no longer including QTJava.zip with Yosemite. This means that continuing to bundle the qtj extension is a recipe for lots of problems in both building and supporting NetLogo.


mrerrormessage:
wrong branch, reopening PR.

Issue #849, opened 2015-09-17
Improved ".nlogo" file format


mrerrormessage:
Ongoing discussion about whether this should be in yaml or xml.

qiemem:
I believe we left off an in-person meeting with "let's create actual examples of different possibilities and compare" rather than just speculatively arguing the abstract merits of different formats.

mrerrormessage:
That's correct. I was going to update this ticket and ask both you and @frankduncan to post representative examples in yaml and xml, respectively.

Criteria:

* Must contain all fields currently in a NetLogo file
* Must be editable in a text editor

Other than that, I think it's a question of format and relative merit. Whatever it is that needs to be done, we'll get it done.

Idloj:
That's a cool quote for life, @mrerrormessage: "Whatever it is that needs to be done, we'll get it done."

mrerrormessage:
[Translated a representative section of WSP into xml this morning](https://gist.github.com/mrerrormessage/b5476c3e4c887fb45f4c ), I would appreciate feedback if @qiemem or @frankduncan have time. Leave any comments on the gist, please, and I will update the file if there are any suggestions. Also, if anyone would care to write up an alternate format I can add it to the gist (or the gist can be forked) for a side-by-side comparison.

alan-isaac:
Is it still the intention for these changes to include allowing users to hide/remove the View via the Settings dialogue, as suggested by Seth long ago?

SethTisue:
@alan-isaac that doesn't seem on-topic for this ticket?

SethTisue:
ah, I see. yes, #849 going forward would unblock #681, but no, that doesn't mean that #849 going forward would automatically mean #681 would happen as well. these are separate, only tangentially related issues.

SethTisue:
some notes on a possible new format: https://github.com/NetLogo/NetLogo/wiki/The-nlogox-format

ERSUCC:
Models are now saved as XML as of NetLogo 7. Closing the issue as completed.

PR #1043, opened 2016-05-13
Clean up model loading

mrerrormessage:
Still a few changes left to be made, but this encapsulates the basic ideas. This is a necessary step towards #849, but doesn't actually change the file format. Instead, it detaches serialization code from other concerns (like the UI) to allow serialization code to change independently. As part of the changeover, a number of major UI classes have been converted from Java to Scala.

I want to highlight a few directional changes that are implicit in this PR but might be missed by a quick skim, as they haven't been universally applied:

* Use `java.net.URI` to communicate the location of the model as opposed to a simple string. Strings are basically awful, and you don't know what they're relative to. Additionally, as NetLogo moves into the 21st century it would be nice at some point to be able to open a NetLogo model from the world-wide-web, say from the Modelling Commons or Community Models Library.
* `org.nlogo.core.Model` now has a very central role of communicating data about the model. It has a list of optional values tagged with a string and a type, as well as the following fields which all NetLogo models must define.
  * code
  * info
  * interface
  * turtle shapes
  * link shapes
* Several of the traits formerly in `AbstractWorkspaceTraits` have been moved to separate files. This was done because the file was becoming clumsily long (500+ lines) and the compiler errors were starting to look strange (no line numbers, just referencing `AbstractWorkspaceTraits`).


nicolaspayette:
I only eye-balled the diffs, but this is _very_ nice! The new `org.nlogo.core.Model` will allow significant improvements to the model testing code.

PR #1034, opened 2016-05-18
Glass L&F improvement (6.0-M6)

mrerrormessage:
Meta-issue for all Glass L&F issues in 6.0-M6. Please comment below and we'll keep a list of tasks here. __Please include a screenshot if possible__ (annotated ones are even better).

- [x] Consistent button borders
- [x] Make weird border thing less weird, or border-y, or less of a thing
- [ ] Focused, active JButtons should have white text


mrerrormessage:
See #1045.

PR #1047, opened 2016-05-18
Fix copying of ProbLab models and copy Oil Cartel HubNet to HubNet Activities

nicolaspayette:


mrerrormessage:
Looks good, thanks!

PR #1064, opened 2016-05-26
Remove `isDirectory` check on `.git` when converting a model to normal mode

nicolaspayette:
With the new project structure, `.git` is a file pointing to `../.git/modules/models` instead of being a directory.


mrerrormessage:
Looks good here, I'll merge this as soon as Travis passes.

PR #695, opened 2016-05-31
String representation of tasks

mrerrormessage:
IT would be great to have a way of returning a string representation of a task. I think ideally it would be the default reporter value for a task. Currently if I write: `let atask task [ask turtles [fd 1]] show atask`, the `show` gives me:

    observer> let atask task [ask turtles [fd 1]] show atask
    observer: (command task from: Command Center)

Rather, I'd want it to return something like: 

    observer> let atask task [ask turtles [fd 1]] show atask
    observer: (command task from: Command Center: [ask turtles [fd 1]])

I guess alternatively we could introduce a new reporter, e.g. `show-task`, that takes a task as an argument and reports only the actual text. That [i]might[/i] be useful for models where tasks are generated and people want to dynamically check (or change) tasks.


mrerrormessage:
Sample of the changes involved can be found [here](https://github.com/NetLogo/NetLogo/pull/695/files#diff-f0da6dbe662e3fa38f6f489534a4c31cR1). Does this change match what you were looking for @arthurhjorth ? I like this change overall and I think it makes showing primitives work more "as expected."

qiemem:
Looks awesome, Robert. One thing: Currently, extensions can return task-subclasses to allow arbitrary Java code to be run from tasks. `cf` is one example. In those cases, what should/does happen? Related: #948

arthurhjorth:
That looks great, thanks so much @mrerrormessage !

I was going to ask whether this also fixed #755 but then I checked on it and saw that you'd already done that and asked me about it. Oops, sorry about that. 

@qiemem do you have an old version of LS with `to-otpl` lying around so we can check #755? Alternatively, do we want `to-otpl` as an underscoreunderscore prim? Maybe this is the wrong place to have this conversation. If so, we can move it somewhere more appropriate.

mrerrormessage:
I would prefer to have any conversation about `to-otpl` take place in a separate thread (feel free to open a new issue if you need work done on NetLogo). I'll go ahead and merge this in. Thanks for taking a look!

qiemem:
> @qiemem do you have an old version of LS with to-otpl lying around so we can check #755? 

Well, we both do, in our git histories, but it won't work with 6.0.

I agree with Robert that the to-OTPL discussions belongs elsewhere.

qiemem:
Apologies for commenting on a closed issue, but I wonder if an evaluatable representation would be better. E.g.:

```
(task [ "foo" ])
```
instead of
```
(reporter task: [ "foo" ])
```

PR #948, opened 2016-06-02
Allow `run` and `runresult` to run `api.Task`s

mrerrormessage:
The purpose of this is to allow extensions to easily generate tasks that wrap JVM code that can be used by the rest of NetLogo. Currently, you have to instantiate `nvm.*Task`  with bogus data. It's not a lot of code, but it gross. The CF extension currently does it here: https://github.com/qiemem/ControlFlowExtension/blob/0e703c8c47a406cd8364d6749f9ae417b8bf8e2a/src/main/CFExtension.scala#L24

It would be nice if extensions just had to override `api.ReporterTask.report`/`api.CommandTask.perform`.

### DIFF
--- a/netlogo-core/src/main/api/Task.scala
+++ b/netlogo-core/src/main/api/Task.scala
@@ -2,10 +2,24 @@
 
 package org.nlogo.api
 
-sealed trait Task
+import org.nlogo.core.Syntax
+
+sealed trait Task {
+  /**
+   * Used to specify the number of arguments required to run a task.
+   * Note that a task may be supplied with more arguments depending
+   * on the number of arguments supplied to run/runresult.
+   * While this is only used to calculate argument count at the moment,
+   * future versions may check whether the task is variadic and may
+   * make use of the type information to provide error messages.
+   */
+  def syntax: Syntax
+}
+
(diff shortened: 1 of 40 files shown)


netlogo-core/src/main/api/Task.scala
```diff
@@ -2,10 +2,58 @@
...
+    * they accept. For instance, the task returned by `task [?1 + ?2]`
+    * can be run with 10 arguments and the last 8 will be ignored.
+    * The current versions only makes assertions that the number of arguments is
+    * greater than the syntax minimum.
+    *
+    * Future versions ''may'' make greater use of the information provided by
+    * `syntax`. Primitives returning tasks will want to ensure the task's syntax is
+    * not more restrictive than the expected argument(s).
+    */
+  def syntax: Syntax
```
qiemem: I realize this is already merged, but as I'm working with this now, a default implementation of `Syntax.reporterSyntax(ret = Syntax.WildcardType)` would be nice.

mrerrormessage: So that would need to go on the `ReporterTask` trait. It would work really well for scala extensions, but I think would be confusing for java extensions.

Issue #681, opened 2015-01-05
View visibility changes


TheBizzle:
Alan Isaac writes:

> One can hide the Graphics Window by setting size width to a negative number, but this is undocumented and clunky.  Please consider the following.
>
> 1. A `View` menu with a hide/show switch for the Graphics Window. (Some `Tools`, such as Hiding/showing the Command Center, could be moved there as well.)
>
> 2. A Boolean primitive to accomplish this (e.g., `display-graphics-window`). This should be usable in the `startup` procedure.
>
> More radically, since some models do not need a graphics window, one might introduce a Boolean primitive to allow disabling its creation (in the startup procedure).

TheBizzle:
I agree with the general sentiment of this idea.  I think it's odd that NetLogo forces models to have a View, because it's not true that all models _use_ the View—the vast majority, yes, but not all.  Some examples of models that don't need a View are those concerned with just plots, just textual output (be it into the Command Center or some other widget), or simply monitoring readings from hardware devices.

Regarding adding primitives for controlling the visibility of the View, I have some opposition to that—albeit not strong opposition.  My opinion is that we should try to keep the language primitives focused on the objects that live _within_ the model (e.g. agentsets and variables—stuff used for computer new model states), rather than having primitives for introspecting the model's settings.

Instead, I think that the View should be there by default (like it currently is), but that users should just be able to delete the View widget like any other widget, and be able to have at most one View in their model (like how we treat Output widgets).  If a View is present, we draw to it on tick (or when it is created), otherwise not.

It's not clear to me what use case(s) exist for hiding/showing the View dynamically throughout the run of the same model.

nicolaspayette:
Related: #326, #409.

SethTisue:
if this is added, I'd suggest making it a checkbox in the Settings dialog.

TheBizzle:
Bumping and assigning to the Hexy milestone.

Rather than relying on kludges with negative patch sizes or hiding the View behind other widgets, I continue to believe that we should embrace the fact that the View is just another widget, and it should be addable/removable like other widgets.

It can—and probably should—still be there by default when a "blank" model is started, but I think it would make sense to be able to remove it like other widgets, and to add it back later using the standard widget-adding controls.  A model should only be permitted to have one view widget at most (like how the output widget is handled).

mrerrormessage:
Also related #542

SethTisue:
I like Jason's suggestion better than mine.

[...]

PR #930, opened 2016-06-06
Remove View resizing arrows

mrerrormessage:
And the 3D view button. And the ticks counter (move it under the speed slider). Get rid of the whole view top bar.


mrerrormessage:
Here's a side-by-side comparison (before on the left, after on the right). A few things to note:

* The "after" version uses about 5-6 more vertical pixels
* If we got rid of the weird bezel around the interface panel we could probably save 4 of those pixels
* The View is smaller, so there's a net gain in vertical real estate (at least for models that use views).
* Moving the tick counter will be necessary for #681 
* I haven't yet added "open in 3D view" to the right-click menu for the view. I wanted to open it up for feedback before I started working on that.

screen shot 2016-06-06 at 2 55 04 pm

There are no visible differences within NetLogo 3D.

mrerrormessage:
OK, I've now removed the weird bezel around the interface panel. The resulting side-by-side looks like this:
screen shot 2016-06-06 at 3 45 41 pm

nicolaspayette:
Under Linux, there is a weird line at the bottom of the toolbar that is somehow interrupted between the first separator and the right of the "Settings..." button:

weirdline

Also, it's hard to see from the screenshots, but the various borders are all slightly different sizes, which gives it an unbalanced look:

Top: 6px
Left 6px
Right: 4px
Bottom: 5px

Here are the corners, zoomed in 8x:

topleft bottomright

That being said, I find the overall change to be a significant improvement. In my view: the less clutter, the better.

mrerrormessage:
Thanks for checking it out on linux. I'll tighten up the view corners and see if I can figure out what's going on with the bottom border below the speed slider.

netlogo-gui/src/main/app/InterfaceTab.scala
```diff
@@ -31,6 +31,7 @@ class InterfaceTab(workspace: GUIWorkspace,
     // way so that only a vertical scrollbar is really needed - ST 7/13/04
     ScrollPaneConstants.VERTICAL_SCROLLBAR_ALWAYS,
     ScrollPaneConstants.HORIZONTAL_SCROLLBAR_AS_NEEDED)
+  scrollPane.setBorder(new javax.swing.border.EmptyBorder(0, 0, 0, 0))
```
Idloj: I guess `BorderFactory.createEmptyBorder` is nicer?

http://stackoverflow.com/a/3333696

mrerrormessage: That's a good point. I'm finishing up some changes to get this looking right in linux, I'll have this use BorderFactory as well.

Idloj: I should also point out that the Interface tab toolbar's height is inconsistent with the Info tab and the Code tab toolbars (on Linux with Nimbus LaF, anyway)

mrerrormessage: Yeah, I've noticed this height inconsistency before. It isn't ideal, but I don't think it's worth the time it would take me to fix it.

mrerrormessage:
OK, this is the result after the latest round of changes:
screen shot 2016-06-07 at 12 39 59 pm

@nicolaspayette , I was unfortunately unable to fix the linux toolbar border issue you're seeing. There are a lot of components involved and despite trying for over an hour I couldn't get the border to appear. I'm willing to invest more time at some point, but I don't think we have enough linux users for me to justify the time right now.

mrerrormessage:
@nicolaspayette you brought up on Gitter that the code here shrinks the saved size of the view by 2 pixels. I'm also seeing this for some other widgets, are you also seeing that? It turns out we save the size of widgets including borders (which I would really like to change), but changing that will require some pretty substantial changes to widget code. Do you have any thoughts?

nicolaspayette:
> I'm also seeing this for some other widgets, are you also seeing that? 

I haven't seen that, but I've only saved a couple of models using this branch. Maybe it's only specific types of widgets (which the models I tried didn't happen to have)?

> It turns out we save the size of widgets including borders (which I would really like to change), but changing that will require some pretty substantial changes to widget code. Do you have any thoughts?

Hum... why would you _not_ include borders?

From an outside perspective, if you want to know how much space a widget takes up, having the border included makes sense. Let's say you want to edit the model file "by hand" (like some users actually do) and leave exactly 5px of white space between each widget: not having the border included would require you to know how wide it is and take it into account in your calculations.

mrerrormessage:
I hadn't considered that people edit NetLogo model files by hand. I guess I had assumed that since borders are / were not editable by users that they shouldn't be included in the user save file. However, since people do use them, it makes sense to keep things as they are. I'll go through and check this only takes off 2 pixels from view widgets, then go ahead and merge.

nicolaspayette:
I just noticed that, in NetLogo 3D, the ticks counter is left of the speed slider:

3dview

Maybe it doesn't matter, but it could be argued that 2D and 3D should be consistent...

mrerrormessage:
@nicolaspayette , I think this is ready to merge, but I wanted to get the go-ahead from you since this might break the models library build.

nicolaspayette:
:+1: 

> this might break the models library build

No problem! I will adjust my tests for the new view dimensions.

Also: :tada:. I like the new look a lot.

qiemem:
Slight problem: kiosk users.

Kiosk mode doesn't include the bar at the top of the window. The speed slider as on the view control strip with the tick counter.

Maybe a small bar at the top of the window?

image

SethTisue:
I think it was an error, both conceptually and pragmatically, to move the tick counter out of the model interface and into the toolbar. it means the ticks won't show up in `export-interface` and `movie-grab-interface`, which is a disaster. The tick counter is properly part of the model UI.

mrerrormessage:
I think it was the right choice. Not all models use ticks (Termites, for instance), for many others the exact number of ticks isn't relevant (Ising), and for many more models the number of ticks is a dimensionless "time" unit (Wolf Sheep, GasLab). I think for these types of models, this decision is at worst neutral:

* For units without ticks it saves screen real-estate and avoids view clutter
* For units where the exact tick count is irrelevant it removes a potentially confusing number
* For units where ticks are a time unit, it groups ticks with the speed slider, helping to associate the speed slider with ticks, the measure of speed

There are many models for which ticks do represent a meaningful unit. These models can add a monitor for ticks, but they can also control the positioning and label of the monitor without needing a special "Tick Counter Label" set by the view. I'm open to including a "Ticks" monitor as part of the default model so that user's see it without needing to add it (while giving other users who don't want it the ability to delete it).

SethTisue:
> For units where the exact tick count is irrelevant it removes a potentially confusing number

Your other two points are valid, but this one isn't; the Settings dialog has had a "Show tick counter" checkbox for a long time.

PR #1072, opened 2016-06-09
Suggestion box while typing in the editor

Akarshit:
This is the initial version of the suggestion box for the community to try. 
Some bugs related to backspace are still left to fix. Meanwhile I would be happy happy to fix any other bugs encountered.


mrerrormessage:
Glad to see this PR. I'll review it in a few hours.

netlogo-gui/src/main/editor/Actions.scala
```diff
@@ -38,6 +38,8 @@ object Actions {
   def quickHelpAction(colorizer: Colorizer, i18n: String => String) =
     new MyTextAction(i18n("tabs.code.rightclick.quickhelp"),
       e => e.getHelpTarget(e.getSelectionStart).foreach(t => colorizer.doHelp(e, t)))
+  def codeCompletionAction(codeCompletionPopup: CodeCompletionPopup) = new MyTextAction("code-completion",
```
mrerrormessage: Doesn't look like the parameter is actually used here, could this just be `def codeCompletionAction = ...`?

netlogo-gui/src/main/editor/CodeCompletionPopup.scala
```diff
@@ -0,0 +1,135 @@
...
+import org.nlogo.api.CompilerServices
+import org.nlogo.core.DefaultTokenMapper
+import org.nlogo.window.AutoSuggest
+
+case class CodeCompletionPopup(compiler: CompilerServices) {
+  // To control when the popup has to automatically show
+  var isPopupEnabled = false
+  // To store the last value user selected
+  var lastSuggested = ""
+  var editorArea: EditorArea = null
```
mrerrormessage: This should be `Option[EditorArea]` instead of `null`. In scala you should use an `Option` whenever you have a `var` that might be empty.

mrerrormessage:
A few notes from testing this:
* When I started a completion and backspaced out until it disappeared, NetLogo locked and wouldn't do anything until I quit.
* I like that primitives are colored, but the colors are swapped. Commands should be in blue, reporters should be in purple.

netlogo-gui/src/main/editor/CodeCompletionPopup.scala
```diff
@@ -0,0 +1,135 @@
...
+    showPopup()
+    val position = editorArea.getCaretPosition
+    window.setLocation(editorArea.getLocationOnScreen.x + editorArea.modelToView(position + 1).x,
+      editorArea.getLocationOnScreen.y + editorArea.modelToView(position).y + editorArea.getFont.getSize)
+  }
+
+  /**
+    * Updates the list of the suggestion box and makes it visible.
+    */
+  def showPopup(): Unit = {
```
mrerrormessage: Both `enablePopup` and `showPopup` have comments explaining what they do and their name is not part of that comment. I would like to see them named a bit differently to indicate their functions - perhaps `displayPopup` and `updatePopup`?

netlogo-gui/src/main/editor/CodeCompletionPopup.scala
```diff
@@ -0,0 +1,135 @@
...
+    * Makes the suggestion box initially and displays it.
+    * Should only be called when the user presses the shortcut to
+    * enable the suggestion. To update the suggestions list call showPopUp()
+    */
+  def enablePopup(): Unit = {
+    isPopupEnabled = true
+    showPopup()
+    val position = editorArea.getCaretPosition
+    window.setLocation(editorArea.getLocationOnScreen.x + editorArea.modelToView(position + 1).x,
+      editorArea.getLocationOnScreen.y + editorArea.modelToView(position).y + editorArea.getFont.getSize)
```
mrerrormessage: When the popup is first shown to the user, I can definitely see it appear initially at the top left corner of the screen and then move to the current location. Could we make this call to `setLocation` come before `window.setVisible`?

netlogo-gui/src/main/editor/CodeCompletionPopup.scala
```diff
@@ -0,0 +1,135 @@
...
+  * This is the renderer to render elements of the list displayed
+  * in the suggestion box.
+  * @param editorArea
+  */
+class SuggestionListRenderer(editorArea: EditorArea) extends ListCellRenderer[String]{
+
+  override def getListCellRendererComponent(list: JList[_ <: String], value: String, index: Int, isSelected: Boolean, cellHasFocus: Boolean): Component = {
+    val label = new JLabel(value.asInstanceOf[String])
+    label.setForeground(if (DefaultTokenMapper.getCommand(value.asInstanceOf[String]).isEmpty) Color.BLUE
+    else new Color(0x551A8B))
```
mrerrormessage: I would prefer these colors to come out of `window.SyntaxColors`, since those are the colors used by the EditorColorizer.

netlogo-gui/src/main/editor/EditorArea.java
```diff
@@ -101,6 +114,30 @@ public EditorArea(int rows, int columns,
...
+      }
+
+      @Override
+      public void removeUpdate(DocumentEvent e) {
+        codeCompletionPopup.showPopup();
+      }
+
+      @Override
+      public void changedUpdate(DocumentEvent e) {}
+    });
```
mrerrormessage: All of this code works, but it's a lot of detail for a particular feature of the `EditorArea`. I'm going to leave a longer comment on the PR suggesting some ways that we can add functionality here without having `EditorArea` depend directly on `CodeCompletionPopup`.

netlogo-gui/src/main/window/AutoSuggest.scala
```diff
@@ -0,0 +1,218 @@
...
+  val trie = new TrieNode()
+  for(commandName <- commandNames) {
+    trie.append(commandName.toLowerCase)
+  }
+  for(reporterName <- reporterNames) {
+    trie.append(reporterName.toLowerCase)
+  }
+
+  def editDistance(s1: String, s2: String): Int = {
+    val memo = scala.collection.mutable.Map[(List[Char],List[Char]),Int]()
```
mrerrormessage: This should be a `scala.collection.mutable.WeakHashMap`. Also, please import the class at the beginning of the file and just use `Map`/`WeakHashMap` in code. Fully-qualified names are totally acceptable in the java files within NetLogo, but the devel team prefers scala code use imports where possible.

netlogo-gui/src/main/window/AutoSuggest.scala
```diff
@@ -0,0 +1,218 @@
...
+    def min(a:Int, b:Int, c:Int) = Math.min( Math.min( a, b ), c)
+    def sd(s1: List[Char], s2: List[Char]): Int = {
+      if (memo.contains((s1,s2)) == false)
+        memo((s1,s2)) = (s1, s2) match {
+          case (_, Nil) => s1.length
+          case (Nil, _) => s2.length
+          case (c1::t1, c2::t2)  => min( sd(t1,s2) + 1, sd(s1,t2) + 1,
```
mrerrormessage: I prefer not to have spaces between the open parens of a method invocation and its first argument. If you want to space out the arguments it's totally acceptable to move each argument to its own line.

netlogo-gui/src/main/window/AutoSuggest.scala
```diff
@@ -0,0 +1,218 @@
...
+            case Some(child) => helper(currentIndex + 1, child, items)
+            case None => items
+          }
+        }
+      }
+
+      helper(0, this, new ListBuffer[String]())
+    }
+
+    // Optimize this for @tailrec
```
mrerrormessage: Are you planning to optimize this for `@tailrec`? This comment will need to be removed either way.

netlogo-gui/src/main/window/AutoSuggest.scala
```diff
@@ -0,0 +1,218 @@
...
+  for(commandName <- commandNames) {
+    trie.append(commandName.toLowerCase)
+  }
+  for(reporterName <- reporterNames) {
+    trie.append(reporterName.toLowerCase)
+  }
+
+  def editDistance(s1: String, s2: String): Int = {
+    val memo = scala.collection.mutable.Map[(List[Char],List[Char]),Int]()
+    def min(a:Int, b:Int, c:Int) = Math.min( Math.min( a, b ), c)
```
mrerrormessage: In scala this can be written `a min b min c`

netlogo-gui/src/main/window/AutoSuggest.scala
```diff
@@ -0,0 +1,218 @@
...
+    def sd(s1: List[Char], s2: List[Char]): Int = {
```
mrerrormessage: What does `sd` mean? I would prefer a more descriptive name unless this is a domain-specific term (if it is just let me know).

netlogo-gui/src/main/window/AutoSuggest.scala
```diff
@@ -0,0 +1,218 @@
...
+
+  }
+
+  class TrieNode(val char : Option[Char] = None, var word: Option[String] = None) extends Trie {
+
+    val children: mutable.Map[Char, TrieNode] = new java.util.TreeMap[Char, TrieNode]().asScala
+
+    override def append(key: String) = {
+
+      @tailrec def appendHelper(node: TrieNode, currentIndex: Int): Unit = {
```
mrerrormessage: I really appreciate seeing `@tailrec`. It's super-important for a lot of recursive functions in scala. In general I prefer function annotations in the line before the function. Something like:
```scala
@tailrec
def appendHelper(...
```

mrerrormessage:
First, let me say that this PR looks really really good overall. I made a lot of comments, but many of them were about really small textual things. The general structure of the code looks good and I appreciate that you've put a lot of effort into it and generally gotten it working well.

I want to talk a little bit about how this code (and other code you'll be writing this summer) fits in architecturally. You'll notice that this branch failed the `depend` task on Travis. The general purpose of the depend task is to keep our code flexible by reducing the couplings between different modules. The particular violations in your code were that `CodeCompletionPopup` depends on `CompilerServices` and `window.AutoSuggest`. The reason they failed is because `CompilerServices` lives in `api` and `AutoSuggest` in `window`. We don't allow `editor` to depend on anything except for `core`. The reason for this is that we otherwise risk dependency cycles (since `window` depends on `editor`, for instance). We're going to need to take a few steps to fix the dependency issue here. The first is moving the files you created to a new package, we'll call it `ide`. That package can have an entry in `project/Depend.scala` allowing it to depend on `api` and `editor`. We'll also add an entry to `window` to allow that class to depend on `editor`.

Moving these files into a new package will get us most of the way there, but we'll still need to cut the dependency between `EditorArea` and `codeCompletionPopup`. We can do that by having `EditorArea` take a `Map<KeyStroke, TextAction>` and registering each pair with the input map on initialization. Then we'll pass in a Map containing the appropriate keystroke and code completion action when we build a new `EditorArea` in `CodeEditor`. This also means that we'll have to change the behavior of `codeCompletionAction` it can still be a `MyTextAction`, but it will have to be defined somewhere else, probably in the `ide` package. Notice that `MyTextAction` is built with a procedure that gives access to the `EditorArea`. Instead of just telling the popup to show itself, we can actually `init` a `codeCompletionPopup` at the time we get the action and add the appropriate `DocumentListener` to the `EditorArea` as well. This will allow us to have code completion without `EditorArea` needing to know anything about how code completion is setup.

All of what I just said is pretty complex, so please don't hesitate to ask any questions you may have. I'll be available tomorrow to answer questions and discuss any problems that may arise. Thanks again for the PR, it looks terrific!

Akarshit:
@mrerrormessage Thanks for reviewing the PR in detail. I will make the changes you suggested above.

netlogo-gui/src/main/window/AutoSuggest.scala
Akarshit: Actually it was String Distance. Anyway I will change that to something more descriptive.

netlogo-gui/src/main/ide/CodeCompletionPopup.scala
```diff
@@ -0,0 +1,152 @@
...
+        eA.getLocationOnScreen.y + eA.modelToView(position).y + eA.getFont.getSize)
+      updatePopup()
+    }
+  }
+
+  /**
+    * Updates the list of the suggestion box and makes it visible.
+    */
+  def updatePopup(): Unit = {
+    println("I was called")
```
mrerrormessage: This should be removed.

mrerrormessage:
The structure of the code looks much improved here. `netlogo/depend` is still failing, but you can add the appropriate entries for `ide` - allowing `ide` to depend on `editor` and `api`, as we discussed - and allowing `app` to depend on `ide`.

The colors for reporters and commands are still swapped. Commands should be blue, reporters should be purple. The code looks good for the most part, I'll go through and leave more comments line-by-line (but I may not be able to do that for all the code until tomorrow).

In testing, I'm running into exceptions for the following two cases, each of which raises a hard error in the NetLogo UI.
* Open a code completion popup, press escape, go somewhere else in the text, open a code completion popup, choose a selection and press enter.
* Open a code completion popup, press space, choose an item and press enter.

These look like separate bugs, but may end up having similar causes. Either way let's try to catch all the exceptions in `CodeCompletionPopup` so they don't propagate out to the user.

netlogo-gui/src/main/ide/CodeCompletionPopup.scala
```diff
@@ -0,0 +1,152 @@
...
+          list.foreach(dlm.addElement(_))
+          if (!list.isEmpty) {
+            lastSuggested = ""
+            window.validate()
+            window.setVisible(true)
+            // Required to keep the caret in the editorArea visible
+            eA.getCaret.setVisible(true)
+            suggestionDisplaylist.setSelectedIndex(0)
+          } else {
+            isPopupEnabled = false
```
mrerrormessage: It looks to me like there's a state machine made up of `isPopupEnabled`, `isVisible` and whether `autoSuggestListener` is registered. Think about whether that can be extracted into an object that manages the state to keep things simpler in the swing code.

netlogo-gui/src/main/ide/CodeCompletionPopup.scala
```diff
@@ -87,48 +91,66 @@ case class CodeCompletionPopup(compiler: CompilerServices) {
...
     for{eA <- editorArea} {
       isPopupEnabled = true
-      val token = compiler.getTokenAtPosition(eA.getText(), eA.getCaretPosition)
-      val position = if (token != null) token.start else eA.getCaretPosition
+      val tokenOption = getTokenTillPosition(eA.getText(), eA.getCaretPosition)
+      var position = 0
+      tokenOption match {
+        case Some(token) => position = token.start
+        case None => position = eA.getCaretPosition
+      }
```
Idloj: The above 5 lines can be simplified to:
`val position = tokenOption.map(_.start).getOrElse(eA.getCaretPosition)`

mrerrormessage:
Regarding `project/Depend.scala` - add the following line around line 63: `"ide" -> List("editor", "api"),`, also modify line 44 to be: "app" -> List("ide", "window"),`. The general format of project/Depend.scala is a map from package names to the names of packages which those packages are allowed to depend on.

netlogo-gui/src/main/ide/CodeCompletionPopup.scala
```diff
@@ -5,13 +5,15 @@ package org.nlogo.ide
...
 import java.awt.event._
 import javax.swing._
+import javax.swing.event.DocumentEvent
 
 import org.nlogo.api.CompilerServices
-import org.nlogo.core.DefaultTokenMapper
+import org.nlogo.core.{DefaultTokenMapper, Token}
 import org.nlogo.window.SyntaxColors
 import org.nlogo.editor.EditorArea
+import org.nlogo.lex.Tokenizer
```
mrerrormessage: This will need to be removed to get `depend` passing. Use `org.nlogo.core.TokenizerInterface` instead.

netlogo-gui/src/main/ide/CodeCompletionPopup.scala
```diff
@@ -87,48 +91,66 @@ case class CodeCompletionPopup(compiler: CompilerServices) {
...
+              window.setVisible(false)
+            }
+          case None => window.setVisible(false)
         }
       }
     }
   }
-  def isVisible() = window.isVisible
+  def getTokenTillPosition(source: String, position: Int): Option[Token] = {
+    val iterator = Tokenizer.tokenizeString(source)
```
mrerrormessage: Instead of using `Tokenizer`, use `org.nlogo.core.Femto.scalaSingleton[TokenizerInterface]("org.nlogo.lex.Tokenizer")`. Why go through all of that just to use the same object in the end? Because it ensures that this class depends on the stable published interface `org.nlogo.core.TokenizerInterface` instead of `org.nlogo.lex.Tokenizer`, which might change for any number of reasons unrelated to this class.

netlogo-gui/src/main/ide/CodeCompletionPopup.scala
```diff
@@ -87,48 +91,66 @@ case class CodeCompletionPopup(compiler: CompilerServices) {
...
-            window.validate()
-            window.setVisible(true)
-            // Required to keep the caret in the editorArea visible
-            eA.getCaret.setVisible(true)
-            suggestionDisplaylist.setSelectedIndex(0)
-          } else {
-            isPopupEnabled = false
+      if (isPopupEnabled ) {
+        var position = eA.getCaretPosition
+        for{docEvent <- docEventOption} {
```
mrerrormessage: Only lines 111-119 require the `docEventOption` parameter. Perhaps it would be better to extract a separate method. that calls into this one?

netlogo-gui/src/main/ide/CodeCompletionPopup.scala
Akarshit: Oh nice!
Thanks.

mrerrormessage:
This looks great! I'm going to merge this in, as well as open a few new issues for a few enhancements I think we'll want down the road but don't need to do before merging this in. Terrific work Akarshit!

PR #330, opened 2016-06-10
bump API version number to 6

mrerrormessage:
at present on master the API version number is still 5.0, but if we can't promise that 5.0 extensions are binary compatible with the current code, the API version will need to be bumped (and everyone will need to recompile their extensions)


mrerrormessage:
@qiemem , as the extensions expert, would you mind to review the documentation in this PR as well as the referenced GitHub page(s) to make sure it seems correct and complete?

qiemem:
Happy to! I'll do it tomorrow.

qiemem:
Looks good. The transition guide needs a little work (removal of `DefaultCommand`/`DefaultReporter`).

qiemem:
Under the "Syntax representation" section, we should describe the split between java and scala syntax declarations.

I can do this stuff, just making notes.

PR #165, opened 2016-06-17
movie primitives generate invalid Quicktime files (say some programs)

mrerrormessage:
some programs can understand them, others can't. we never knew what the problem was before but on the MASON list Sean Luke writes:

>the second problem is that JMF generates Quicktime wrappers with corrupted format description strings.  This is also easily dealt with in the same way: load into Quicktime or iMovie or whatnot and save back out again under a different codec.
>
>Because JMF is abandonware, you should never expect these to get fixed.  And there is no Java library I am aware of which can replace JMF realistically.  MASON, NetLogo, Processing, and Repast all have this issue.  [...]

see also #165


mrerrormessage:
I'm not *quite* finished hooking everything up to get autoconversion working, but I wanted to post this for review as the general approach is basically complete. `AstRewriter` is the bulk of new code in this request and it probably leaves a lot of room open for improvement but seems to properly replace the `AutoConverter2` removed in the 6.0 transition. I'm also considering moving `AutoConverter1` to `fileformat` (from `compiler`) as it seems to me that converting old code is a task that properly belongs to the model loader instead of the compiler (though it obviously requires use of the parser).

PR #929, opened 2016-06-24
Add error view into plot pen

mrerrormessage:
Add button to display errors modally next to edit and delete.


mrerrormessage:
OK, it doesn't seem we have consensus on incrementing the plot pens when there are errors, **so** I'm going to go ahead and merge this. If we decide we want to add in auto-incrementing, we can do that in a separate issue/PR.

PR #1074, opened 2016-06-24
RNG in plots is reseeded every update

mrerrormessage:
Every time any update or setup code in a plot is run, its RNG is seeded with the same seed. Here's a model which demonstrates this: https://gist.githubusercontent.com/qiemem/fffd478921a720bc415bb1219ebb5e0f/raw/f60e65b7499ada996c86fb89c55ec7b7b3b16079/gross-plots.nlogo

sample output:

```
observer> reset-ticks
plot setup
0.6551991655274944
first pen setup
0.6551991655274944
second pen setup
observer: 0.6551991655274944
plot update
0.6551991655274944
first pen update
0.6551991655274944
second pen update
0.6551991655274944
observer> tick
plot update
0.6551991655274944
first pen update
0.6551991655274944
second pen update
0.6551991655274944
observer> tick
plot update
0.6551991655274944
first pen update
0.6551991655274944
second pen update
0.6551991655274944
```

### DIFF
--- a/netlogo-gui/src/main/workspace/Evaluator.scala
+++ b/netlogo-gui/src/main/workspace/Evaluator.scala
@@ -68,12 +68,12 @@ class Evaluator(workspace: AbstractWorkspace) {
 
   private object ProcedureRunner {
     var context: Context = null
-    def run(p: Procedure): Try[Boolean] = {
+    def run(p: Procedure, ownerOption: Option[JobOwner] = None): Try[Boolean] = {
       val oldActivation = context.activation
       val newActivation = new Activation(p, context.activation, 1)
       val oldRandom = context.job.random
       context.activation = newActivation
-      context.job.random = workspace.world.mainRNG.clone
+      context.job.random = ownerOption.map(_.random).getOrElse(workspace.world.mainRNG.clone)
       val procedureResult = Try {
         context.runExclusiveJob(workspace.world.observers, 0)
         val stopped = workspace.completedActivations.get(newActivation) != true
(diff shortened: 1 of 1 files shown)


SethTisue:
thought about this some more. I think the old behavior was very much by design. and I misunderstood — I thought @qiemem was reporting a hexy regression, not reporting longstanding behavior.

we want plots to be reproducible so that model runs as a whole as remain reproducible. using the aux RNG does not accomplish that. the aux RNG is for use by things that aren't part of the run, e.g. monitors and slider bounds. but plotting code is very much part of the run — e.g. because plotted results are included in `export-world`

qiemem:
Good points, Seth.

That said, the current behavior is definitely *wrong* even if it's not a regression. Right now, this pen:

screen shot 2016-06-29 at 7 06 10 pm

Produces this plot:

image

So it seems like there's a few goals here:

- Plots must be reproducible. If you see a run the same, you must end up with the same plot.
- Randomness in plots should actually be random. That is `plot random 10` in a pen should plot a jagged line, not a straight one.
- Plots shouldn't mess with the RNG of other stuff. That is, changing a plot should never change the behavior of the rest of your model for a given seed. That said, I think plots sharing an RNG would be fine,  but I could be wrong.

I think Roberts third proposal addresses these. Specifically, when `clear-all` or `random-seed` is called, the plot manager is given a clone of the RNG that all plots then use. I think this addresses all three points sufficiently. Alternatively, each pen could be given its own clone, but I think a bunch of pens with `plot random 10` should each end up with different plots.

mrerrormessage:
Update: I should have re-run checksums before checking this in. I'm now sorting it out on the release-60-m9 branch. I'm planning to implement the third solution, barring any unforseen difficulties.

PR #1079, opened 2016-06-25
Show Usage

Akarshit:
This is the initial version of the PR so that it could be reviewed by @mrerrormessage.


Idloj:
Shouldn't it be "Show Usage**s**"?

netlogo-gui/src/main/ide/ShowUsageBox.scala
```diff
@@ -0,0 +1,126 @@
...
+  def getUsage(source: String, token: Token): Seq[Token] = {
+    var list = Seq[Token]()
+    val iterator = Femto.scalaSingleton[TokenizerInterface]("org.nlogo.lex.Tokenizer").tokenizeString(source)
+    for(t <- iterator) {
+      if(t == token) {
+        list :+= t
+      }
+    }
+    list
+  }
```
mrerrormessage: This is generally the right idea, but needs a little work. In this situation in scala we use `.filter` instead of building a list with `for` and a mutable `Seq`. I think you may also want a different comparison than `t == token`. Tokens include the text as well as position of the token. Since we want to find tokens in different positions you'll probably want to match on `token.text` and probably use `equalsIgnoreCase`.

netlogo-gui/src/main/ide/ShowUsageBox.scala
```diff
@@ -0,0 +1,126 @@
...
+
+  def getTokenTillPosition(source: String, position: Int): Option[Token] = {
+    val iterator = Femto.scalaSingleton[TokenizerInterface]("org.nlogo.lex.Tokenizer").tokenizeString(source)
+    iterator.find(p => p.start < position && p.end >= position)
```
mrerrormessage: I would like to see this method renamed to something like `findTokenAtPosition` since, if I understand it correctly, it finds the token which starts before the specified position and ends after it (if one exists).

netlogo-gui/src/main/ide/ShowUsageBox.scala
```diff
@@ -0,0 +1,126 @@
...
+  }
+  class LineNumberRenderer extends DefaultTableCellRenderer {
+    override def setValue(value: AnyRef) = {
+      setText(int2Integer(editorArea.offsetToLine(value.asInstanceOf[Token].start)).toString)
```
mrerrormessage: You don't need to convert to `Integer` to use `.toString` on `.start`, you can just do `.start.toString`.

netlogo-gui/src/main/ide/ShowUsageBox.scala
```diff
@@ -0,0 +1,126 @@
...
+      override def keyReleased(e: KeyEvent): Unit = {}
+    })
+  }
+
+  def showBox(me: MouseEvent, position: Int): Unit = {
+    val tokenOption = getTokenTillPosition(editorArea.getText(), position)
+    for {token <- tokenOption} {
+      val tokens = getUsage(editorArea.getText(), token)
+      dataModel.synchronized(dataModel.setRowCount(0))
+      dataModel.synchronized(
```
mrerrormessage: As far as I can tell, you're only updating `dataModel` on the event thread here. Is there a reason to use `synchronized` that I'm missing?

netlogo-gui/src/main/ide/ShowUsageBox.scala
```diff
@@ -0,0 +1,126 @@
...
+
+      override def mouseEntered(e: MouseEvent): Unit = {}
+
+      override def mousePressed(e: MouseEvent): Unit = {}
+
+      override def mouseReleased(e: MouseEvent): Unit = {
+        usageBox.setVisible(false)
+        val token = usageTable.getValueAt(usageTable.getSelectedRow, 0).asInstanceOf[Token]
+        dataModel.synchronized(dataModel.setRowCount(0))
+        println(token.start + " " + token)
```
mrerrormessage: I know this is here for debugging, just commenting to make sure we remember to remove it before we merge in.

netlogo-gui/src/main/ide/ShowUsageBox.scala
```diff
@@ -0,0 +1,126 @@
...
+    })
+    dataModel.setColumnCount(2)
+    usageTable.setSelectionMode(javax.swing.ListSelectionModel.SINGLE_SELECTION)
+    usageTable.getColumnModel.getColumn(0).setMaxWidth(40)
+    usageTable.getColumnModel.setColumnSelectionAllowed(false)
+    usageTable.getTableHeader().setReorderingAllowed(false)
+    usageTable.setTableHeader(null)
+    usageTable.setRowMargin(3)
+    usageTable.setRowHeight(20)
+//    usageTable.setBackground(Color.WHITE)
```
mrerrormessage: In general, I remove commented code whenever I see it. It's totally fine when you're building stuff on a branch, just make sure it gets cleaned up before merge.

netlogo-gui/src/main/ide/ShowUsageBox.scala
Akarshit: OK I get your point about `t == token`. But I have doubt, why is it still working?

netlogo-gui/src/main/ide/ShowUsageBox.scala
Idloj: `EditorArea` has a similar function (though a bit different. It returns Option[String], it is more complicated and it uses the EditorArea's source instead of getting it through a parameter): https://github.com/NetLogo/NetLogo/blob/hexy/netlogo-gui/src/main/editor/EditorArea.java#L505-518

Maybe this function can be moved to `EditorArea`, and `EditorArea.getHelpTarget` will use it?

mrerrormessage: This isn't a bad idea at first glance, but there are some very particular rules about what `EditorArea.getHelpTarget` should find that I don't want to mess with. Akarshit and I discussed this and we determined that he had different needs than `EditorArea.getHelpTarget`.

netlogo-gui/src/main/ide/ShowUsageBox.scala
mrerrormessage: This is a great question and I actually learned something about scala by figuring out how to answer it! You can find the `equals` implementation of a case class by using  `scala -Xprint:typer`, so I ran `scala -Xprint:typer -e 'case class Token(text: String, value: AnyRef)(val start: Int, val end: Int, val filename: String)'` and got the following:

```
...
          override <synthetic> def equals(x$1: Any): Boolean = Token.this.eq(x$1.asInstanceOf[Object]).||(x$1 match {
  case (_: this.Token) => true
  case _ => false
}.&&({
            <synthetic> val Token$1: this.Token = x$1.asInstanceOf[this.Token];
            Token.this.text.==(Token$1.text).&&(Token.this.value.==(Token$1.value)).&&(Token$1.canEqual(Token.this))
          }))
...
```
Which tells us that the `==` method on token only takes `text`, `tpe`, and `value` into account. `tpe` will (pretty obviously) be the same for both tokens (less obviously, `value` will almost always be the same except in a few corner cases which I don't care to get into). However, `text` is the same only when both strings are the same (it's case-sensitive). In most languages this would be the correct behavior, but NetLogo is case-insensitive, so we need to compare without respect to case.

mrerrormessage:
OK, this looks really great overall. I like the way lines are displayed and having the line numbers helps provide a bit of context.

I'd like to discuss a few changes:
* Adding color to the display of the usages would be great, if possible
* We need to figure out something other than Ctrl+click. On mac, that's also how the context menu is triggered when using a one-button mouse. I think this feature is specialized enough that it would fit well in the editor area context menu.
* Let's see if we can find a way to initialize `ShowUsageBox` that doesn't require `EditorArea` calling `init`.
* We're doing a lot of customizing of `EditorArea` and I think it would be a good idea to convert it to scala. I'm happy to work on that conversion, or you can work on it. If you want to work on it I think it would be best to include it in this PR. If I work on it I can do it for this PR or afterwards (your choice).
* At the moment, this finds usage of keywords and special punctuation (like `to`, `end`, `[`). I would like to get rid of these. You can detect "interesting" words by looking at the `TokenType`. For our purposes `Command`, `Reporter` and `Ident` are the interesting token types.

Let me know if you have any questions about these. I've sketched a bunch of ideas in very general terms and I'm happy to flesh out more as needed.

netlogo-gui/src/main/ide/ShowUsageBox.scala
Akarshit: OK thanks :)

netlogo-gui/src/main/ide/ShowUsageBox.scala
Akarshit: I didn't want to confuse the method with `compiler.getTokenAtPosition()` which return the "interesting" token. Maybe we can rename that method to something else and this to `getTokenAtPosition`

mrerrormessage: What about `findTokenContainingPosition`?

Akarshit:
@mrerrormessage Thanks for the ideas. Some of these were in my todo list.
Can you elaborate more on 
>Adding color to the display of the usages would be great, if possible

mrerrormessage:
> Adding color to the display of the usages would be great, if possible

I just meant making commands blue and reporters purple, just basic syntax coloring.

netlogo-gui/src/main/ide/ShowUsageBox.scala
Akarshit: Hmm that sound good to me.

Idloj:
There is no horizontal scrollbar, and it can be necessary in some cases.

Idloj:
I have some enhancement ideas:
 - Exclude the line the user has clicked on.
 - Show the searched token in a different background
 - Maybe do the same for the actual editor area (i.e. highlighting the searched token in the code itself in addition to showing the popup). Maybe also do this highlighting whenever the user selects an identifier.

netlogo-gui/src/main/app/CodeTab.scala
```diff
@@ -28,7 +30,25 @@ class CodeTab(val workspace: AbstractWorkspace) extends JPanel
       dirty()
     }
   }
+  val showUsageBox = new ShowUsageBox()
+  val showUsageBoxAction = new ShowUsageBoxAction(showUsageBox)
+  val mL = new MouseListener {
```
Idloj: This can use MouseAdapter

netlogo-gui/src/main/ide/ShowUsageBox.scala
```diff
@@ -0,0 +1,116 @@
...
+      override def mouseClicked(e: MouseEvent): Unit = {
+        usageBox.setVisible(false)
+        val token = usageTable.getValueAt(usageTable.getSelectedRow, 0).asInstanceOf[Token]
+        dataModel.synchronized(dataModel.setRowCount(0))
+        editorArea.setCaretPosition(token.start)
+      }
+    })
+
+    usageTable.addKeyListener(new KeyListener(){
+      override def keyPressed(e: KeyEvent): Unit = {
```
Idloj: This can be replaced by KeyAdapter

Akarshit:
@mrerrormessage Did the following things
> Adding color to the display of the usages would be great, if possible

> Let's see if we can find a way to initialize ShowUsageBox that doesn't require EditorArea calling init.

> At the moment, this finds usage of keywords and special punctuation (like to, end, [). I would like to get rid of these. You can detect "interesting" words by looking at the TokenType. For our purposes Command, Reporter and Ident are the interesting token types.

About the `EditorArea` conversion to Scala I think it would be better if you could do it. I still have a very basic knowledge of Scala.

@Idloj I have implemented these
> Exclude the line the user has clicked on.

> Make the lines clickable, so it scrolls the editor area to that line(This was already there. Is it not working for you?)

I am trying to make the backgroud of the selected `Token` different in the `ShowUsageBox` but I am having some problems in that. Maybe @mrerrormessage can help me here. I am trying to select some text from the `pane` after [this](https://github.com/NetLogo/NetLogo/pull/1079/files#diff-7fe590a02d72cba4a230ce0f1d0c2144R122) but the text is not getting selected.

mrerrormessage:
> About the EditorArea conversion to Scala I think it would be better if you could do it. I still have a very basic knowledge of Scala.

Sounds good, I'll try to work on that later this week.

> I am trying to make the backgroud of the selected Token different in the ShowUsageBox

Instead of making the background different, what about just making the font of the selected token bold? Would that make the problem easier? I'm not sure about the text selection, but if that's necessary I can look into it more.

Idloj:
> Make the lines clickable, so it scrolls the editor area to that line(This was already there. Is it not working for you?)

Oh, sorry, didn't realize. My only problem with this is that the wanted line is at the very bottom of the window, so it took me several moments to realize where it is (even with the aid of comparing the line numbers in the editor and in the show usage box). Maybe it can be at the top of the window or in the middle? Also, Selecting the token or the entire line would help a lot.

Idloj:
Another enhancement idea: change the mouse shape to pointer when hovering over the show usage box. Currently, there is simply nothing to indicate that you can click the lines (which is probably why I missed it)

Akarshit:
@mrerrormessage Same problem with making the token bold. I have tried using JTextPane which supports bold but then I am not able to add the `Colorizer` to it.

Akarshit:
@Idloj Added the hand cursor to the box.

Akarshit:
@mrerrormessage After the user click on a item in the `ShowUsageBox` should I select the token in that line in the `EditorArea`?

Akarshit:
@Idloj I tried the zebra stripes but that was not looking good. The main reason being that user will move up/down in the suggestion list frequently as opposed to a simple click in `Show Usage`. The screenshots of the suggestion box are below with zebra strips.
screenshot from 2016-06-29 15 16 02
screenshot from 2016-06-29 15 16 28

mrerrormessage:
I think after clicking an item in the `ShowUsageBox` it would be great to select that token in that line in `EditorArea`.

Also, I've managed to get **bold** tokens working on [this branch](https://github.com/NetLogo/NetLogo/tree/wip-bold-usages). There's a decent amount of hackery there, but I think you can see the basic approach which is to override the `HighlightEditorKit` and `HighlightView` as appropriate to add in bold text. Converting either (or both) of these classes to scala might make changing them easier. There's certainly some inefficient code in `HighlightView` that could be made a lot better by using scala's `Range` class. Let me know if you have any questions about the code on that branch.

netlogo-gui/src/main/ide/ShowUsageBox.scala
```diff
@@ -0,0 +1,133 @@
...
+    iterator.filter(t => t.text.equalsIgnoreCase(token.text) && t.start != token.start)
+  }
+
+  def findTokenContainingPosition(source: String, position: Int): Option[Token] = {
+    val iterator = Femto.scalaSingleton[TokenizerInterface]("org.nlogo.lex.Tokenizer").tokenizeString(source)
+    iterator.find(p => p.start < position && p.end >= position)
+  }
+
+  class LineRenderer extends TableCellRenderer {
+    override def getTableCellRendererComponent(table: JTable, value: scala.Any, isSelected: Boolean, hasFocus: Boolean, row: Int, column: Int): Component = {
```
Idloj: You can just use `Any` instead of `scala.Any`. (and maybe `AnyRef` is more appropriate here? I guess this is `Object` in the Java as it's Swing API)

Idloj:
You can use `pane.getStyledDocument.setCharacterAttributes` (https://docs.oracle.com/javase/8/docs/api/javax/swing/text/StyledDocument.html#setCharacterAttributes(int,%20int,%20javax.swing.text.AttributeSet,%20boolean)) and somehow pass it a FontAttribute, I don't know the details, just looked a bit through the javadocs

netlogo-gui/src/main/ide/ShowUsageBox.scala
mrerrormessage: Good catch, should definitely be `AnyRef`.

netlogo-gui/src/main/ide/ShowUsageBox.scala
```diff
@@ -0,0 +1,168 @@
...
+  usageTable.setRowHeight(20)
+  usageTable.getColumnModel.getColumn(1).setMinWidth(200)
+  usageTable.getColumnModel.getColumn(0).setMinWidth(40)
+  usageTable.setFont(editorArea.getFont)
+
+  usageTable.addMouseListener(new MouseAdapter() {
+    override def mouseClicked(e: MouseEvent): Unit = {
+      usageBox.setVisible(false)
+      val token = usageTable.getValueAt(usageTable.getSelectedRow, 0).asInstanceOf[Token]
+      dataModel.synchronized(dataModel.setRowCount(0))
```
mrerrormessage: Don't think this needs to be synchronized.

netlogo-gui/src/main/ide/ShowUsageBox.scala
```diff
@@ -0,0 +1,168 @@
...
+  class LineNumberRenderer extends DefaultTableCellRenderer {
+    override def setValue(value: AnyRef) = {
+      setText(editorArea.offsetToLine(value.asInstanceOf[Token].start).toString)
+    }
+  }
+  import org.nlogo.editor.{ HighlightEditorKit, HighlightView }
+  class BoldEditorKit(selectedString: String) extends HighlightEditorKit(null, editorArea.colorizer) {
+    import scala.collection.immutable.Range
+    class BoldView(elem: javax.swing.text.Element) extends HighlightView(this.pane, elem, this.colorizer) {
+      var boldingRange = Option.empty[Range]
```
mrerrormessage: I made this as a proof-of-concept, but this should detect when there are multiple occurrences of the token within a line and bold all of them. You could represent this with a `Seq[Range]`.

netlogo-gui/src/main/ide/ShowUsageBox.scala
```diff
@@ -0,0 +1,168 @@
...
+        pane.setForeground(usageTable.getSelectionForeground)
+      }
+      pane.putClientProperty("Nimbus.Overrides", defaults)
+      pane.putClientProperty("Nimbus.Overrides.InheritDefaults", true)
+      pane
+    }
+  }
+  class LineNumberRenderer extends DefaultTableCellRenderer {
+    override def setValue(value: AnyRef) = {
+      setText(editorArea.offsetToLine(value.asInstanceOf[Token].start).toString)
```
mrerrormessage: This line has too much leading whitespace.

netlogo-gui/src/main/ide/ShowUsageBox.scala
```diff
@@ -0,0 +1,168 @@
...
+      pane.setBorder(BorderFactory.createEmptyBorder(1,0,0,0))
+      val editorKit = boldedString match {
+        case None => new HighlightEditorKit(null, editorArea.colorizer)
+        case Some(selectedString) => new BoldEditorKit(selectedString)
+      }
+      pane.setEditorKit(editorKit)
+      pane.setText(value.asInstanceOf[String])
+      val alternate = UIManager.getColor("Table.alternateRowColor")
+      val defaults = new UIDefaults()
+      if (row%2 == 0){
```
mrerrormessage: I thought we decided not to alternate row colors? If we're going to alternate row colors we should definitely test the colors we use on multiple platforms.

netlogo-gui/src/main/ide/ShowUsageBox.scala
```diff
@@ -0,0 +1,168 @@
...
+    for {token <- tokenOption} {
+      if(token.tpe == TokenType.Ident || token.tpe == TokenType.Command || token.tpe == TokenType.Reporter) {
+        val tokens = getUsage(editorArea.getText(), token)
+        dataModel.setRowCount(0)
+        for (t <- tokens) {
+          dataModel.addRow(Array[AnyRef](t, editorArea.getLineText(t.start).trim))
+        }
+        if (dataModel.getRowCount != 0) {
+          usageTable.setDefaultRenderer(classOf[String], new LineRenderer(Some(token.text)))
+          usageTable.setPreferredScrollableViewportSize(usageTable.getPreferredSize())
```
mrerrormessage: Too much leading whitespace on this line.

netlogo-gui/src/main/ide/ShowUsageBox.scala
```diff
@@ -0,0 +1,168 @@
...
+  usageTable.setCursor(Cursor.getPredefinedCursor(Cursor.HAND_CURSOR))
+
+  def showBox(me: MouseEvent, position: Int): Unit = {
+    val tokenOption = findTokenContainingPosition(editorArea.getText(), position)
+    for {token <- tokenOption} {
+      if(token.tpe == TokenType.Ident || token.tpe == TokenType.Command || token.tpe == TokenType.Reporter) {
+        val tokens = getUsage(editorArea.getText(), token)
+        dataModel.setRowCount(0)
+        for (t <- tokens) {
+          dataModel.addRow(Array[AnyRef](t, editorArea.getLineText(t.start).trim))
```
mrerrormessage: Weird whitespace on this line.

mrerrormessage:
I'm not sure whether I misunderstood our conversation earlier today - I thought you asked me to review this but it doesn't seem to be compiling. Did you forget to push? Were you referring to another PR? It isn't a big deal, I just want to be able to give you feedback in the right place. I left a few comments, but this wasn't compiling for me so I can't finish testing / reviewing it until it compiles.

netlogo-gui/src/main/ide/ShowUsageBox.scala
Akarshit: @mrerrormessage It looks OK to me(2 space indented). What do you exactly mean by weird?

netlogo-gui/src/main/ide/ShowUsageBox.scala
Akarshit: The line in indented 2 space from the `if`. Should I change that?

netlogo-gui/src/main/ide/ShowUsageBox.scala
mrerrormessage: This and the one a few lines down are fine, github was wrapping them strangely.

netlogo-gui/src/main/ide/ShowUsageBox.scala
Akarshit: IIRC we decided to keep the alternate colours here and not in "Auto Suggest". Let me know if this has changed.

netlogo-gui/src/main/ide/ShowUsageBox.scala
```diff
@@ -0,0 +1,169 @@
...
+  import org.nlogo.editor.{ HighlightEditorKit, HighlightView }
+  class BoldEditorKit(selectedString: String) extends HighlightEditorKit(null, editorArea.colorizer) {
+    import scala.collection.immutable.Range
+    class BoldView(elem: javax.swing.text.Element) extends HighlightView(this.pane, elem, this.colorizer) {
+      var boldingRange = Option.empty[Range]
+      override def studyLine(lineIndex: Int): Unit = {
+        super.studyLine(lineIndex)
+        val elem = getElement.getElement(lineIndex)
+        val lineText = getDocument.getText(elem.getStartOffset, elem.getEndOffset - elem.getStartOffset max 0)
+        val beginningIndex = lineText.toUpperCase.indexOf(selectedString.toUpperCase)
```
Akarshit: @mrerrormessage I will have to change this as in case of "set current-lives lives -1" it will be shown as "set current-**lives** lives - 1"

mrerrormessage: Sounds good. I wrote this code knowing there were some important cases it wouldn't work exactly properly on for the purpose of illustrating how to bold certain parts of text.

netlogo-gui/src/main/ide/ShowUsageBox.scala
mrerrormessage: That's correct. Both of these colors are the same on Mac OS X, so I assumed we didn't have it in this case.

mrerrormessage: What does this look like on Linux, I'm not seeing lines alternate colors on Windows either.

mrerrormessage:
This is a windows screenshot: 
screen shot 2016-07-06 at 9 48 04 am
Are the lines delimiting the table supposed to appear? They look a little strange to me. I think the horizontal lines would be fine, but the vertical lines and the horizontal lines together are a bit much.

Akarshit:
No those lines should not appear. Here is a screenshot of the window on linux. 
screenshot from 2016-07-07 00 16 44

I will make some changes to make it work on Mac.

mrerrormessage:
This looks much better on windows/mac:
Windows:
screen shot 2016-07-07 at 11 06 13 am
Mac:
screen shot 2016-07-07 at 11 13 04 am

I'd still like to see this as a choice in the right-menu shortcut and not just the default action on control-click, but the User Interface seems done to me other than that.

nicolaspayette:
This looks great, but... any chance we could use the monospace font? Seeing code in a proportional font is always a bit unsettling. I also think the numbers should be aligned right.

mrerrormessage:
@nicolaspayette , those are terrific suggestions, @Akarshit , let's make the font monospace and make the numbers right-aligned.

Akarshit:
@mrerrormessage On it.Other than that, should I keep the `Ctrl + Space` and the right click menu or only the right click menu?

mrerrormessage:
Just the right click menu for now.

Akarshit:
@mrerrormessage line numbers are right aligned and the font has been set to `Monospace`. I have moved the show usage to right click menu, you might want to have a look at the changes I made for that.

mrerrormessage:
Sounds great! I'll review first thing tomorrow.

mrerrormessage:
OK, so the right click menu looks really good on Mac OS X.

I wasn't seeing font-bolding (not sure whether you're still working on that or not). I decided to do a little experimentation with font bolding on Mac OS X. I found up about 3 different problems that I've researched this morning and I'll share what I've found here (there's no way you could test this on a non-mac platform). Please make sure these changes are incorporated into the font-bolding monospace changes. 

netlogo-gui/src/main/ide/ShowUsageBox.scala (line ~144)
```diff
       }
       pane.putClientProperty("Nimbus.Overrides", defaults)
       pane.putClientProperty("Nimbus.Overrides.InheritDefaults", true)
+      pane.setFont(org.nlogo.awt.Fonts.monospacedFont)
       pane
     }
   }
```
From what I could see you were setting the monospaced font in the table, it needs to be set in the particular pane where code is displayed. For future reference, when we set fonts we should be using `org.nlogo.awt.Fonts` which produces the standard fonts on each platform.

netlogo-gui/src/main/awt/Fonts.scala:
```diff

   lazy val platformMonospacedFont =
     if (os("Mac"))
-      "Monaco"
+      "Menlo"
     else if (os("Windows"))
       GraphicsEnvironment.getLocalGraphicsEnvironment.getAvailableFontFamilyNames
         .find(_.equalsIgnoreCase("Lucida Console")).getOrElse("Monospaced")
```
"Monaco" was replaced in Mac OS X 10.6 by "Menlo". "Monaco" has no bold variant which means that there is no bolding, even with the change made above.

ShowUsageBox.scala:
```diff
       }

       override def drawText(g: java.awt.Graphics, x: Int, y: Int, p0: Int, p1: Int, isSelected: Boolean): Int = {
+        g match {
+          case g2d: java.awt.Graphics2D =>
+            g2d.setRenderingHint(
+                      java.awt.RenderingHints.KEY_TEXT_ANTIALIASING,
+                      java.awt.RenderingHints.VALUE_TEXT_ANTIALIAS_GASP)
+          case _ =>
+        }
         var endX = super.drawText(g, x, y, p0, boldingRanges.head.start, isSelected)
         val originalFont = g.getFont
         val boldFont = originalFont.deriveFont(java.awt.Font.BOLD)
```
So everything looks *ok* without this, but some of the letters are out of position. With this change everything looks just right on Mac OS X.

I'm about to test this on Windows. I'll comment separately if I find anything on there.

mrerrormessage:
Looks great on Windows.

Akarshit:
@mrerrormessage Thanks for the testing on Mac and Windows. I have updated the PR accordingly.

netlogo-gui/src/main/ide/ShowUsageBox.scala
```diff
@@ -0,0 +1,203 @@
...
+      usageTable.addKeyListener(new KeyAdapter() {
+        override def keyPressed(e: KeyEvent): Unit = {
+          e.getKeyCode match {
+            case KeyEvent.VK_ESCAPE => usageBox.setVisible(false)
+            case _ => e.setSource(editorArea)
+              editorArea.dispatchEvent(e)
+          }
+        }
+      })
+//      usageTable.setFont(new Font(Font.MONOSPACED, Font.PLAIN, editorArea.getFont.getSize))
```
mrerrormessage: This line should be removed.

netlogo-gui/src/main/editor/EditorArea.scala
```diff
@@ -0,0 +1,435 @@
...
+      val isTextSelected = Option(text).exists(_.length > 0)
+      copyItem.setEnabled(isTextSelected)
+      cutItem.setEnabled(isTextSelected)
+      pasteItem.setEnabled(
+        Toolkit.getDefaultToolkit.getSystemClipboard.isDataFlavorAvailable(DataFlavor.stringFlavor))
+      val point = new Point(invoker.getLocationOnScreen)
+      point.translate(x, y)
+      for(item <- menuItems){
+        item.putValue("cursorLocation", mousePos)
+        item.putValue("popupLocation", point)
```
mrerrormessage: I like this a lot. It's a nice method of providing data to actions that need it without coupling the editor to the individual actions.

mrerrormessage:
OK, the font changes are showing up well on Mac.

screen shot 2016-07-08 at 1 38 09 pm

A few things I did notice:
* What do you think about moving line numbers to the right?
* Still no line striping. I'm not sure it's necessary, but it does differ between Windows/Linux and Mac
* I would probably prefer the text area to be a little wider, since it usually doesn't fit the whole line even for short lines of code.

I did notice that if I have a procedure like:

```NetLogo
to foo
  bAr
  bar
end
```

and I "Show Usage" on either "bar" or "bAr", an exception is thrown. That needs to be fixed. I've also left a few comments on various lines, which you can look through (mostly minor stuff).

netlogo-gui/src/main/window/CodeEditor.scala
```diff
@@ -0,0 +1,18 @@
...
+
+import org.nlogo.editor.EditorArea
+
+class CodeEditor(rows: Int, columns: Int,
+                 font: java.awt.Font,
+                 enableFocusTraversalKeys: Boolean,
+                 listener: java.awt.event.TextListener,
+                 colorizer: org.nlogo.editor.Colorizer,
+                 i18n: String => String,
+                 actionMap: Map[KeyStroke, TextAction] = EditorArea.emptyMap, actions: Seq[Action] = EditorArea.emptySeq) extends
```
mrerrormessage: Let's move `actions` to the next line, just as all the other arguments have their own line.

netlogo-gui/src/main/window/CodeEditor.scala
```diff
@@ -0,0 +1,18 @@
+// (C) Uri Wilensky. https://github.com/NetLogo/NetLogo
+
+package org.nlogo.window
+
+import javax.swing._
```
mrerrormessage: I know I put this in here, let's see if we can import individual classes instead of a wildcard import.

Akarshit:
@mrerrormessage 
> What do you think about moving line numbers to the right?

Here is a screenshot of how it looks
screenshot from 2016-07-09 03 04 34
> Still no line striping. I'm not sure it's necessary, but it does differ between Windows/Linux and Mac

If it is not I will remove it form the linux too. If it is I don't think I would be able to test on a Mac or Windows because I don't have either of them

Akarshit:
@mrerrormessage The width of the text area has been increased, fixed the bug and made the changes that you mentioned. Let me know if the line numbers have to be moved right.

Akarshit:
Also added the feature to centre the line on click. When a `Usage` is selected from the list first that line is brought into the centre of the `EditorArea` and then the `Token` is selected.

mrerrormessage:
I think this looks good. Should be ready to merge. Can you merge `hexy` into it or rebase it on top of `hexy`? Want to make sure the merge is clean and have travis verify that all tests pass, then it can be merged.

Idloj:
I think a horizontal scrollbar is quite necessary

mrerrormessage:
I disagree. If a person wants to see more than a snippet, they can always click on the line and view it in the editor.

Idloj:
Yeah, that's a point, though it seems weird to have a line in the usage box where you can't even see the token you're looking for (if the token itself is outside of the view, like in the sixth line in the last screenshot).

I also noticed that there is no left margin, the text sticks to the left edge (maybe it's only disturbing for me)

mrerrormessage:
hmmm...it is an interesting case when the token is outside of the view. Ideally we would handle that by showing the immediate context of the token, but that could get a little tricky.

I don't miss the left margin personally, but I can see how some people might. @Akarshit , what do you think?

netlogo-gui/src/main/app/EditorFactory.scala
```diff
@@ -2,32 +2,39 @@
...
                                  java.awt.Font.PLAIN, 12)
     val colorizer = new org.nlogo.window.EditorColorizer(compiler)
     val codeCompletionPopup = new CodeCompletionPopup()
-    val actionMap = mapAsJavaMap(Map(KeyStroke.getKeyStroke(java.awt.event.KeyEvent.VK_SPACE, java.awt.event.InputEvent.CTRL_DOWN_MASK)
-      -> new AutoSuggestAction("auto-suggest", codeCompletionPopup))).asInstanceOf[java.util.Map[KeyStroke, TextAction]]
+    val showUsageBox = new ShowUsageBox()
+    val actions = Seq[Action](new ShowUsageBoxAction(showUsageBox))
+    val actionMap = Map(
+      KeyStroke.getKeyStroke(java.awt.event.KeyEvent.VK_SPACE, java.awt.event.InputEvent.CTRL_DOWN_MASK)
+      -> new AutoSuggestAction("auto-suggest", codeCompletionPopup))
```
mrerrormessage: Minor code style tweak: it would be more typical style to have `->` at the end of line 32 than at the beginning of line 33.

netlogo-gui/src/main/editor/EditorArea.scala
```diff
@@ -0,0 +1,473 @@
...
+  override def paintComponent(g: Graphics): Unit = {
+    val g2d = g.asInstanceOf[Graphics2D]
+    g2d.setRenderingHint(RenderingHints.KEY_ANTIALIASING, RenderingHints.VALUE_ANTIALIAS_ON)
+    super.paintComponent(g)
+  }
+
+  private var bracketMatcherEnabled: Boolean = true
+
+  override def enableBracketMatcher(enabled: Boolean): Unit = {
+    if (bracketMatcherEnabled!= enabled) {
```
mrerrormessage: Minor code style tweak: add space between `bracketMatcherEnabled` and `!=`

netlogo-gui/src/main/ide/ShowUsageBox.scala
```diff
@@ -0,0 +1,204 @@
...
+
+  def getUsage(source: String, token: Token): Seq[Token] = {
+    val iterator = Femto.scalaSingleton[TokenizerInterface]("org.nlogo.lex.Tokenizer").tokenizeString(source)
+    var iter = scala.collection.mutable.Seq[Token]()
+    var prevLineNo = -1
+    val tokenLineNo = editorArea.offsetToLine(editorArea.getDocument.asInstanceOf[PlainDocument], token.start)
+    for(t <- iterator){
+      if( t.text.equalsIgnoreCase(token.text) &&
+        (editorArea.offsetToLine(editorArea.getDocument.asInstanceOf[PlainDocument], t.start)) != tokenLineNo &&
+        editorArea.offsetToLine(editorArea.getDocument.asInstanceOf[PlainDocument], t.start) != prevLineNo) {
```
mrerrormessage: This `for`-comprehension is fine, but for your knowledge it's possible to add the `if` as part of the comprehension, so:
```scala
for (t <- iterator) {
  if (meetsCondition(t)) {
    // some action...
  }
}
```

Becomes:
```scala
for (t <- iterator if meetsCondition(t)) {
  // some action...
}
```
You can do *a lot* in scala with `for`, and it's also OK to use alternatives instead of `for`.

netlogo-gui/src/main/ide/ShowUsageBox.scala
mrerrormessage: I'm still not seeing the line-striping on Windows or Mac. I'm OK with leaving it in for linux-only, but if we leave it in, leave a comment saying that this isn't visible on Windows or Mac as of mid-2016. That way if someone touches this code they won't worry that they broke something if these don't show up (or conversely if Oracle changes styles they'll know where to look).

mrerrormessage:
This looks great! Thanks for all your hard work on this. I left a few small comments on certain lines. The comment/uncomment functionality works great when the semicolon is at the beginning of the line, but if there's a gutter it doesn't remove leading semicolons. For instance, the "Wolf Sheep Predation" library model has

```NetLogo

to setup
  clear-all
  ask patches [ set pcolor green ]
  ;; check GRASS? switch.
  ;; if it is true, then grass grows and the sheep eat it
  ;; if it false, then the sheep don't need to eat
  if grass? [
     ...
  ]
end
```
When I highlight the three commented-out lines and "toggle comments" it adds a semicolon at the start of the line. I would instead expect that it would remove one or both of the existing semicolons. I really thought it used to work that way, not sure whether something got lost in my port from 5.x.

qiemem:
Not sure if this has been discussed, but currently, this doesn't respect scope of local variables and arguments. For instance:

image

Also, as also demonstrated in the screenshot, the line numbers in the show usage box are one off from the toggleable line numbers in the gutter.

qiemem:
A couple other things I think would be really great:

- Show line contex instead of line number. E.g. procedure name (if the line is in a procedure), globals (if a the line is in the globals declaration), turtles-own, etc.
- Show usages in widgets. If I hit show usages and nothing comes up, I think "great, I can delete this unused procedure"... but of course that doesn't work if it's invoked by a widget. Also great for figuring out where a variable declaration comes from. This would work well with the previous comment, where the context would say "button" or "slider" or whatever. Clicking on the usage would take you to the interface tab.

These may be outside the scope of this PR, in which case I'd be happy to take a crack at them later.

mrerrormessage:
Good catch on the line numbers, those definitely should match the ones in the gutter (think it's just a matter of adding 1, so it should be fine).

I think the other two items are out of scope for this PR, unless you're interested in working on them right now @Akarshit .

mrerrormessage:
With respect to local variable scoping, I would like to see the following approach taken:
* When "find usage" is activated, an attempt is made to parse the code in the code tab.
* If parsing succeeds, find usage looks at the parsing data and determines the extent of the variable in question, limiting usages to the appropriate scope.
* If parsing fails, find usage behaves the same as it currently does, showing all variables regardless of scope.

I know this is a big change, but I think it's quite important for this feature.

Akarshit:
All the features are quite important but I think the context and widget integrations would be better done in a separate PR. I will work on them when ASA I get some time.
I am also writing a small parsing code for the "Jump to definition" may be that would help with variable scope.
@mrerrormessage Do you need the scope resolution before merging this PR?

mrerrormessage:
That plan sounds good to me. I think it would be OK to merge this without scope resolution, as long as it gets handled with "Jump to definition". I *would* like to see line numbers adjusted in this PR to match line numbers in the gutter. After that we can go ahead and merge this in with the plan to handle scope as part of "Jump to definition".

mrerrormessage:
Looks great. I'm merging this in and will be looking forward to the jump to definition enhancements.

Issue #1077, opened 2016-06-24
Autosuggestion tweaks


qiemem:
This is for tweaks to the new autosuggestion popup outside of ordering (#1075) and expansion of source list (#1076).
- [ ] Add to command center.
- [ ] Make items on the box selectable with the mouse.

Idloj:
Filer out command or reporters based on context. For example, after `create-turtles ` there should be only reporters in the list, and in the start of a row (directly within the procedure) only commands should be listed.

Idloj:
Zebra-stripe the lines like in #1079

Akarshit:
@Idloj I tried the zebra stripes but that was not looking good. The main reason being that user will move up/down in the suggestion list frequently as opposed to a simple click in `Show Usage`. The screenshots of the suggestion box are below with zebra strips.
![screenshot from 2016-06-29 15 16 02](https://cloud.githubusercontent.com/assets/7762605/16448025/f6255c64-3e0c-11e6-8d82-ad75d0e13b68.png)
![screenshot from 2016-06-29 15 16 28](https://cloud.githubusercontent.com/assets/7762605/16448024/f625ce60-3e0c-11e6-8301-b57863d993ad.png)

mrerrormessage:
I agree that zebra-striping looks strange.

Idloj:
Another idea: if there is only one item, don't show the box and just complete the text in the only way available

mrerrormessage:
These issues look to have been resolved in 6.0

Idloj:
@mrerrormessage, can you leave this issue open? I believe my context-based-completion suggestion is still valid. I see this (wrongly, it seems) as an umbrella issue for whatever suggestions people come up with.

mrerrormessage:
I'd rather create a new issue. I would gladly copy-and-paste your comment into a new issue, or you're welcome to open one, whichever you would prefer.

PR #1080, opened 2016-06-27
Element can be chosen from the list with mouse click

Akarshit:
Completion of one of the task in #1077

### DIFF
--- a/netlogo-gui/src/main/ide/CodeCompletionPopup.scala
+++ b/netlogo-gui/src/main/ide/CodeCompletionPopup.scala
@@ -38,6 +38,12 @@ case class CodeCompletionPopup() {
       suggestionDisplaylist.setLayoutOrientation(JList.VERTICAL)
       suggestionDisplaylist.setVisibleRowCount(10)
       suggestionDisplaylist.setCellRenderer(new SuggestionListRenderer(eA))
+      
+      suggestionDisplaylist.addMouseListener(new MouseAdapter {
+        override def mouseClicked(e: MouseEvent): Unit = {
+          autoCompleteSuggestion(eA, autoSuggestDocumentListener)
+        }
+      })
 
       suggestionDisplaylist.addKeyListener(new KeyListener {
         override def keyTyped(e: KeyEvent): Unit = {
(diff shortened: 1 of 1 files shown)


netlogo-gui/src/main/ide/CodeCompletionPopup.scala
```diff
@@ -38,6 +38,20 @@ case class CodeCompletionPopup() {
       suggestionDisplaylist.setLayoutOrientation(JList.VERTICAL)
       suggestionDisplaylist.setVisibleRowCount(10)
       suggestionDisplaylist.setCellRenderer(new SuggestionListRenderer(eA))
+      
+      suggestionDisplaylist.addMouseListener(new MouseListener {
```
Idloj: You can use `MouseAdapter` here so you don't have to specify all the empty methods

netlogo-gui/src/main/ide/CodeCompletionPopup.scala
```diff
@@ -38,6 +38,20 @@ case class CodeCompletionPopup() {
...
+        override def mouseExited(e: MouseEvent): Unit = {}
+
+        override def mouseClicked(e: MouseEvent): Unit = {}
+
+        override def mouseEntered(e: MouseEvent): Unit = {}
+
+        override def mousePressed(e: MouseEvent): Unit = {}
+
+        override def mouseReleased(e: MouseEvent): Unit = {
```
Idloj: Why not on `mouseClicked`?

Akarshit: In the method that is adding the selected suggestion from the list of suggestions, I am making use of the selected element in the list. Now I was not sure will the new element get selected before `mouseClicked` is called or not, but I was pretty sure that the selected element would have changed till `mouseReleased` is called.

qiemem: `mouseClicked` also happens on release, as it were, but only if the press event happened on the same element (I believe). Thus it should be safe to use. Which one is better depends on the following: suppose a user presses the mouse button on one item, then moves the pointer to another and releases. Should the event trigger for the new item? Or should nothing happen?

I'm not sure which would be be better, tbh.

Akarshit: @qiemem I usually write code in `mouseReleased` because in my experience the place where I release the mouse is usually the right one. Let me know if this has to be changed.

PR #1081, opened 2016-06-27
Suggestion box while typing in the command line

Akarshit:
Added as according to #1077 
This PR also changes the location of the suggestion box based on the position of caret in the Editor Area. 
If there is not enough below the word, the list is displayed above it.


mrerrormessage:
This works well in the command center, but I don't see the popup go away after I press space. Is there anything you changed here that might have caused that difference in behavior? Is that happening for you on linux?

mrerrormessage:
I'd like to see some action on this PR soon. Also, we've had reports of people pressing ctrl+space and then typing and seeing the autocomplete appear (in the code tab, not the command center). Would you prefer to create a separate PR to deal with that or to address it in this one?

Akarshit:
I have modified this one so that the popup disappears after pressing space. Sorry for not mentioning this after the commit. 
I didn't exactly get the issue. Can you describe it to me once again.

mrerrormessage:
Excellent! I'll be reviewing this PR again shortly.

The particular issue is this:

* A user is in the code tab
* The user has a space behind the cursor
* A user presses Ctrl+space, nothing happens (this is correct, and expected)
* The user types a letter and sees the autocomplete box appear (this is not expected)
* The behavior this user expected was that no autocomplete box appears unless there is a character behind the cursor and the user presses Ctrl+space.

As I said, I'm happy to create another issue for this, just let me know. Also, if you feel the current behavior is correct, let me know and we can discuss.

mrerrormessage:
This looks good. It needs to be updated to not conflict with `hexy` (either merging in hexy or rebasing on hexy is fine). After that it can be merged.

.idea/uiDesigner.xml
```diff
@@ -0,0 +1,124 @@
+<?xml version="1.0" encoding="UTF-8"?>
```
mrerrormessage: This file should not be checked in to source control, can you add it to the `.gitignore` and remove from this PR?

extensions/array
```diff
@@ -1 +1 @@
-Subproject commit ac5f2d223b88e1d3b35a37c9dc081f4d974c1be9
+Subproject commit ce8575df722cf7852cd152106b1ee5914a286ce4
```
mrerrormessage: Did you change the submodules or did they change in hexy?

mrerrormessage:
This looks good other than the two code comments. On either this or a future PR I would like to see `WeakHashMap` switched out for a regular `Map` (of some kind) in `AutoSuggest.scala`. I get intermittent errors from the `WeakHashMap` when using auto complete and it causes our Travis tests to fail intermittently. If you would like to put it in this PR, please go ahead; otherwise please open a new PR sometime in the next day or two.

Akarshit:
I will put that in this PR only.

mrerrormessage:
Looks terrific, thanks!

PR #1082, opened 2016-06-27
Add a preferences dialog, including the "User Interface Language" option

Idloj:
Removed __change-language prim
Locale country in I18N is now loaded from "user.country"

I don't have a Mac, so I couldn't check that Application Menu > Preferences... shows and works correctly.
@mrerrormessage, it needs to be checked that the dialog shows correctly despite [this comment](https://github.com/NetLogo/NetLogo/blob/hexy/netlogo-gui/src/main/app/AboutWindow.scala#L88-93) (more precisely, despite the absence of that line there)
Should I remove `netlogo-gui/src/main/prim/etc/i18n.scala` entirely?

### DIFF
--- a/netlogo-gui/src/main/app/App.scala
+++ b/netlogo-gui/src/main/app/App.scala
@@ -277,6 +277,7 @@ class App extends
   lazy val owner = new SimpleJobOwner("App", workspace.world.mainRNG, AgentKind.Observer)
   private var _tabs: Tabs = null
   def tabs = _tabs
+  lazy val preferencesDialog = new PreferencesDialog(frame, Preference.Language)
   var helpMenu:HelpMenu = null
   var fileMenu: FileMenu = null
   var monitorManager:AgentMonitorManager = null
@@ -598,7 +599,6 @@ class App extends
           org.nlogo.log.Files.deleteSessionFiles(System.getProperty("java.io.tmpdir"))
         else
           logger.deleteSessionFiles()
-      case CHANGE_LANGUAGE => changeLanguage()
       case _ =>
     }
   }
@@ -627,21 +627,6 @@ class App extends
         org.nlogo.awt.EventQueue.invokeLater(() => openFromSource(source, path, ModelType.Library))
       }
     }
-  }
-
-  def changeLanguage() {
-    val locales = I18N.availableLocales
-    val languages = locales.map{l => l.getDisplayName(l) }
-    val index = org.nlogo.swing.OptionDialog.showAsList(frame,
-      "Change Language", "Choose a new language.", languages.asInstanceOf[Array[Object]])
-    if(index > -1) {
-      val chosenLocale = locales(index)
-      val netLogoPrefs = java.util.prefs.Preferences.userRoot.node("/org/nlogo/NetLogo")
-      netLogoPrefs.put("user.language", chosenLocale.getLanguage)
-      netLogoPrefs.put("user.country", chosenLocale.getCountry)
-    }
-    val restart = "Langauge changed.\nYou must restart NetLogo for the changes to take effect."
-    org.nlogo.swing.OptionDialog.show(frame, "Change Language", restart, Array(I18N.gui.get("common.buttons.ok")))
   }
 
   ///
@@ -847,6 +832,13 @@ class App extends
     showAboutWindow()
   }
 
+  /**
+   * This is called reflectively by the mac app wrapper.
+   */
+  def handleShowPreferences(): Unit = {
+    showPreferencesDialog()
+  }
+
   @throws(classOf[java.io.IOException])
   def libraryOpen(path: String) {
     dispatchThreadOrBust(fileMenu.openFromPath(path, ModelType.Library))
@@ -1045,7 +1037,16 @@ class App extends
    * Internal use only.
    */
   // used both from HelpMenu and MacHandlers - ST 2/2/09
-  def showAboutWindow() { new AboutWindow(frame).setVisible(true) }
+  def showAboutWindow(): Unit = {
+    new AboutWindow(frame).setVisible(true)
+  }
+
+  /**
+   * Internal use only.
+   */
+  def showPreferencesDialog(): Unit = {
+    preferencesDialog.setVisible(true)
+  }
 
   /**
    * Internal use only.
--- a/netlogo-gui/src/main/app/Preference.scala
+++ b/netlogo-gui/src/main/app/Preference.scala
@@ -1,0 +1,41 @@
+// (C) Uri Wilensky. https://github.com/NetLogo/NetLogo
+
+package org.nlogo.app
+
+import java.util.Locale
+import java.util.prefs.Preferences
+import javax.swing.{ JComboBox, JComponent }
+
+import org.nlogo.core.I18N
+
+trait Preference {
+  val i18nKey: String
+  val component: JComponent
+  val restartRequired: Boolean
+  def load(prefs: Preferences): Unit
+  def save(prefs: Preferences): Unit
+}
+
+object Preference {
+  object Language extends Preference {
+    case class LocaleWrapper(val locale: Locale) {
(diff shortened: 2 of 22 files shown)


mrerrormessage:
Looks awesome! I'm going to try to review this tomorrow.

netlogo-gui/src/main/app/Preference.scala
```diff
@@ -0,0 +1,41 @@
...
+  val i18nKey: String
+  val component: JComponent
+  val restartRequired: Boolean
+  def load(prefs: Preferences): Unit
+  def save(prefs: Preferences): Unit
+}
+
+object Preference {
+  object Language extends Preference {
+    case class LocaleWrapper(val locale: Locale) {
```
Idloj: The best name I came up with. If I should change it to something else, let me know

netlogo-gui/src/main/app/PreferencesDialog.scala
```diff
@@ -0,0 +1,76 @@
...
+  JComponent, JDialog, JLabel, WindowConstants, SwingConstants }
+
+import org.nlogo.awt.Positioning
+import org.nlogo.core.I18N
+import org.nlogo.swing.{ OptionDialog, RichAction, TextFieldBox, Utils => SwingUtils }
+import org.nlogo.swing.Implicits._
+
+object PreferencesDialog {
+  private val restartIcon =
+    new ImageIcon(classOf[PreferencesDialog].getResource("/images/forever2.gif"))
```
Idloj: This looks great IMO, but maybe you guys want a dedicated icon for this?

netlogo-gui/src/main/app/PreferencesDialog.scala
```diff
@@ -0,0 +1,76 @@
...
+}
+
+class PreferencesDialog(parent: Frame, preferences: Preference*)
```
Idloj: The actual preferences are injected from App. Is that OK, or should PreferencesDialog construct the list of preferences itself?

netlogo-gui/src/main/app/App.scala
```diff
@@ -847,6 +832,13 @@ class App extends
     showAboutWindow()
   }
 
+  /**
+   * This is called reflectively by the mac app wrapper.
+   */
+  def handleShowPreferences(): Unit = {
+    showPreferencesDialog()
```
Idloj: I basically followed the way handleShowAbout is done, but why is this indirection good? Why not have mac-app call `showPreferencesDialog` and `showAboutMenu`?

netlogo-gui/src/main/app/PreferencesDialog.scala
```diff
@@ -0,0 +1,76 @@
...
+    setVisible(false)
+  }
+  private val apply = () => preferences foreach (_.save(netLogoPrefs))
+  private val cancel = () => {
+    reset()
+    setVisible(false)
+  }
+
+  private val preferencesPanel = new TextFieldBox(SwingConstants.TRAILING)
+  preferencesPanel.setBorder(BorderFactory.createEmptyBorder(20, 10, 20, 10))
```
Idloj: This can be changed on will, of course. It just looks good for me.

netlogo-gui/src/main/app/PreferencesDialog.scala
mrerrormessage: I think this is fine for now.

netlogo-gui/src/main/app/App.scala
mrerrormessage: I want to keep these methods separate so that we don't break the mac-app wrapper when changing them for some other reason.

Idloj: How can it break mac-app in a way that would not break if called by handleShowAbout?

mrerrormessage: So let's say that for some reason the `HelpMenu` wants to pass an argument into `showAboutMenu`. The author of the change adds an argument to `showAboutMenu` and recompiles. He sees that `handleShowAbout` doesn't compile anymore. He looks at the method and realizes that he has to provide some sort of argument in order for his change to work on mac. In the hypothetical world where he doesn't have `handleShowAbout` to catch his mistake he makes the change and recompilation succeeds, he goes on his way, *never having noticed that he broke NetLogo on Mac*. you could certainly put comments on the other methods, but nothing gets people's attention like a broken compile.

mrerrormessage:
Thanks for the PR! I think the appearance of the dialog looks good. Can you add the text "(requires restart)" in the preferences box somewhere around "User Interface Language", just so users get a head-up? Also, if there's nothing that `Preferences`/`PreferencesDialog` need in `org.nlogo.app` I would probably slightly prefer that they live in `org.nlogo.window` instead. If they depend on things in `org.nlogo.app`, they can stay there.

Testing on mac showed that the dialog wouldn't pop to the foreground if there were other dialogs up as well, so we **do** need those lines for sure.

netlogo-gui/src/main/app/App.scala
Idloj: ok, thanks

Idloj:
I don't know if you've noticed but there is a tooltip that says "You must restart NetLogo for the change to take effect"

netlogo-gui/src/main/app/Preference.scala
mrerrormessage: This name is good.

mrerrormessage:
I see the tooltip now that you mention it. A tooltip is a fine start, but I would prefer that it be a bit clearer. Typically when people see "Apply" they assume that updates will be reflected immediately, we don't want people to think that language selection is broken just because they never restarted NetLogo.

netlogo-gui/src/main/app/PreferencesDialog.scala
mrerrormessage: Injecting preferences seems like a good approach here.

Idloj:
Should I remove the icon an the tooltip then (only showing '(restart required)')?

mrerrormessage:
Oh, I totally missed that the icon was indicating restart (although it seems obvious now that you've told me). I don't think that the icon for that is standardized enough to be used without text. You could put some kind of a footnote in the panel like "<icon> = restart required" or just replace the icon/tooltip with "(restart required)" - up to you.

Idloj:
I thought of the first option, but it seemed strange to do this for only one preference, so I guess I would go for option 2, keeping option 1 for later where there will be more settings in the dialog

Idloj:
Updated

Idloj:
side questions:
a) why is the `window` package better? isn't it supposed to be for Interface Tab stuff (that's what the Package Guide says)?
b) why is it better for the preferences to be injected and not have `val preferences = List(Preference.Language)` in PreferencesDialog itself?

mrerrormessage:
* Maybe the `window` package isn't the appropriate place for it. I don't want the `app` package to become a junk drawer. Let's leave it in `app` for now.
* I think injecting the preferences is better because it allows for more flexibility. For instance this makes it easier to transition to a design where preferences can be registered with pico or `App` at runtime, which we may want at some point. Most importantly (in my mind) we don't have to change `PreferencesDialog` each time we add a new preference.

Idloj:
Maybe `app` can be split up to several smaller packages and have `app` depend on them (or have them nested inside) ? For example, we can dedicate a package to each tab, so there will be `window` for Interface, `info` for Info, and `ide` for Code. Maybe also `tools` where the PreferencesDialog and Command Center stuff will live (maybe more stuff that I miss).

I'm happy to do that if this suggestion is accepted (in this form or another), on a different PR, of course.

mrerrormessage:
I think splitting it up is a terrific idea! I'm not sure what the names should be (I'll think about that, ask me about it if I haven't gotten back to you by tomorrow), but in view of that this PR doesn't need to be changed to support the package. I'll retest on mac in a bit and then merge in assuming everything looks good.

PR #1084, opened 2016-06-27
Pressing Esc on the shapes editor and closing the window both confirm cancel

Idloj:
(The previous behavior of closing the window was saving the shape)


mrerrormessage:
👍 looks great!

PR #1088, opened 2016-06-30
Add the option to Undo actions taken on a Widget

Akarshit:
This is the initial PR so that @mrerrormessage can review the code structure. Though the undo works there are still a lot of bugs to fix and cleaning up to do.


netlogo-gui/src/main/app/InterfaceToolBar.scala
```diff
@@ -34,6 +34,18 @@ class InterfaceToolBar(wPanel: WidgetPanel,
   private val deleteAction = new DeleteAction()
   private val deleteButton = new ToolBarActionButton(deleteAction)
   private val widgetMenu = new WidgetMenu
+  private val undoButton = new JButton("Undo")
```
mrerrormessage: This should subclass `ToolBarActionButton` or a related class so that it has the same appearance as the other buttons.

mrerrormessage:
I don't see where `ChangeManager` is defined, did you forget to add it?

mrerrormessage:
So `ChangeManager` seems important and I would definitely like to have a look at it, however this seems pretty decent generally.

* I want to think more about whether we want to have these new `Action`s be defined on `WidgetWrapper` or on `CoreWidget`. The advantage of `WidgetWrapper` is that it matches `WidgetPanel`, the advantage of `CoreWidget` is that it isn't tied to `WidgetPanel`.
* Similarly to `WidgetWrapper`, I'm a bit worried about adding a lot of new code that depends on `WidgetPanel`. I'm going to open a PR soon (maybe as early as this afternoon) which converts `WidgetPanel` to scala (and I would definitely like to see that change incorporated before this PR), but I want to think about adding some sort of interface which `WidgetPanel` implements.

Sorry for the vague feedback. I'm hoping that working through converting `WidgetPanel` will help me make some of these decisions.

netlogo-gui/src/main/app/InterfaceToolBar.scala
Akarshit: They are just for testing, I don't even plan to add these.

Akarshit:
@mrerrormessage Sorry I forgot to attach the file(it was newly created so showed in the untracked category). It has been added now.

netlogo-gui/src/main/app/ChangeManager.scala
```diff
@@ -0,0 +1,138 @@
+package org.nlogo.app
+
+import java.util.List
+import java.util.ArrayList
+import javax.swing.event.ChangeEvent
+/**
+  * Manages a Queue of Changables to perform undo and/or redo operations. Clients can add implementations of the Changeable
+  * class to this class, and it will manage undo/redo as a Queue.
+  *
+  * @author Greg Cope
```
mrerrormessage: Can you provide the URL of the site where you found this? We need to be sure when we pull in code that it's appropriate for us to be using it.

Akarshit: http://www.algosome.com/articles/implementing-undo-redo-java.html
I went through their copyright page just now, looks pretty tight. May be you can go through that and understand it better than I did.

Idloj: Can't the UndoManager be used to accomplish the task (by subclassing maybe. maybe it's not necessary) ?

mrerrormessage: OK, the copyright is definitely not acceptable for use in NetLogo. We need to remove this file immediately. You can see whether `UndoManager` will work, perhaps it's generalizable enough to fit here. If it won't we can write something with similar behavior to this, but it needs to be original code written without reference to this code.

Akarshit: OK I will try to use the `UndoManager`.

netlogo-gui/src/main/app/WidgetActions.scala
```diff
@@ -1,57 +1,104 @@
 package org.nlogo.app
 
 import java.awt.Rectangle
+import javax.swing.JComponent
+import javax.swing.undo.{AbstractUndoableEdit, UndoManager, UndoableEdit}
 
 import org.nlogo.core.Widget
+import org.nlogo.window.MouseMode
+
+class WidgetUndoManager extends UndoManager{
```
Idloj: I think there is no need to define this class (as you're overriding two methods by just calling the `super`'s counterparts, which I don't see the point of), you can just use UndoManager directly.

Akarshit: @Idloj Yes right now there is no use but I plann to use it in the future. The current state of code is very bad and I only pushed it so that @mrerrormessage can get a idea of what structure I am making.

Idloj: ok, sorry

netlogo-gui/src/main/app/WidgetActions.scala
```diff
@@ -0,0 +1,166 @@
...
+
+  private var counter = 0
+  private val undoManager = new UndoManager() {
+    override def addEdit(anEdit: UndoableEdit): Boolean = {
+      undoButton.setEnabled(true)
+      super.addEdit(anEdit)
+    }
+  }
+
+  var undoButton : ToolBarButton = null
```
mrerrormessage: I think you can get around needing to operate on `undoButton` / `redoButton` just by operating on the `undo`/`redo` actions. You can set an icon for an action and use `setEnabled(isEnabled)` to enable / disable an action (which should enable/disable the button).

netlogo-gui/src/main/app/WidgetActions.scala
```diff
@@ -0,0 +1,166 @@
...
+      }
+    }
+    override def undo(): Unit = {
+      for(widgetWrapper <- wws){
+        val initialBound = new Rectangle(initialMap(widgetWrapper))
+        if(!widgetWrapper.selected()) {
+          initialBound.x += WidgetWrapper.BORDER_E
+          initialBound.width -= WidgetWrapper.BORDER_E + WidgetWrapper.BORDER_W
+          initialBound.y += WidgetWrapper.BORDER_N
+          initialBound.height -= WidgetWrapper.BORDER_N + WidgetWrapper.BORDER_S
```
mrerrormessage: These four lines show up 6 times in this file (or if they aren't the same four lines, they're very similar). A helper method would simplify the code a lot.

netlogo-gui/src/main/app/InterfacePanel.scala
```diff
@@ -155,7 +149,7 @@ class InterfacePanel(val viewWidget: ViewWidgetInterface, workspace: GUIWorkspac
     }
   }
 
-  override private[app] def deleteWidgets(hitList: Seq[WidgetWrapper]): Unit = {
+  override def deleteWidgets(hitList: Seq[WidgetWrapper]): Unit = {
```
mrerrormessage: Is there a reason to make this public?

mrerrormessage:
This is looking good. I left a few comments. Do you think it would be a good idea (and possible) to add ctrl-z ctrl-y shortcuts? It could get sort of confusing with the editor, but I think it would be more convenient in general.

Akarshit:
@mrerrormessage I have made the following changes
1. The focus system has been changed to what we discussed (You should check this)
2. The undo/redo has been added to edit menu
3. Added the shortcuts of `Ctrl Z` and `Ctrl Y`
4. Named the options as "<Undo/Redo> <Widget/Widget(s)> <Addition/Deletion/Resizing/Movement> (Perhaps we can use better/shorter names here).

netlogo-gui/src/main/app/InterfacePanel.scala
```diff
@@ -331,11 +333,11 @@ class InterfacePanel(val viewWidget: ViewWidgetInterface, workspace: GUIWorkspac
   def keyPressed(evt: KeyEvent): Unit = {
```
Idloj: Why do you need to handle the shortcuts yourself? Why not deleting the whole `keyPressed` functionality and allowing the standard menu-items in the Edit menu to do the undo/redo? you're calling setCurrentManager, so I guess it should work fine

netlogo-gui/src/main/app/InterfacePanel.scala
```diff
@@ -344,7 +329,16 @@ class InterfacePanel(val viewWidget: ViewWidgetInterface, workspace: GUIWorkspac
     }
   }
 
-  def keyPressed(evt: KeyEvent): Unit = { }
+  def keyPressed(evt: KeyEvent): Unit = {
```
Idloj: Why do you handle Ctrl+Z / Ctrl+Y yourself? why not deleting the `keyPressed` functionality and letting the Edit menu take care of undo/redo using its menu-items?

mrerrormessage: @Akarshit , is that possible to do here? I agree that would be a good change to make, but I know that it has been troublesome to handle conflicting undo/redo code for the command center in addition to the new undo/redo  functionality introduced in this PR.

netlogo-gui/src/main/app/WidgetPanel.scala
```diff
@@ -625,14 +620,6 @@ class WidgetPanel(val workspace: GUIWorkspace)
       component.setSize(size)
   }
 
-  /// buttons
-
-  protected def loseFocusIfAppropriate(): Unit = {
```
mrerrormessage: Can this method be deleted?

mrerrormessage:
This seems better, but I'm running into a few problems:
* When I add a button and then "undo" adding a button, then "redo" adding the button, the command the button's command is empty. It's possible this also occurs with other widgets.
* After undoing / redoing a widget edit, focus is returned to the command center (on my computer, it's possible this is mac-specific). This makes it really awkward to do anything substantial with undo/redo as I have to click on the interface area, click or type undo/redo, click on the interface area, click or type undo/redo, etc.

Akarshit:
I will look into these. The focus problem is not there on Linus, but I will try to search for a solution to this.

netlogo-gui/src/main/app/WidgetPanel.scala
Akarshit: I didn't get you, are you asking me if it is safe to delete this?

mrerrormessage: I was. That was a silly question - given that it's already been deleted! Nothing more to be done here!

Akarshit:
@mrerrormessage What do you mean by
> click or type undo/redo

How do you execute undo by typing it?

mrerrormessage:
By typing I mean using keyboard shortcuts.

netlogo-gui/src/main/app/InterfacePanel.scala
Akarshit: @Idloj Great idea! Thanks :)

netlogo-gui/src/main/app/InterfaceTab.scala
```diff
@@ -69,12 +75,20 @@ class InterfaceTab(workspace: GUIWorkspace,
...
+    }
+    else {
+      lastFocusedComponent.requestFocus
+    }
   }
 
   final def handle(e: Events.SwitchedTabsEvent) {
-    commandCenterAction.setEnabled(e.newTab == this)
+    if(e.newTab != this) {
+        lastFocusedComponent = if(iP.hasFocus) iP else commandCenter
```
mrerrormessage: This is indented too far.

netlogo-gui/src/main/app/WidgetActions.scala
```diff
@@ -0,0 +1,145 @@
...
+    // The default one doesn't work!
+    override def canRedo: Boolean = {
+      if(editToBeRedone() != null) {
+        return true
+      }
+      false
+    }
+  }
+
+  def addWidget(widgetPanel: WidgetPanel, coreWidget: Widget, x: Int, y: Int): Unit ={
```
mrerrormessage: Space between `=` and `{` (throughout this file, also in `WidgetPanel.reAddWidget`).

mrerrormessage:
So I tested a new situation that may be problematic:

* Add a monitor for `abc` - the monitor will error as expected
* Add a slider for `abc` - the monitor will stop erroring, as expected
* Delete the slider for `abc` - the monitor will start erroring again, as expected
* Undo deletion of the slider for `abc` - the monitor will continue to error, which is unexpected. Since there is a slider for `abc` the monitor should work as expected.

I believe that this is caused by some event or events not being raised at the appropriate time. Can you see if you can fix this problem?

I'm still having trouble with the focus on Mac. I think it's due to the differences in the way Mac handles menus. I'm going to look into it and see if I can figure it out, I'll check in with you tomorrow and let you know how things are going.

Akarshit:
@mrerrormessage I am working on the focus problem. It is happening in Linux too, I have fixes it for all the cases expect one. Working on that now.

mrerrormessage:
OK, great! I'll leave that to you until I hear differently.

Akarshit:
@mrerrormessage I am not seeing the focus problem on Linux(perhaps I was on another branch when I typed the above comment)

mrerrormessage:
Thanks for working more on this. The focus is working as expected.

Here are some things that need to happen before this is merged or I re-review:
* This must be updated against hexy. At the time you pushed it might have been up-to-date, but it's showing conflicts right now.
* When I add a button or a monitor with code and then "undo" adding that widget, then "redo" adding that widget, the code for that widget is empty. Please look into this and figure out what's happening. Also, check whether similar problems effect other types of widgets.

I would also like you to investigate the following:
* Suggestions around focus switching which I'll leave in the code. You did fix the focus bug, but I think it might be possible to do it slightly differently (details in code comments).
* For many widgets, the various properties (like name, size, etc) effect the widgets size on screen. Can you see if it's possible to have edits to widget property tracked by the undo manager?

netlogo-gui/src/main/app/InterfaceTab.scala
```diff
@@ -2,12 +2,19 @@
...
 import org.nlogo.swing.Implicits.thunk2action
-import javax.swing.{JScrollPane, ScrollPaneConstants, Action, ImageIcon, AbstractAction, JSplitPane, JPanel}
-import java.awt.{Graphics2D, Graphics, Component, Container, ContainerOrderFocusTraversalPolicy, Dimension, BorderLayout}
+import javax.swing._
+import java.awt.{BorderLayout, Component, Container, ContainerOrderFocusTraversalPolicy, Dimension, Graphics, Graphics2D}
+
+import org.nlogo.editor.UndoManager
 
+object InterfaceTab {
+  var interfacePanelHadFocus = false
```
mrerrormessage: I want to get rid of this static variable if at all possible. In a prior approach, you had an instance variable, which I would much prefer here.

netlogo-gui/src/main/app/InterfaceTab.scala
```diff
@@ -69,12 +77,22 @@ class InterfaceTab(workspace: GUIWorkspace,
   def getInterfacePanel = iP
 
   override def requestFocus() {
-    if(iP.isFocusable && splitPane.getDividerLocation >= maxDividerLocation) iP.requestFocus()
-    else if(commandCenter != null) commandCenter.requestFocus()
+    if(iP.isFocusable && splitPane.getDividerLocation >= maxDividerLocation) {
+      iP.requestFocus
+    }
+    else if(InterfaceTab.interfacePanelHadFocus){
+      iP.requestFocus
```
mrerrormessage: This is a side-effecting method and should have empty parens `()` at the end.

netlogo-gui/src/main/app/InterfaceTab.scala
```diff
@@ -69,12 +77,22 @@ class InterfaceTab(workspace: GUIWorkspace,
...
+    } else {
+      commandCenter.requestFocus()
+    }
   }
 
   final def handle(e: Events.SwitchedTabsEvent) {
-    commandCenterAction.setEnabled(e.newTab == this)
+    if(e.newTab != this) {
+        InterfaceTab.interfacePanelHadFocus = iP.hasFocus
```
mrerrormessage: [`Component.hasFocus`](https://docs.oracle.com/javase/7/docs/api/java/awt/Component.html#hasFocus()) has been obsoleted by Oracle in favor of [`Component.isFocusOwner()`](https://docs.oracle.com/javase/7/docs/api/java/awt/Component.html#isFocusOwner()). Let's use `isFocusOwner` here.

netlogo-gui/src/main/app/WidgetActions.scala
```diff
@@ -0,0 +1,145 @@
...
+import javax.swing.undo.{AbstractUndoableEdit, UndoManager, UndoableEdit}
+
+import org.nlogo.core.Widget
+import org.nlogo.swing.ToolBarButton
+import org.nlogo.window.MouseMode
+
+object WidgetActions {
+
+  val undoManager = new org.nlogo.editor.UndoManager() {
+    // The default one doesn't work!
```
mrerrormessage: Can you explain this comment? If you just mean that `canRedo` needs an override, let's just create an inner class for that. Otherwise, let's fix the problems with `org.nlogo.editor.UndoManager`.

netlogo-gui/src/main/app/WidgetActions.scala
```diff
@@ -0,0 +1,145 @@
...
+    override def canRedo: Boolean = {
+      if(editToBeRedone() != null) {
+        return true
+      }
+      false
```
mrerrormessage: This method can be rewritten as a one-liner. Any time I find myself returning a `true`/`false` from a conditional, I try to think about whether I can make the conditional predicate (`editToBeRedone() != null`, in this case) do the work instead of writing out the entire conditional.

netlogo-gui/src/main/app/WidgetActions.scala
```diff
@@ -0,0 +1,145 @@
...
+    }
+    override def undo(): Unit = {
+      for(widgetWrapper <- wws){
+        val initialBound = new Rectangle(initialMap(widgetWrapper))
+        if(!widgetWrapper.selected()) {
+          removeSelectionMargin(initialBound)
+        }
+        widgetWrapper.setBounds(initialBound)
+      }
+    }
```
mrerrormessage: `undo` and `redo` are the same method here except for `initialMap`/`finalMap`. Extract a method to avoid repeating code.

netlogo-gui/src/main/app/WidgetActions.scala
SethTisue: also note that you should basically never use `return` in Scala

netlogo-gui/src/main/app/WidgetActions.scala
```diff
@@ -0,0 +1,145 @@
...
+      }
+      widgetWrapper.setBounds(fb)
+    }
+    override def undo(): Unit = {
+      val ib = new Rectangle(initialBounds)
+      if(!widgetWrapper.selected()) {
+        removeSelectionMargin(ib)
+      }
+      widgetWrapper.setBounds(ib)
+    }
```
mrerrormessage: `undo` and `redo` are the same method except for one variable. Extract a method with the common functionality.

netlogo-gui/src/main/app/WidgetPanel.scala
```diff
@@ -451,6 +441,24 @@ class WidgetPanel(val workspace: GUIWorkspace)
...
+    add(widgetWrapper, javax.swing.JLayeredPane.DEFAULT_LAYER)
+    moveToFront(widgetWrapper)
+    widgetWrapper.validate()
+    widgetWrapper.setVisible(true)
+
+    zoomer.zoomWidget(widgetWrapper, true, false, 1.0, zoomFactor)
+    new CompileAllEvent().raise(this)
+    widgetWrapper.requestFocusInWindow()
+
+//    Logger.logAddWidget(widget.classDisplayName, widget.displayName)
```
mrerrormessage: Remove or uncomment this line.

netlogo-gui/src/main/app/InterfaceTab.scala
```diff
@@ -78,14 +80,16 @@ class InterfaceTab(workspace: GUIWorkspace,
...
+    else if(InterfaceTab.interfacePanelHadFocus){
+      iP.requestFocus
+    } else {
+      commandCenter.requestFocus()
     }
   }
 
   final def handle(e: Events.SwitchedTabsEvent) {
     if(e.newTab != this) {
-        lastFocusedComponent = if(iP.hasFocus) iP else commandCenter
```
mrerrormessage: I really prefer this `lastFocusedComponent` approach to the `InterfaceTab.interfacePanelHadFocus` approach. If you added a method to `InterfacePanel` to make `_hasFocus` accessible and used that here instead of `hasFocus`, could you make this approach work? If that won't work, could you have `InterfaceTab` add a `FocusListener` to each of `CommandCenter` and `InterfacePanel` and have the listener(s) change an instance variable in `InterfaceTab`?

Basically, we want to try every possible alternative approach before adding a public static variable like `InterfaceTab.interfacePanelHadFocus`. Public static variables make it really hard to debug code and really hard to change code. If neither of the above approaches work, please let me know and we'll see if we can come up with another alternative.

Akarshit: OK sure I will try to make the `lastFocusedComponent` approach work.

netlogo-gui/src/main/app/interfacetab/WidgetActions.scala
```diff
@@ -0,0 +1,138 @@
...
+import java.awt.Rectangle
+import javax.swing.undo.AbstractUndoableEdit
+
+import org.nlogo.core.Widget
+
+object WidgetActions {
+
+  val undoManager = new org.nlogo.editor.UndoManager() {
+    // The default one doesn't work!
+    override def canRedo = editToBeRedone() != null
```
Akarshit: the `canRedo` method of the UndoManager class is not working here. I tried to debug it, but was unable to get a solutions. So I mentioned it here in case someone wondered why the method has been overridden.

Akarshit:
> For many widgets, the various properties (like name, size, etc) effect the widgets size on screen. Can you see if it's possible to have edits to widget property tracked by the undo manager?

@mrerrormessage What is the final aim of this?

mrerrormessage:
> What is the final aim of this?

The final aim of this is that editing is treated like all other operations on widgets. In other words, if I have a widget in the interface panel, I right click, edit and save when I choose undo then the widget should revert the changes made in the edit. I'm happy to clarify further if necessary.

Akarshit:
OK i get it. But I think that might be out of the scope of this PR. Consider these 2 cases. 
1. User adds a button, sets a display name for it, puts a command in it, presses "OK". Now if the user wants to _undo_, don't you think his intention would be to "undo addition of the widget" instead of "undoing all name and command changes within the widget".
2. User adds a button, sets a display name for it, puts a command in it, presses "OK", edits the command in the button again, moves the button some place else(many times). Now he start undoing. There is no way for him to see when the _command_ inside the button would be changed. His intention was to just bring the button back to the original location, but just one undo more could change the command in the button.

What approach I think we should take(and have taken) is when a user is editing(the widget properties box is opened) a widget then undo the edits. If the user is on the interfaceTab(no edit for any widget is open), undo the things like addition, deletion, movement, resizing.

mrerrormessage:
I think those are some good points. I think the best thing we can do at the moment is get this in front of people and hear back from them.

This PR looks really great! I'm going to merge it later this week because of some stuff I'm working on today, but it is ready to be merged.

PR #1091, opened 2016-07-01
Add "Export code" to allow expoting of the model's code as HTML

Idloj:
I need https://github.com/NetLogo/NetLogo/pull/1091/commits/99169abcf06746bb6486b60ddf909edb92656cba to be reviewed closely to make sure I documented the missing menu-items right, and the additional information on the Tabs menu is worded correctly


autogen/docs/interface.html.mustache
```diff
@@ -449,7 +469,7 @@
           <td class="descriptionRight">
         <p>
             Opens the preferences dialog, where you can customize settings
-            regarding NetLogo.
+            regarding NetLogo. (On a Mac, this item is on the NetLogo menu instead.)
```
mrerrormessage: Don't need to put parentheses around "On a Mac..."

autogen/docs/interface.html.mustache
```diff
@@ -675,26 +707,16 @@
           <td style='vertical-align: top; border-top:none;border-left: none;border-bottom:solid windowtext .5pt;border-right:none; padding:0in 5.4pt 0in 5.4pt'>
         <td class="descriptionRight">
         <p>
-            This menu offers keyboard shortcuts for each of the tabs. (On
-            Macs, it's Command 1 through Command 3. On Windows,
-            it's Control 1 through Control 3.)
+            This menu offers keyboard shortcuts for each of the tabs. (On a
+            Mac, it's Command 1 through Command 3. On Windows,
+            it's Control 1 through Control 3. Additional numbers are used for
+            tabs containing ".nls" files)
```
mrerrormessage: Add period after "files"

mrerrormessage:
The updated menu documentation looks great - thanks for updating that (I've left a comment on one or two items that need small tweaks, but looks good overall.

This looks great overall, I would like a few changes before it is merged in:
* Travis shows that the tests don't compile, the tests must compile and pass.
* Ellipsis appears at the end of the various I18N translations where it has been removed in English. If it is removed from English (which seems good to me) we should remove it from all of them.

Idloj:
Updated

Idloj:
Travis fails on AutoSuggestTests, no idea how it might be related to what I've done

mrerrormessage:
Yeah, that's unrelated. I'm re-running and I'll merge when the tests finish.

PR #709, opened 2016-07-06
Declaring a breed with only a plural name is allowed by the compiler

mrerrormessage:
Example:

    breed [ xs ]

This is currently undocumented and exists for backward compatibility: NetLogo 3 (and earlier?) models had only plural breed names. When they're automatically converted to NetLogo >=4, `breed [ xs ]` is what the converter generates.

Going forward, options include:

- Documenting this possibility;
- Removing it (and dealing with the implications for automatic conversion of very old models);
- Keeping the _status quo_ (with the difference that this ticket now exists as some sort of documentation for the situation).


SethTisue:
🎉

Issue #1075, opened 2016-06-24
Autosuggest suggestion order tuning


mrerrormessage:
While working on #1072 , we didn't fine-tune the order of results. This can be done by modifying `ide/AutoSuggest.scala`. I would like to involve some experienced modelers in tuning the result sets.

qiemem:
Okay, some I'm still playing around, but the big one is that exact prefix matches must come before non-exact. I'm actually wondering if we should, for now, just only allow exact prefix matches (thus dispensing with the edit distance stuff entirely).

That said, you can do some pretty cool stuff with the edit distance stuff. My favorite so far is to take only matches with the same starting character. Then, we filter out anything that requires an edit or delete (as opposed to an insert or append). Finally, we sort by edit distance, with append having a crazy low weight. This makes exact prefix matches come first, but then also gives you, for instance, `user-new-file` if you do `unf`. I'm not sure this is actually a good idea, but I think it's pretty neat and shows the flexibility of the system.

qiemem:
How would people feel if we made this just exact prefix matching for now, ditching the edit distance stuff? That seems the simplest way to do this and get good enough results.

Akarshit:
Here are some ideas I had
* Increase the error threshold (like threshold edit-distance becomes 1 after 4 character have been typed).
* Modify edit distance algo to take distance between keys and then calculate the edit distance.
* Rank according to the edit distance given by previous algorithm.
* Use context to suggest words (like when reporters or command are likely to be typed).
* Implement the `unf -> user-new-file` kind of completion separately.

qiemem:
> Increase the error threshold (like threshold edit-distance becomes 1 after 4 character have been typed).

I really don't think this is a good idea. While I think this is a cool idea, in practice, I found it much more annoying than helpful.

> Use context to suggest words (like when reporters or command are likely to be typed). 

This would be awesome, but sounds hard.

> Rank the keywords manually according to usage and then rank the suggestion according to this (I know this is a pretty wild one).

Interesting, but again, I'm not sure it would actually work well in practice. Having a reliable order so that you can quickly select known things is really helpful.

PR #1093, opened 2016-07-09
Enhancements to AutoSuggest

Akarshit:
Added the `fca => file-close-all`.
Increased the threshold limit so that prefix matches are preferred more.
In relation to #1075


qiemem:
Ah, ya, I posted some thoughts to #1075.

> Increased the threshold limit so that prefix matches are preferred more.

I still think it might be better to just straight up remove the edit distance stuff. Auto-complete alone already acts as a kind of spellcheck. Predictability is really important in systems like this, and the edit distance stuff hurts predictability.

Regardless, prefix matches should *always* be preferred, by sorting.

This is quit jarring:

screen shot 2016-07-11 at 9 51 47 am

screen shot 2016-07-11 at 9 52 04 am

Typing another character should decrease the number of items on the list, not increase.

> Added the `fca => file-close-all`.

Cool!

PR #25, opened 2016-07-11
OutOfMemoryErrors should be handled more gracefully

mrerrormessage:
in particular, we don't need people emailing their stack traces from this to the bugs address


SethTisue:
nice. this has been an ongoing annoyance for a decade and a half

Issue #853, opened 2015-09-17
Bundle 2.5d extension


mrerrormessage:

mrerrormessage:
@qiemem , were you planning to make this change or should it be something for me to do? I'm happy to work on it, I just need to know. In addition to actually including 2.5d in the build, 2.5d will need to be updated for the JOGL upgrade. I can provide additional guidance as needed, but the changes are (for the most part) renames [see this commit for examples](https://github.com/NetLogo/NetLogo/commit/868499af94fdd16cff9d850e9552f03d74fd93bf).

mrerrormessage:
Closed in #1098 .

PR #1098, opened 2016-07-13
Bundle View2.5D extension

qiemem:
Some more removal of `JOGLLoader` related stuff could probably be done, but this does a bunch of it as well.


qiemem:
Fixes #853

PR #1110, opened 2016-07-24
Initial structure for Jump to Declaration

Akarshit:
This is the initial commit for Jump To Declaration for the community to review. Some of the things I know should be changed are
1. Make a utility class that would contain all the function that would be common to Jump to Declaration and Show Usage
2. Efficiently parse the code(like stop parsing for `globals` and `xxxx-own` as soon as first `to` is found)

Any other suggestion would be welcome.


netlogo-gui/src/main/ide/JumpToDeclaration.scala
```diff
@@ -0,0 +1,88 @@
...
+  def jumpToDeclaration(cursorPosition: Int, editorArea: EditorArea): Unit ={
+    val tokenOption = findTokenContainingPosition(editorArea.getText(), cursorPosition)
+    for(token <- tokenOption) {
+      val t = getDeclaration(token, editorArea.getText)
+      editorArea.select(t.start, t.end)
+    }
+  }
+
+  def getDeclaration(token: Token, source: String): Token = {
+    val iterator = Femto.scalaSingleton[TokenizerInterface]("org.nlogo.lex.Tokenizer").tokenizeString(source).toSeq
```
mrerrormessage: I would prefer to see a different name chosen for this than `iterator`. Iterators in scala can't be used after you call a method on them (other than `next()` and `hasNext`). The type of `iterator` is `Seq[Token]` and `Seq`s can obviously have arbitrarily many methods called on them.

netlogo-gui/src/main/ide/JumpToDeclaration.scala
```diff
@@ -0,0 +1,88 @@
...
+    val (globals, owns) = parseOwns(source)
```
mrerrormessage: Tokenizing doesn't take a long time, but since you've created a `Seq[Token]` on line 25 I would prefer to see `parseOwns` take that `Seq` as opposed to taking the source and retokenizing.

netlogo-gui/src/main/ide/JumpToDeclaration.scala
```diff
@@ -0,0 +1,88 @@
+// (C) Uri Wilensky. https://github.com/NetLogo/NetLogo
+
+package org.nlogo.ide
+
+import org.nlogo.core.{Femto, Token, TokenType, TokenizerInterface}
+import org.nlogo.editor.EditorArea
+
+
+object JumpToDeclaration {
```
mrerrormessage: This object needs its own test. I would like to see `jumpToDeclaration` and `findTokenContainingPosition` tested. I know we haven't worked a lot on testing this summer, so please IM me or comment here if you have questions about what sorts of things these tests need to cover.

netlogo-gui/src/main/ide/JumpToDeclaration.scala
```diff
@@ -0,0 +1,88 @@
...
+            argIndex += 1
+          }
+        }
+        globals.get(token.text.toLowerCase).foreach(t => return t)
+        owns.get(token.text.toLowerCase).foreach(t => return t)
+      }
+    }
+    globals.get(token.text.toLowerCase).foreach(t => return t)
+    owns.get(token.text.toLowerCase).foreach(t => return t)
+    null
```
mrerrormessage: It's better scala to return `Some[Token] / None` instead of `token / null`. If the calling code needs `null`, it can always call `.orNull` on the returned `Option`.

netlogo-gui/src/main/ide/JumpToDeclaration.scala
```diff
@@ -0,0 +1,88 @@
...
+            }
+            argIndex += 1
+          }
+        }
+        globals.get(token.text.toLowerCase).foreach(t => return t)
+        owns.get(token.text.toLowerCase).foreach(t => return t)
+      }
+    }
+    globals.get(token.text.toLowerCase).foreach(t => return t)
+    owns.get(token.text.toLowerCase).foreach(t => return t)
```
mrerrormessage: Typically in compiler code we put identifiers in upper case. I would prefer to see `globals` and `owns` also indexed by upper-case strings.

netlogo-gui/src/main/ide/JumpToDeclaration.scala
```diff
@@ -0,0 +1,88 @@
...
+    null
+  }
+
+  def parseOwns(source: String) = {
```
mrerrormessage: Please add a type signature to this method. As a general rule, all scala methods should have type signatures if they're more than one line.

netlogo-gui/src/main/ide/JumpToDeclaration.scala
```diff
@@ -0,0 +1,88 @@
...
+      if (token.text.toLowerCase.endsWith("-own")) {
+        token = iterator.next()
+        while (!token.tpe.equals(TokenType.CloseBracket)) {
+          if (token.tpe.equals(TokenType.Ident) && owns.get(token.text).isEmpty) {
+            owns += token.text -> token
+          }
+          token = iterator.next()
+        }
+      }
+    }
```
mrerrormessage: This code looks like it would be quite efficient and it will work on basically all compiling NetLogo code. However, it isn't hard to devise some non-compiling NetLogo code which will cause it to fail with an exception: `NoSuchElementError`. This needs to either drop the `iterator`/`iterator.next()` calls everywhere and go with a some sort of list comprehension OR check `iterator.hasNext` before *every* call to `iterator.next()`.

netlogo-gui/src/main/ide/JumpToDeclaration.scala
```diff
@@ -0,0 +1,88 @@
...
+    var stack = 0
+    for(i <- indexOfToken to 0 by -1) {
+      val tok = iterator(i)
+      if (tok.tpe.equals(TokenType.CloseBracket)) {
+        stack -= 1
+      }
+      if (tok.tpe.equals(TokenType.OpenBracket)) {
+        stack += 1
+      }
+      if(stack == 0 && tok.text.equalsIgnoreCase(token.text) && iterator(i - 1).text.equalsIgnoreCase("let")) {
```
mrerrormessage: I think this may miss cases where the `let` is defined a level higher than the token in question, but I'm not sure. That would be a good test to incorporate into the unit tests.

netlogo-gui/src/main/ide/JumpToDeclarationAction.scala
```diff
@@ -0,0 +1,18 @@
...
+
+package org.nlogo.ide
+
+import java.awt.event.{ActionEvent, MouseEvent}
+import java.awt.Point
+import javax.swing.{AbstractAction, Action}
+
+import org.nlogo.editor.EditorArea
+
+class JumpToDeclarationAction() extends AbstractAction {
```
mrerrormessage: I really like these small actions. They let you keep the other classes free of Swing logic while still telling Swing how to present the action to the user.

netlogo-gui/src/main/ide/JumpToDeclarationAction.scala
```diff
@@ -0,0 +1,18 @@
...
+  putValue(Action.NAME, "Jump to Declaration")
```
mrerrormessage: This string should use `I18N`.

mrerrormessage:
This looks like it's off to a good start. I had a lot of comments on specific parts of the code, but the general approach seems great.

netlogo-gui/src/main/ide/JumpToDeclaration.scala
Akarshit: Do you want the variable names to be capitalized?

netlogo-gui/src/main/ide/JumpToDeclaration.scala
Akarshit: I have tested this on the following model 
```
globals [
  scared
]

me-own[
  something
]
you-own [
  something
]

to newF
  scared
  let scared
  scared
END
to new [scared]
  scared
  [
    let scared
    scared
    [
      scared
      let scared
      scared
      something
    ]
  ]
  scared
end
```
For this model it works everywhere.

netlogo-gui/src/main/ide/JumpToDeclaration.scala
mrerrormessage: Yes, I think `globals` should be a list of capitalized strings.

netlogo-gui/src/main/ide/JumpToDeclaration.scala
Akarshit: Thanks for this.

netlogo-gui/project/autogen/i18n/GUI_Strings_en.txt
```diff
@@ -165,6 +165,7 @@ tabs.info.help = Help
 
 tabs.code = Code
 tabs.code.rightclick.quickhelp = Quick Help
+tabs.code.rightclick.jumptodeclaration = Jump To Declaration
```
mrerrormessage: I would prefer this a line higher or a line lower so it doesn't break up `tabs.code.rightclick.quickhelp` items.

netlogo-gui/src/main/ide/JumpToDeclaration.scala
```diff
@@ -0,0 +1,91 @@
...
+            argIndex += 1
+          }
+        }
+        globals.get(token.text.toUpperCase) orElse owns.get(token.text.toUpperCase)
+      }
+    }
+    globals.get(token.text.toUpperCase) orElse owns.get(token.text.toUpperCase)
+  }
+
+  def parseOwns(tokens: Seq[Token]): (scala.collection.mutable.Map[String, Token], scala.collection.mutable.Map[String, Token]) = {
```
mrerrormessage: Glad to see a type signature here, but I would prefer if the type was `(Map[String, Token], Map[String, Token])`. The code that uses the result of this shouldn't care whether or not the maps that are returned are mutable or immutable.

netlogo-gui/src/main/ide/JumpToDeclaration.scala
```diff
@@ -0,0 +1,91 @@
...
+    val globals = scala.collection.mutable.Map[String, Token]()
```
mrerrormessage: Please import scala standard library classes at the top of the file. When you need to import a class with the same name as another class, use renaming imports like `import scala.collection.mutable.{ Map => MutableMap }`.

netlogo-gui/src/main/ide/JumpToDeclaration.scala
```diff
@@ -0,0 +1,91 @@
...
+    val owns = scala.collection.mutable.Map[String, Token]()
+    var i = 0
+    while(i < tokens.length) {
+      var token = tokens(i)
+      if (token.text.equalsIgnoreCase("GLOBALS")) {
+        i += 1
+        token = tokens(i)
+        while (!token.tpe.equals(TokenType.CloseBracket) && i < tokens.length) {
```
mrerrormessage: Use `==` instead of `.equals`.

netlogo-gui/src/main/ide/JumpToDeclaration.scala
```diff
@@ -0,0 +1,91 @@
...
+          if (token.tpe.equals(TokenType.Ident) && owns.get(token.text).isEmpty) {
+            owns += token.text.toUpperCase -> token
+          }
+          token = tokens(i)
+          i += 1
+        }
+      }
+      i += 1
+    }
```
mrerrormessage: I see you have stopped using `iterator` here, which is good. There are still cases where this can throw an exception. For instance, if I have a code tab with just `globals` (or `globals [` or `globals [ foo `) and I right-click and choose "jump to definition", I see an `OutOfBoundsException`.

Anytime you're indexing into a sequence, make sure the sequence is defined at that point. Part of the reason that list-comprehensions are nice is that they check for element existence for you. What you're doing here could be rewritten as a `foldLeft`, although it would be quite complex. In this case it might be enough to just check before each index into the list that index is less than the length of the list.

netlogo-gui/src/main/ide/JumpToDeclaration.scala
```diff
@@ -0,0 +1,91 @@
...
+    var stack = 0
+    for(i <- indexOfToken to 0 by -1) {
+      val tok = tokens(i)
+      if (tok.tpe.equals(TokenType.CloseBracket)) {
+        stack -= 1
+      }
+      if (tok.tpe.equals(TokenType.OpenBracket)) {
+        stack += 1
+      }
+      if(stack == 0 && tok.text.equalsIgnoreCase(token.text) && tokens(i - 1).text.equalsIgnoreCase("let")) {
```
mrerrormessage: You gave an example yesterday where this worked (and it does work in many cases). In the following model, it works for the first use of `bar`, but not the second:

```NetLogo
to foo
  let bar 2
  show bar + 3
  ask turtles [
    fd bar
  ]
end
```

netlogo-gui/src/main/ide/JumpToDeclaration.scala
```diff
@@ -0,0 +1,91 @@
...
+        return Some(tok)
+      }
+      if(tok.text.equalsIgnoreCase("to") || tok.text.equalsIgnoreCase("to-report")) {
+        // i + 1 is the function name
+        if(tokens(i + 2).tpe.equals(TokenType.OpenBracket)){
```
mrerrormessage: This errors for the following code when I right-click on the open bracket and choose "Jump to Declaration":

```NetLogo
to foo [
```

mrerrormessage:
I'm happy to see my earlier comments addressed. I left a number of test cases where I'm seeing errors or other problems. It's OK if this doesn't work in every case (it *can't* work in many cases, due to the nature of the problem). I don't want this to display an exception to the user unless there is a truly exceptional event, like out of memory.

I really want to see tests that cover this behavior, both to prevent errors as well as to document what cases this covers and what cases it doesn't cover (yet).

netlogo-gui/src/main/ide/JumpToDeclaration.scala
Akarshit: Any specific reason for this?

Akarshit:
I will try to cover these cases. Also I will handle the exception in the code itself, instead of throwing them.
I will also make a test for this. The example which I gave earlier along with the cases you mentioned should be enough for this. Right?
Also what is the exact syntax for a function.
```
to foo1 [arg1 arg2]
  // the function starts
end
```

```
to foo2 [arg1 arg2] [
  // Is this valid?
] end
```

```
to foo3 [
  // Function starts
] end
```

netlogo-gui/src/main/ide/JumpToDeclaration.scala
mrerrormessage: You might already know this: in scala `==` is the same thing as `equals`. The scala equivalent of java `==` (reference equality) is `eq`. The scala equivalent of java `!=` (reference inequality) is `ne`. Idiomatic scala uses `==` and `!=` basically everywhere java would use `.equals(...)` and `! .equals(...)`.

mrerrormessage:
I'm not sure whether the list of cases, plus your case, covers everything or not. Of the function syntaxes you gave, only the first is a legal function. Procedure syntax is:

```
procedure := procedure-decl ident args? body "END"
procedure-decl := "TO" | "TO-REPORT"
args := "[" ident * "]"
body := ident*
ident := standard NetLogo identifiers, see TokenLexer for a complete set of rules
```

>  Also I will handle the exception in the code itself, instead of throwing them.

We should prevent creating exceptions in the first place. I'm not sure whether that's what you meant by "handle the exception," but we must avoid creating new exceptions. In the event an exception is thrown (because something completely unexpected happens) we *want* that exception to propagate up to the user, so we should avoid catching exceptions.

netlogo-gui/src/main/ide/JumpToDeclaration.scala
```diff
@@ -15,6 +18,9 @@ object JumpToDeclaration {
   def jumpToDeclaration(cursorPosition: Int, editorArea: EditorArea): Unit = {
     val tokenOption = findTokenContainingPosition(editorArea.getText(), cursorPosition)
     for(token <- tokenOption) {
+      if(token.tpe == TokenType.Ident || DefaultTokenMapper.allCommandNames.contains(token.text)
```
mrerrormessage: This should be `token.tpe != TokenType.Ident`

mrerrormessage:
Code structure looks good.

mrerrormessage:
OK, glad to see those tests added! Here's what I would like to see to finish up this pull request:

1. Tests added for each of the failing scenarios I've outlined in the comments. The valid code I've provided is in the tests already, but not the invalid cases. Those should be tested. It's OK to test that jump to definition doesn't return anything in those cases.
2. The test code should be cleaned up a little bit. There's a lot of boilerplate in each test case and much (if not all) of that could get extracted to a helper function.

Akarshit:
OK I will add those test. Also now the Jump to Declaration works for all those cases.

netlogo-gui/src/test/ide/JumpToDeclarationTests.scala
```diff
@@ -0,0 +1,97 @@
...
+      |]
+      |set scared 5
+      |end
+    """.stripMargin
+
+  def createToken(text: String, tokenType: TokenType, startPos: Int) = new Token(text, tokenType, null)(startPos, 0, null)
+
+  def testTokenPosition(tokenToFindDeclaration: Token, startIndexOfDeclaration: Int): Unit = {
+      val tokenOption = JumpToDeclaration.getDeclaration(tokenToFindDeclaration, source1)
+      assertResult(true)(tokenOption.isDefined)
```
mrerrormessage: This is OK as is, but you could also write it as `assert(tokenOption.isDefined)`.

netlogo-gui/src/test/ide/JumpToDeclarationTests.scala
```diff
@@ -0,0 +1,97 @@
...
+    |end
+    """.stripMargin
+
+  val source2 =
+    """
+      |to foo
+      |[
+      |  let scared
+      |  set scared 0
+      |]
```
mrerrormessage: Brackets immediately after a procedure name indicate arguments. This may give you the result you want in your tests, but it isn't representative code. `ask [ let scared ... ]` is correct code and would test the appropriate scoping rules.

mrerrormessage:
This looks really good! Thanks for the hard work on it. I want to see a few more changes before we're ready to merge:
* Tests should cover invalid syntax (including badly invalid syntax). The examples I gave earlier (`globals`, `globals [`, `globals [foo`) are good ones to test. The tests can verify that no definition is found for any token.
* The token stream should filter out any comments before iterating through tokens. Comments can come at any point in the stream and are identified by `token.tpe == TokenType.Comment`.

netlogo-gui/src/test/ide/JumpToDeclarationTests.scala
Idloj: `ask turtles [ let scared ... ]`

Akarshit: OK will change this.

netlogo-gui/src/test/ide/JumpToDeclarationTests.scala
Akarshit: Thanks for the tip.

mrerrormessage:
Looking good. Can you squash this down to one or two commits and get it so there are no merge conflicts with hexy? You can merge hexy into this or rebase hexy into it, either is fine.

Akarshit:
OK will rebase this.

mrerrormessage:
This still looks ready to merge, but I'm seeing a merge conflict in `netlogo-gui/src/main/editor/EditorField.scala`. Make sure you've updated it against the latest `hexy` from `NetLogo/NetLogo` (you may have an outdated copy of `hexy` in `Akarshit/NetLogo`). This will be merged as soon as merge conflicts are resolved.

Idloj:
I understand that jumping to procedure definitions will be on another PR?

PR #1125, opened 2016-08-11
6.0-M9: Clash of local variable definitions

mrerrormessage:
```
to test
  clear-all
  if true [
    let x "if block"
    print x
  ]
  ask one-of patches [
    let x "ask block"
    print x
  ]
end
```

And then:
```
observer> test
if block
if block
```

If you invert the order of the two blocks (i.e., put `ask` before `if`), you correctly get:

```
observer> test
ask block
if block
```

I came upon this while trying to isolate the cause of a `java.lang.NullPointerException` raised by a user model. I strongly suspect that both errors have the same cause, but I haven't (yet) been able to recreate the NPE in a simple model. (I can provide a private link to the user model if needed.)

### DIFF
--- a/netlogo-gui/src/main/compiler/LocalsVisitor.scala
+++ b/netlogo-gui/src/main/compiler/LocalsVisitor.scala
@@ -3,8 +3,8 @@
 package org.nlogo.compiler
 
 import CompilerExceptionThrowers._
-import org.nlogo.core.I18N
-import org.nlogo.nvm.Procedure
+import org.nlogo.core.{ I18N, Let }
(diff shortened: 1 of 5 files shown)


nicolaspayette:
:+1: : I confirm that this fixes both the problem in my minimal example above and the NPE in the aforementioned user model.

mrerrormessage:
Thanks for reporting! This appears only in a very limited number of cases, but it obviously has a major impact on cases where it *does* appear.

SethTisue:
nice catch!

Issue #1046, opened 2016-05-18
ok button should be selected in user-input dialog


qiemem:
so that you can press enter when finished entering input

![image](https://cloud.githubusercontent.com/assets/28215/15371383/c6b575cc-1d00-11e6-9796-66c54bfd2131.png)

H-Theking:
@qiemem How do I get to this dialogue?

SethTisue:
with the `user-input` primitive

H-Theking:
I did a minor edit and it turns out that all `UserDialog`s will be affected. Is it the required behaviour?: 
![userinput](https://cloud.githubusercontent.com/assets/6599060/17643410/462e5b26-6161-11e6-8b8c-a196887c466e.png)
![useroneof](https://cloud.githubusercontent.com/assets/6599060/17643411/588b38ac-6161-11e6-9a41-d7b96db48a84.png)
![yesorno](https://cloud.githubusercontent.com/assets/6599060/17643412/5cef6aa8-6161-11e6-9ba2-8295f2794401.png)

qiemem:
That looks pretty ideal to me.

mrerrormessage:
Closed by #1127

PR #1127, opened 2016-08-13
Select confirm button in UserDialog

H-Theking:
Fix for issue - ok button should be selected in user-input dialog #1046


mrerrormessage:
Thanks for submitting. I'll review this tomorrow.

mrerrormessage:
Looks great, thanks for the fix!

PR #1128, opened 2016-08-14
Added the fca => file-close-all and removed editDistance

Akarshit:


mrerrormessage:
The suggestions here look great, I'm seeing an old problem reappear where when I type a word and then backspace out of that word, the suggestion box disappears (correct). However, when I start typing a new word, the suggestion box reappears, which I don't think is desirable. Does this PR cause that behavior in some way?

Akarshit:
I have updated the PR to fix this issue.

mrerrormessage:
This needs to be rebased / merged with hexy to eliminate merge conflicts. After that it should be good to merge.

Akarshit:
@mrerrormessage Is there something else that has to be done here ?

mrerrormessage:
No, I think this is good. We're still in the process of building the beta so I'm trying to keep changes to a minimum on hexy right now. I know this was opened a few weeks ago, and I probably should have merged it in before, but it will go in after the beta is built and released. Thanks!

PR #1133, opened 2016-08-15
Added highlighting for the current line

Akarshit:


netlogo-gui/src/main/editor/EditorArea.scala
```diff
@@ -470,4 +477,48 @@ class EditorArea(
...
+              p1: Int,
+              bounds: Shape,
+              c: JTextComponent) {
+      try {
+        val r = c.modelToView(c.getCaretPosition)
+        g.setColor(color)
+        g.fillRect(0, r.y, c.getWidth, r.height)
+        if (lastView == null) lastView = r
+      } catch {
+        case ble: BadLocationException => println(ble)
```
mrerrormessage: Catching exceptions is good, but let's print this to `stderr` instead of `stdout`. Also, let's provide a little more context.

mrerrormessage:
This looks good overall. Do you think we could get all of the functionality for this into a single file? It seems spread out across `EditorArea` and `HighlightView`. I haven't looked at this closely enough to have any suggestions on how to do this, but let me know if you have any ideas.

PR #1140, opened 2016-08-29
Bundle LevelSpace

qiemem:


mrerrormessage:
So, the test failing is `org.nlogo.headless.TestCompileAll` due to some of the models in the LevelSpace extension not having the most up-to-date NetLogo version. Can you look into that?

PR #1143, opened 2016-08-30
fix bin/shell.sh for new build structure

SethTisue:
running through sbt like this means you see some sbt noise
at startup now. but meh, it's just us developers using it

### DIFF
--- a/bin/shell.sh
+++ b/bin/shell.sh
@@ -1,11 +1,4 @@
 #!/bin/sh
 
-ROOTDIR="$(dirname -- "$0")/.."
-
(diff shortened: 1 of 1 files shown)


SethTisue:
I see that `headless.Shell` exists in both the netlogo-gui and netlogo-headless projects. I wasn't sure which one is the right one to use.

mrerrormessage:
Thanks for fixing this!

mrerrormessage:
Thanks for fixing this!

mrerrormessage:
Thanks for fixing this!

PR #1146, opened 2016-09-01
Update LS docs

qiemem:


mrerrormessage:
This branch is failing `org.nlogo.headless.TestCompileAll` on Travis. Can you investigate why that is?

bainco:
Looks like it's failing mostly because of the prim change `ls:load-gui-model` and `ls:load-headless-model`.

Some of those could be fixed by updating the NetLogo/models subcommit. The others are from the test models in the NetLogo/levelspace (same problem). 

You could probably use this commit to fix the models library: 1663b1ddeb15f8071789aed33537a3c91eb63eea

That's from a PR (NetLogo/models#295).

PR #1152, opened 2016-09-02
Use normal-sized bullets instead of JEditorPane's bullets. Close #236

### LINKED ISSUE #236: Info tab aesthetics: unordered list bullets are big and ugly
they're like cannonballs. can they be made smaller somehow?

screen shot showing their vastness and darkness:
screen shot

Idloj:
We use a custom image to serve as the bullet.


netlogo-gui/src/main/app/infotab/InfoFormatter.scala
```diff
@@ -44,12 +44,12 @@ object InfoFormatter {
...
 
-  def wrapHtml(body: HTML, fontSize: Int = defaultFontSize): HTML = {
-    "<html><head>"+styleSheet(fontSize)+"</head><body>"+body+"</body></html>"
-  }
+  def wrapHtml(body: HTML, fontSize: Int = defaultFontSize): HTML =
+    "<html><head>"+styleSheet(fontSize)+"</head><body>"+
+      body.replaceAll("<li>", "<li>&bull; ")+
+      "</body></html>"
 
-  def apply(content: String, fontSize: Int = defaultFontSize, attachModelDir: String => String = identity) = {
```
Idloj: This parameter doesn't seem to be used, so I removed it. Have I missed something?

nicolaspayette:
> (a) use an alternative to JEditorPane

I would support the idea. JEditorPane is very limited in its support of HTML/CSS. If we wanted to generally improve the look of the info tab, having a modern HTML component would help a lot.

netlogo-gui/src/main/app/infotab/InfoFormatter.scala
mrerrormessage: Nope, thanks for removing!

netlogo-gui/src/main/app/infotab/InfoFormatter.scala
```diff
@@ -44,12 +44,12 @@ object InfoFormatter {
                          MaxParsingTimeMillis)
       .markdownToHtml(str)
 
-  def wrapHtml(body: HTML, fontSize: Int = defaultFontSize): HTML = {
-    "<html><head>"+styleSheet(fontSize)+"</head><body>"+body+"</body></html>"
-  }
+  def wrapHtml(body: HTML, fontSize: Int = defaultFontSize): HTML =
+    "<html><head>"+styleSheet(fontSize)+"</head><body>"+
+      body.replaceAll("<li>", "<li>&bull; ")+
```
mrerrormessage: Did you check whether the generated HTML contained corresponding `</li>` tags? If it does, I would also like those removed so that we have valid HTML both before and after this transformation.

Idloj: I'm not sure I understand. I believe the markdown parser generates `</li>`, but I didn't check. Why removing the closing tag would make the HTML valid, isn't the closing tag supposed to be there?

My only check was to edit the Info tab and see that a two items long list looks right, no HTML checking.

mrerrormessage: Oh, you're totally right! I missed that you were just replacing `<li>` with `<li>&bull;`, totally my mistake! This looks good. I plan to merge it in shortly.

mrerrormessage:
I do like the look of these bullets more, but they appear even in ordered lists. From the "Info Tab Example" in the models library.

screen shot 2016-09-06 at 4 19 05 pm

Idloj:
I fixed the ordered-lists problem, but I still can't get paragraphs right, they get shown on a separate line than the bullet

Idloj:
@mrerrormessage everything's fixed, I think it's ready for merge

mrerrormessage:
These bullets looked a bit "blurry" to me. Was that intentional on your part? I'm not sure I have a strong preference for hard-edged bullets, but they looked a bit more like what I'm used to seeing. It also made me wonder whether the "blurriness" was some sort of compression artifact. At any rate, the code here looks fine, at this point we're in to questions of styling.

For comparison, original:

screen shot 2016-09-12 at 10 22 08 am

Your changes:
screen shot 2016-09-12 at 10 20 24 am

Bullets with hard edges (I edited the bullet.png in GIMP):
screen shot 2016-09-12 at 10 19 21 am

Idloj:
It wasn't intentional in the first place, but when I saw how it turned up I didn't think it's so bad, though it certainly different than the usual so I guess we should go for the hard-edged.

You want to give me your bullet.png for me to replace with?

mrerrormessage:
Find here: bullet

Idloj:
Updated

PR #1164, opened 2016-09-12
Alternative order for running combinations of behaviourspace settings(fix #278)

### LINKED ISSUE #278: BehaviorSpace: change order of runs
Suggested by Bruce Edmonds:

> May I suggest the following:

> If there are N possible settings for independent runs of a simulation (e.g. A, B, C, D), each done a number of times, that the simulation (under BehaviorSpace control) does them (roughly) in this order:

> (1) A, B, C, D, A, B, C, D, A... etc.

> Rather than:

> (2) A, A, A, A, .... B, B, B, .... C, C, C, .... D, ...etc.

> In this way if something happens (the power goes off, you run out of time before a paper dealine etc.) then you have something nearer an even sample of the parameter space. 

> This would be especially useful if one is trying to generate an indefinite number of runs with different settings.  I know one can do this using behaviorspace-run-number, mod and some NetLogo coding, but the above method would be much more convenient.

> It would not matter if scheme (1) happened within each thread doing simulations, it would still be better than (2).

H-Theking:
An alternative info text for the new checkbox is welcome
Tried to fix some tests. Others still failing though (mostly extension related)

### DIFF
--- a/netlogo-core/src/main/api/LabProtocol.scala
+++ b/netlogo-core/src/main/api/LabProtocol.scala
@@ -7,6 +7,7 @@ case class LabProtocol(name: String,
                     goCommands: String,
                     finalCommands: String,
                     repetitions: Int,
+                    sequentialRunOrder: Boolean,
                     runMetricsEveryStep: Boolean,
                     timeLimit: Int,
                     exitCondition: String,
@@ -28,9 +29,14 @@ case class LabProtocol(name: String,
         case set::sets =>
           set.iterator.flatMap(v =>
             combinations(sets).map(m =>
-              (set.variableName,v) :: m))
+              if(sequentialRunOrder) (set.variableName,v) :: m
+              else m :+ set.variableName -> v))
       }
-    combinations(valueSets)
+    if(sequentialRunOrder) combinations(valueSets)
       .flatMap(Iterator.fill(repetitions)(_))
+    else {
+      val runners = combinations(valueSets.reverse).toList
+      (for(i <- 1 to repetitions) yield runners).flatten.toIterator
+    }
   }
 }
--- a/netlogo-core/src/main/fileformat/LabSaver.scala
+++ b/netlogo-core/src/main/fileformat/LabSaver.scala
@@ -39,7 +39,8 @@ object LabSaver {
   }
   def attributes(specs: (String,String)*) = {
     val result = new AttributesImpl
-    for((name,value) <- specs)
+    for((name,value) <- specs if name != "sequentialRunOrder" || 
+        (name == "sequentialRunOrder" && value == "false"))
       result.addAttribute("", "", name, "CDATA", value)
     result
   }
@@ -56,6 +57,7 @@ object LabSaver {
     hd.startElement("", "", "experiment",
                     attributes(("name", protocol.name),
                                ("repetitions", protocol.repetitions.toString),
+                               ("sequentialRunOrder", protocol.sequentialRunOrder.toString),
                                ("runMetricsEveryStep", protocol.runMetricsEveryStep.toString)))
     if(protocol.setupCommands.trim != "")
       element("setup", protocol.setupCommands)
--- a/netlogo-gui/resources/system/behaviorspace.dtd
+++ b/netlogo-gui/resources/system/behaviorspace.dtd
@@ -11,6 +11,7 @@ section of a NetLogo model file.
 <!ATTLIST experiment
   name CDATA ""
   repetitions CDATA "1"
+  sequentialRunOrder (true|false) "true"
   runMetricsEveryStep (true|false) "true">
 
 <!ELEMENT setup (#PCDATA)>
--- a/netlogo-core/src/main/fileformat/LabLoader.scala
+++ b/netlogo-core/src/main/fileformat/LabLoader.scala
@@ -84,6 +84,9 @@ class LabLoader(literalParser: LiteralParser) {
       readOptional("go"),
       readOptional("final"),
       element.getAttribute("repetitions").toInt,
+      { val defaultOrder = element.getAttribute("sequentialRunOrder").toString
+        if(defaultOrder == "") true else defaultOrder == "true"  
+      },
       element.getAttribute("runMetricsEveryStep") == "true",
       if(!exists("timeLimit")) 0 else readOneAttribute("timeLimit","steps").toInt,
       if(!exists("exitCondition")) "" else readOptional("exitCondition"),
--- a/netlogo-core/src/main/fileformat/NLogoLabFormat.scala
+++ b/netlogo-core/src/main/fileformat/NLogoLabFormat.scala
@@ -41,7 +41,7 @@ trait LabFormat[A <: ModelFormat[Array[String], A]]
       converter.convertStatement(setupCommands),
       converter.convertStatement(goCommands),
       converter.convertStatement(finalCommands),
-      repetitions, runMetricsEveryStep, timeLimit,
+      repetitions, sequentialRunOrder, runMetricsEveryStep, timeLimit,
       if (exitCondition == "") "" else converter.convertReporterExpression(exitCondition),
       metrics.map(converter.convertReporterExpression),
       valueSets)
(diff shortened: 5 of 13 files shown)


H-Theking:
```scala 
I18N.gui("repetitions.info")+"</html>"),
         Property("useDefaultRunOrder", Property.Boolean, "Use default run order",
                  "<html>"+"Example, having variable [\"var\" 1 2 7 8] with 2 repetitions:<br>" + 
                    "default run order: [\"var\" 1], [\"var\" 1], [\"var\" 2], [\"var\" 2]...<br>" +
                    "batched run: [\"var\" 1], [\"var\" 2], [\"var\" 7], [\"var\" 8], [\"var\" 1]..."+"</html>"),
Property("metrics", Property.ReporterOrEmpty,
```
@mrerrormessage Is this text ok or does anyone have a shorter, more meaningful text?

How do I transfer it to the GUI_String file? I've tried adding it manually and using I18N to get the key but I keep having a bad key error or something like that.

mrerrormessage:
First, tremendous work on this! This PR looks like it's getting quite close to review.

>  Is this text ok or does anyone have a shorter, more meaningful text?

The text is:
```
Example, having variable ["var" 1 2 7 8] with 2 repetitions:
default run order: ["var" 1], ["var" 1], ["var" 2], ["var" 2]...
batched run: ["var" 1], ["var" 2], ["var" 7], ["var" 8], [\"var\" 1]...
```

I would prefer the names "alternating" and "sequential" to "default" and "batch". If you have other suggestions, I'm open to them. We can add an explanatory comment in the manual letting users know that "sequential" is the historical default.

I think it would be clearer (although not much shorter) to have it as:
```
For a protocol with 2 repetitions and the variable ["var" 1 2 3],
with respect to "var", a sequential protocol will run experiments ordered 1, 1, 2, 2, 3, 3
while an alternating protocol will run them ordered 1, 2, 3, 1, 2, 3.
```
What do you think?

> How do I transfer it to the GUI_String file? I've tried adding it manually and using I18N to get the key but I keep having a bad key error or something like that.

The format of the GUI_Strings file is quite picky. [This section](https://github.com/NetLogo/NetLogo/blob/hexy/netlogo-gui/project/autogen/i18n/GUI_Strings_en.txt#L170-L197) of the file has several keys which use `\` to allow continuing a long line on a later line and `\n` to insert a newline within the string itself.

Idloj:
I suggest the following text (a mix of the two):

> For example, having ["var" 1 2 3] with 2 repetitions, the experiments' "var" values will be:
> sequential order: 1, 1, 2, 2, 3, 3
> alternating order: 1, 2, 3, 1, 2, 3

I reckon this is not verbose on one hand (so you can see in a glance what the two options are) and quite clear on the other hand.

It's probably worth noting that the word "protocol" is never used in the New Experiment dialog (unless I missed it, of course).

mrerrormessage:
@Idloj , I like that text. I think it's both clearer and shorter than mine. Thanks for the suggestion!

H-Theking:
@Idloj Short and concise. Thanks

netlogo-core/src/main/api/LabProtocol.scala
```diff
@@ -23,14 +24,29 @@ case class LabProtocol(name: String,
...
+          case set::sets =>
+            set.iterator.flatMap(v =>
+              combinations(sets).map(m =>
+                (set.variableName,v) :: m))
+        }
+      case _ => 
+        sets match {
+          case Nil => Iterator(Nil)
+          case set::sets =>
+            set.iterator.flatMap(v =>
```
mrerrormessage: Notice that line 30-32 and 38-40 are identical. Can you come up with a way to reduce this duplication?

netlogo-core/src/main/api/LabProtocol.scala
```diff
@@ -23,14 +24,29 @@ case class LabProtocol(name: String,
...
+              combinations(sets).map(m =>
+                  m :+ new Tuple2(set.variableName,v)))   
+        }
+    } 
+    if(useDefaultRunOrder) combinations(valueSets)
       .flatMap(Iterator.fill(repetitions)(_))
+    else {
+      val runners = combinations(valueSets.reverse).toList
+      (for(i <- 1 to repetitions) yield runners).flatten.toIterator
+    }
```
mrerrormessage: This function branches on `useDefaultRunOrder` once above (in the inner function) and once here. Perhaps it should be two functions instead: `sequentialCombinations` and `alternatingCombinations`?

netlogo-gui/resources/system/behaviorspace.dtd
```diff
@@ -11,6 +11,7 @@ section of a NetLogo model file.
 <!ATTLIST experiment
   name CDATA ""
   repetitions CDATA "1"
+  useDefaultRunOrder CDATA #IMPLIED
```
mrerrormessage: I'd like to see the property name here changed from `useDefaultRunOrder` to `sequentialRunOrder` or `useSequentialRunOrder`. Although "Default" is clearer right now, it will be quite unclear in 6 months.

netlogo-core/src/main/api/LabProtocol.scala
```diff
@@ -23,14 +24,29 @@ case class LabProtocol(name: String,
...
+              combinations(sets).map(m =>
+                (set.variableName,v) :: m))
+        }
+      case _ => 
+        sets match {
+          case Nil => Iterator(Nil)
+          case set::sets =>
+            set.iterator.flatMap(v =>
+              combinations(sets).map(m =>
+                  m :+ new Tuple2(set.variableName,v)))
```
H-Theking: I used (set.variableName,v) and the compiler complained so I used Tuple2. Nice to  know about the other constructor.

netlogo-core/src/main/api/LabProtocol.scala
```diff
@@ -25,12 +26,17 @@ case class LabProtocol(name: String,
     def combinations(sets: List[ValueSet]): SettingsIterator =
       sets match {
         case Nil => Iterator(Nil)
-        case set::sets =>
+        case set::sets =>
```
mrerrormessage: trailing whitespace

netlogo-core/src/main/fileformat/LabSaver.scala
```diff
@@ -39,7 +39,7 @@ object LabSaver {
   }
   def attributes(specs: (String,String)*) = {
     val result = new AttributesImpl
-    for((name,value) <- specs)
+    for((name,value) <- specs if !name.equals("sequentialRunOrder") && value != true)
```
mrerrormessage: use `==` instead of `.equals`.

Idloj: `!=`

Also, if `value` is a string it's always going be `!= true`

Idloj: maybe the intention was "true" ?

mrerrormessage:
I ran into a problem while testing, probably related to [this line](https://github.com/NetLogo/NetLogo/pull/1164#discussion-diff-79397609R42) where I saved a file with "run in sequential order" unchecked, then when I reloaded the file it was checked again.

netlogo-core/src/main/fileformat/LabSaver.scala
H-Theking: @Idloj Thanks for pointing that out.

H-Theking:
The whole point was not to add a ```sequentialRunOrder``` to a file that didn't have one if the box was still checked. Missed the logic though.

netlogo-gui/resources/system/behaviorspace.dtd
```diff
@@ -11,6 +11,7 @@ section of a NetLogo model file.
 <!ATTLIST experiment
   name CDATA ""
   repetitions CDATA "1"
+  sequentialRunOrder CDATA #IMPLIED
```
mrerrormessage: Sorry I didn't catch this before. Is it possible to define this as `(true|false) "true"` (the same as `runMetricsEveryStep` below)?

mrerrormessage:
Awesome work on this. Thanks so much!

PR #1165, opened 2016-09-13
Syntax-highlight code in Info tab. Closes #383

### LINKED ISSUE #383: info tab: code blocks should have NetLogo syntax-highlighting
would require Pegdown upgrade though (#261)

Idloj:
I'll open a PR on NetLogo/models for the "Info Tab Example" model

This colors both blocks (three back-ticks) and inline code (one back-tick)


Idloj:
I still need to fix InfoFormatterTests

mrerrormessage:
Very excited for all the tests to pass so I can merge this in. Looks *awesome*!
screen shot 2016-09-13 at 4 50 02 pm

netlogo-gui/src/test/app/infotab/InfoFormatterTests.scala
```diff
@@ -87,13 +88,13 @@ class InfoFormatterTests extends FunSuite {
   // here we get a <pre> block because we have a blank line before the fence
   test("fenced code blocks 1") {
     assertResult("""|<p>foo</p>
-              |<pre><code>   bar
+              |<pre><code class="text">   bar
```
Idloj: Is this change acceptable?

netlogo-gui/src/test/app/infotab/InfoFormatterTests.scala
```diff
@@ -104,13 +105,13 @@ class InfoFormatterTests extends FunSuite {
 
   // in contrast, here no <pre> because no blank line
   test("fenced code blocks 2") {
-    assertResult("""|<p>foo<br/><code>
-              |   bar
-              |  baz
-              |qux
-              |</code><br/>oof</p>""".stripMargin) {
+    assertResult("""|<p>foo<br/><code class="text">
```
Idloj: This test fails because, I think the missing newlines confuses pegdown and the block is handled as code (so we get all kind of `font` tags from the conversion to HTML).

But I don't think this should be supported anyway, you're supposed to put a newline before all sorts of things, including code blocks.

Idloj: Thinking about it again, I think I know what actually happens here.

Because it is part of a paragraph (there's no blank line), there is *no special meaning to ```*. The first two backticks contain nothing so they render as nothing, then there is another pair of backticks that contain the "block" (inlined though), and then there is another empty inlined code.

This means that:
 - This actually works correctly because we parse inline code as NetLogo code (that should be colorized)
 - This test should be renamed

netlogo-gui/src/test/app/infotab/InfoFormatterTests.scala
mrerrormessage: Looks good to me.

netlogo-gui/src/test/app/infotab/InfoFormatterTests.scala
mrerrormessage: I agree. It's fine for info tabs to require newlines between sections. The test can (and should) be renamed.

Idloj: What's the purpose of this test, then? Does it check inline code? If so, the test itself would better be changed to be more clear?

I'll also remove the comment since it's misleading.

mrerrormessage: I think the purpose of this test is to clarify that code blocks appearing within a markdown paragraph get rendered with the same `<p>` in HTML as the surrounding text. I don't believe this is important because `<p>` isn't really semantic. We don't name any of the `<p>` so it isn't as though users can link to them. So the question remaining is whether this is a style concern. I've gone through and checked the models library and there are no models (that I could find) which relied on this old behavior (only a few models used backtick-delimited code sections) so that isn't a concern either. If there *are* display problems, we should be fixing them by altering CSS, not by using the `<pre>` tag.

So in summary, yes, let's delete the test!

mrerrormessage:
This is ready for merge. I'm coordinating with @nicolaspayette on whether we need to get this sync-ed up with https://github.com/NetLogo/models/pull/300 in order to avoid breaking the models build.

PR #1182, opened 2016-10-13
Lower precedence of ifelse-value

### LINKED ISSUE #588: `ifelse-value` parses arguments differently than `if` and `ifelse`
`ifelse-value` doesn't deal with infix operator in the first argument correctly.

`ifelse-value 0 < 1 [ "foo" ] [ "bar" ]` should return `"foo"`, but it fails to compile with

> ERROR: IFELSE-VALUE expected 3 inputs, a TRUE/FALSE, a reporter block and a reporter block. 

Note that this is different that `ifelse` and `if`, which handle infix operators as expected.

qiemem:
Fixes #588

### DIFF
--- a/netlogo-gui/src/test/compile/TestAllSyntaxes.scala
+++ b/netlogo-gui/src/test/compile/TestAllSyntaxes.scala
@@ -95,7 +95,7 @@ class TestAllSyntaxes extends FunSuite {
                      |_hubnetmessagetag ,string,OTPL,None,10,0,0
                      |_hubnetmessagewaiting ,TRUE/FALSE,OTPL,None,10,0,0
                      |_hubnetoutqsize ,number,OTPL,None,10,0,0
-                     |_ifelsevalue TRUE/FALSE/reporter block/reporter block,anything,OTPL,None,10,3,3
+                     |_ifelsevalue TRUE/FALSE/reporter block/reporter block,anything,OTPL,None,3,3,3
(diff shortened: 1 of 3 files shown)


qiemem:
Current break is due this procedure in Robby the Robot:

```
to-report state
  let north patch-at 0 1
  let east patch-at 1 0
  let south patch-at 0 -1
  let west patch-at -1 0
  report ifelse-value (is-patch? north) [ifelse-value (any? cans-on north) [81] [0]] [162] +
         ifelse-value (is-patch? east ) [ifelse-value (any? cans-on east ) [27] [0]] [ 54] +
         ifelse-value (is-patch? south) [ifelse-value (any? cans-on south) [ 9] [0]] [ 18] +
         ifelse-value (is-patch? west ) [ifelse-value (any? cans-on west ) [ 3] [0]] [  6] +
                                         ifelse-value (any? cans-here)     [ 1] [0]
end
```

Essentially, code where you have an infix operator between two `ifelse-value`s no longer works. For instance:

```
ifelse-value true [ 5 ] [ 6 ] + ifelse-value false [ 7 ] [ 8 ]
```

There is no way to reconcile this with have the following work:

```
ifelse-value 1 + 2 = 3 [ 5 ] [ 6 ]
```

using just precedence. Personally, I think the having the second form work is more important than having the first form work, but it would be a breaking change.

mrerrormessage:
This didn't get merged for 6.0 as it would have broken existing models in hard-to-fix ways. I'm going to go ahead and close it for now. I would be open to adding a ternary primitive in the future which binds more loosely, but we would probably need to give it a different name.

PR #1187, opened 2016-10-28
Fix link consistency issues

### LINKED ISSUE #1169: Fix link primitive consistency issues
Relatively recently, we have discovered a number of consistency issues with the link primitives:
- #1130 - `my-in-links`/`my-out-links` inconsistent with each other and with docs; unresolved
- #739 - `link-neighbor?` wrong for directed, breeded links; resolved by changing `link-neighbor?` stricter in accordance with docs
- #740 - `*-link-neighbor?` primitives wrong for undirected link breeds; resolved by making `out-link-neighbor?` stricter in accordance with docs
- #639 - `my-links` documentation is wrong; resolved by keeping `my-links` flexible and changing docs

Despite resolution of some specific issues, consistency issues remain. Furthermore, it's not clear how the inconsistencies should be resolved. This is meant to be a mega-issue for all link primitive consistency problems.

Here is a table showing the current behavior of all relevant link primitives:

image

The column headers indicate the type of link or neighbor being tested: "un" is undirected, "out" is directed, outgoing, and "in" is directed, incoming. The "-breeded" variants indicate that a breeded link is being used. The breeded primitives are automatically adjusted to target the breed of the link being tested.

A few inconsistencies:
- `link-with` errors when using directed, non-breeded links, but returns directed, breeded links. This is both internally inconsistent and conflicts with `link-neighbors` and `link-neighbor?`
- `my-out-links`,  is similarly inconsistent between breeded and non-breeded.
- `my-out-links` and `my-in-links` handle undirected links differently.
- `out-link-to` and `in-link-from` handle undirected links differently.
- `my-links` return everything whereas, eg, `link-neighbors` does not.

Breeded variants of primitives are internally consistent with each other, and are almost consistent with unbreeded variants in unbreeded models, if we assume that unbreeded links should act like breeded links with respect to the unbreeded primitives.

Now, before discussing how to resolve these issues, keep in mind a few things about links in NetLogo:
- A model may only use breeded or non-breeded links, not both. (which type a model can use depends on what kind of link is created first, not on the declaration of breeds or anything)
- Non-breeded links may be directed or undirected, but all non-breeded links in a model must have the same directedness.
- Breeded links are declared undirected or directed.
- Mixed directedness is possible, but only through multiple link breeds.

I see a few possible ways of resolving these issues, with two being strict and one being flexible.
## Strict

This is likely what people had in mind when originally designing the link system, as this is the resolution the docs are most consistent with:

image

The idea here is that:
- An unspecified `link` _always_ refers to an undirected link. (the except being `links` itself)
- `out-link` _always_ refers to a directed, outgoing link
- `in-link` _always_ refers to a directed, incoming link

I've removed any erroring behavior whatsoever: sometimes, you don't know the directedness of your links before you start making them (they could be determined by a switch, or imported from a file). Thus, being able to do things like `(link-set my-links my-out-links)` is crucial.
## Super strict

image

This one is even stricter in that, instead of returning `false`, it errors in the situations where a un-breeded primitive can never return anything. This makes it so that unbreeded primitives in unbreeded models behave like breeded primitives in breeded models.

I believe this is what was originally intended for link primitives in NetLogo, as it's the closest match to how they behave currently (with the biggest, perhaps justifiable exception being `my-links`).

I would strongly advise against this method, as it makes models where you don't know directedness beforehand basically impossible, as you can't do stuff like: `(turtle-set link-neighbors out-link-neighbors)`. 
## Flexible

This is a more pragmatic approach, rooted in what I think is more useful given my experiences in network analysis:

image

The idea here is that:
- An unspecified `link` refers to any type of link.
- `out-link` refers to a outgoing link, regardless of directedness; that is, any link that you can use to travel out from node A to node B.
- `in-link` refers to a incoming link, regardless of directedness; that is, any link that you can use to travel into node A from node B.

Practically, the times when this differs from other methods are in mixed-directed networks and in networks where you don't know the directedness beforehand.

I have a strong preference for the flexible approach.
- Generally, in network science and graph theory, undirected links are treated as both incoming and outgoing links. Virtually all graph algorithms treat undirected links as both incoming and outgoing links, to the point where in the most basic network analytics, each undirected link is like two directed links (thus, the `2 * L` in all the undirected forms of network analytics). Case-in-point: In the NW extension, the link and neighbor fetching functions that all the algorithms are based on use the flexible approach.
- For networks where you don't know the directedness beforehand, this allows you to write algorithms that naturally work with either. In the strict resolution, you have to use a lot of `(link-set ...)` and `(turtle-set ...)`.
- In mixed networks, you can achieve the strict results by either just using the breeded primitives themselves (which you would probably do anyway), or, worst case, doing things like `my-out-links with [ is-directed-link? self ]`.

While the flexible approach differs more with NetLogo's current behavior, the only models where results will actually change are multi-link-breed, mixed-directedness models where the un-breeded primitives are being used. Otherwise, it will only make what were error-causing situations into non-error-causing situations.

The strict resolutions, on the other hand, would change the behavior of even unbreeded models, given the differences to `my-links` and the `out-link` primitives.

We could also consider hybrid resolutions, where, say, the unspecified link primitives do any directedness, the `in-link` do specifically directed, incoming links, and the `out-link` primitives take either the flexible or strict approach.

qiemem:
Fixes #1169.

Still working on doc updates, but wanted open this up so you can start reviewing whenever you want, @mrerrormessage. Tests _should_ be passing at this point (we'll no for sure in a few minutes).


netlogo-gui/src/main/prim/threed/_breedat.scala
```diff
@@ -3,39 +3,38 @@
...
     val patch =
       try context.agent.asInstanceOf[Agent3D].getPatchAtOffsets(dx, dy, dz)
       catch {
         case e: AgentException =>
-          return new ArrayAgentSet(AgentKind.Turtle, 0, false)
+          return AgentSet.fromArray(AgentKind.Turtle, Array())
       }
     if (patch == null)
-      new ArrayAgentSet(AgentKind.Turtle, 0, false)
+      AgentSet.fromArray(AgentKind.Turtle, Array())
```
mrerrormessage: Do you think it would make sense to add an `AgentSet.empty` method to name this case explicitly? If `ArrayAgentSet` is immutable, `AgentSet.empty` could even return pre-allocated AgentSets.

netlogo-gui/src/test/headless/TestLanguage.scala
```diff
@@ -180,3 +191,8 @@ TurtleSet_2D
     assert(tests.toString == "List(LanguageTest(test,TurtleSet_2D,List(Command(crt 1,Observer,Success()), Reporter([turtle-set self] of turtle 0 = turtles,Success(true)))))")
   }
 }
+
+class LanguageTestException (line: String, cause: Exception)
+extends Exception("The test errored on the following line: " + line, cause) {
+  setStackTrace(Array())
```
mrerrormessage: Does this suppress the original exception stack trace in the output? Sometimes that contains important information.

netlogo-headless/test/benchdumps/VirusNet.txt
```diff
@@ -1165,70 +1165,43 @@ procedure SETUP-SPATIALLY-CLUSTERED-NETWORK:[NUM-LINKS _repeatlocal:1]{OTPL}:
...
-                GOTO L12
-               L11
-               FRAME SAME
-                ICONST_1
-               L12
-               FRAME SAME1 I
+                INVOKEINTERFACE org/nlogo/agent/LinkManager.linksWith (Lorg/nlogo/agent/Turtle;Lorg/nlogo/agent/Turtle;Lorg/nlogo/agent/AgentSet;)[Lorg/nlogo/agent/Link;
+                CHECKCAST [Ljava/lang/Object;
+                INVOKEVIRTUAL scala/Predef$.refArrayOps ([Ljava/lang/Object;)Lscala/collection/mutable/ArrayOps;
+                INVOKEINTERFACE scala/collection/mutable/ArrayOps.nonEmpty ()Z
```
mrerrormessage: `nonEmpty` is definitely more descriptive, but on a pure-performance level I suspect using `.length > 0` would be slightly faster (field access and compare vs loading object `scala.Predef$` and invoking method). Is performance of `link-neighbor` important in most link-based models? This bytecode is generated by `_linkneighbor.scala` and is generated by the following line of the VirusNet Benchmark:

```
let choice (min-one-of (other turtles with [not link-neighbor? myself]) [distance myself])
```

parser-core/src/main/core/BreedIdentifierHandler.scala
```diff
@@ -36,24 +36,21 @@ object BreedIdentifierHandler {
...
+    LinkPrimitive ( "<BREED>"               , Reporter,  "etc._linkbreedsingular" ),
+    LinkPrimitive ( "<BREED>-NEIGHBOR?"     , Reporter,  "etc._linkneighbor"      ),
+    LinkPrimitive ( "<BREED>-NEIGHBORS"     , Reporter,  "etc._linkneighbors"     ),
+    LinkPrimitive ( "<BREED>-WITH"          , Reporter,  "etc._linkwith"          ),
+    LinkPrimitive ( "MY-IN-<BREEDS>"        , Reporter,  "etc._myinlinks"         ),
+    LinkPrimitive ( "MY-<BREEDS>"           , Reporter,  "etc._mylinks"           ),
+    LinkPrimitive ( "MY-OUT-<BREEDS>"       , Reporter,  "etc._myoutlinks"        ),
+    LinkPrimitive ( "OUT-<BREED>-NEIGHBOR?" , Reporter,  "etc._outlinkneighbor"   ),
+    LinkPrimitive ( "OUT-<BREED>-NEIGHBORS" , Reporter,  "etc._outlinkneighbors"  ),
+    LinkPrimitive ( "OUT-<BREED>-TO"        , Reporter,  "etc._outlinkto"         )
```
mrerrormessage: 👍

netlogo-gui/src/main/agent/ArrayAgentSet.scala
```diff
@@ -0,0 +1,245 @@
...
+      index += 1
+      result
+    }
+  }
+
+  // extended to skip dead agents
+  private class DeadSkippingIterator extends Iterator {
+    // skip initial dead agents
+    while (index < array.size && array(index).id == -1)
+      index += 1
```
qiemem: Since `hasNext` is just `index < array.size`, it expects `index` to already be pointing at the next element. Thus, pre-initialization is needed. The solution would be to make `hasNext` do the work of getting the next agent. In my experience, that's actually the standard way to writer iterators. Performance wise, I would think they'd be identical, but obviously I'd have to measure. I think I will leave it as is for now.

netlogo-gui/src/main/prim/threed/_breedat.scala
qiemem: Yes, absolutely. I kept thinking about doing it, but never got around to it.

netlogo-gui/src/test/headless/TestLanguage.scala
qiemem: Nope, the stack trace of the cause is preserved. I'll add a comment.

netlogo-gui/src/main/agent/ArrayAgentSet.scala
mrerrormessage: Sounds reasonable, this is good then.

mrerrormessage:
Looks good after latest changes. Minor issue, remove curly braces here (https://github.com/NetLogo/NetLogo/pull/1187/commits/64fc55c3d23d54e9953056b2eccb588d26bb6b6f#diff-90319e23b657120d6428b8ac2dddfb76R5).

mrerrormessage:
One major outstanding issue is documentation. I'll merge this as soon as that's done.

qiemem:
I believe the Travis failure is a result of the nondeterminism in the NW extension discussed in NetLogo/NW-Extension#173. In other words, not this codes fault. I've restarted Travis as most of the time the test passes.

PR #1192, opened 2016-11-02
Procedure collapse 

mrerrormessage:
Still needs some tidy-up in the git history, but wanted to open a PR for review


netlogo-gui/src/main/app/FileManager.scala
```diff
@@ -0,0 +1,403 @@
...
+      manager.newModel()
+    }
+  }
+
+  class OpenAction(manager: FileManager, parent: Component)
+    extends ExceptionCatchingAction(I18N.gui.get("menu.file.open") + Ellipsis, parent) {
+    putValue(UserAction.ActionCategoryKey, UserAction.FileCategory)
+    putValue(UserAction.ActionGroupKey,    UserAction.FileOpenGroup)
+    putValue(UserAction.ActionRankKey,     Double.box(2))
+    putValue(ACCELERATOR_KEY,              UserAction.KeyBindings.keystroke('O', withMenu = true))
```
Idloj: Can the `putValue` here and in many other places throughout the files be factored out for convenience and DRYness?

netlogo-gui/src/main/editor/EditorMenu.scala
```diff
@@ -0,0 +1,9 @@
+// (C) Uri Wilensky. https://github.com/NetLogo/NetLogo
+
+package org.nlogo.editor
+
+import javax.swing.Action
+
+trait EditorMenu {
+  def offerAction(action: Action): Unit
```
Idloj: Why is this named `offerAction`, isn't it more of an `addAction`?

Idloj:
I know it's me to blame, but I think `object Preference` would be better as `object Preferences` and probably on it's own file, not with `trait Preference`

netlogo-gui/src/main/editor/EditorMenu.scala
mrerrormessage: In java awt / swing `add` is used on containers to indicate that a component should be added. The `add` method implies two things: (1) that the object will definitely be added after `add` is called and (2) if `add(a)` is called before `add(b)`, `a` will appear "before" `b` within the container. In this case, those implications are false. A menu may or may not add an entry for an action and the order in which the actions are added doesn't influence the order in which they are contained. In light of that, I thought `offer` was a clearer term. Since `offer` doesn't necessarily imply that an action was `add`ed, `revoke` doesn't necessarily imply that it will be removed (that is, it isn't an error to `revoke` an action even if the menu to which it was offered didn't add it during the offer).

netlogo-gui/src/main/app/FileManager.scala
mrerrormessage: I considered that at one point, I added [RichUserAction](https://github.com/NetLogo/NetLogo/pull/1192/files#diff-19e3e2b47ad613d4b4c9077ac7262c03R76) as a step to get rid of the corresponding `getValue` methods. I think adding a trait that allows setting these more naturally would make things more convenient.

mrerrormessage:
You may be right about `Preference`/`Preferences`. The good news is that the amount of code involved is still small and can be easily changed (although I don't plan to change it in this PR).

netlogo-gui/src/main/editor/EditorMenu.scala
Idloj: That makes sense to me, but it makes me ask: "in what situations will a menu decide not to add a menu entry?"

mrerrormessage: Here are a two cases:

* Some component is holding a bunch of actions and doesn't know which one of them are menu entries, so it offers all of them to the menu to figure out what to present
* A menu is offered an action that doesn't correspond to its function (we currently split based on the menu category in `MenuBar`, but we could also check this in the menus themselves).

So the whole point of this massive menu refactoring (which is probably the majority of the code in this PR) is actually to separate different concerns. Menu's shouldn't know or care about *how* to perform the actions they contain; items which have a collection of actions (editor and tabs in this case) shouldn't know or care about *whether* an action will be placed in the menu; actions shouldn't care about *where* in the menu they are presented (and indeed, shouldn't really care whether they're presented at all, since the action doesn't know the state of the whole program and menu actions disappear for reasons unrelated to the action).

So by introducing the possibility that a menu may or may not add an action, we've moved the concern about whether an action is appropriate to present to the user to the two parties which know the most about the action: the menu and the action itself. If the action container knows that a menu item isn't appropriate (as Tabs sometimes does) it can use `revoke` without needing to worry about whether the menu is presenting the action or not.

Idloj: Wow!, that's very smart. I think this approach can be used in a lot of Swing apps.

mrerrormessage: It's definitely possible that it could be. At some point in the future I would like to enhance the system by making keyboard shortcuts configurable (and maybe making what operations appear configurable). For now, it lets us add actions to the menu bar and helps to reduce some of the dependency problems in `App.scala`.

netlogo-gui/src/main/app/FileManager.scala
mrerrormessage: Just updated. Thanks for the suggestion! As I mentioned, I had considered this, but it was good to get some feedback that this would actually be helpful.

Idloj:
Are the internal changes going to be documented somewhere?

netlogo-gui/src/main/app/tools/Preference.scala
```diff
@@ -38,4 +39,20 @@ object Preference {
       prefs.put("user.country", chosenLocale.getCountry)
     }
   }
+
+  class LineNumbers(tabs: TabsInterface) extends Preference {
+    val i18nKey = "editorLineNumbers"
+    val component = new JCheckBox()
+    val restartRequired = true
```
Idloj: Restart is not required, you just have to hit "Apply".

This experience made me think: do we want that "Apply" button or should changes immediately apply as they're changed? (The second option is more appealing to me. In general, you want the feedback to be as immediate as possible)

Idloj:
I noticed that keywords are now bold (unlike before), is that intentional?

Idloj:
I don't see the Import/Export submenus

Idloj:
There's no undo/redo while editing the Info tab

mrerrormessage:
> I don't see the Import/Export submenus

I certainly see them. I just got done with a series of rebases and had some small corrections (just pushed). It's possible that without the small corrections they wouldn't show up. 

> There's no undo/redo while editing the Info tab

Thanks for spotting this.

> I noticed that keywords are now bold (unlike before), is that intentional?

This is intentional.

> Restart is not required, you just have to hit "Apply".

Thanks for pointing this out, I'll adjust.

> Are the changes going to be documented somewhere?

UserAction.scala has a little documentation. What sort of documentation would be helpful to you (or others) in understanding this?

netlogo-gui/src/main/swing/UserAction.scala
```diff
@@ -0,0 +1,148 @@
...
+import javax.swing.{ Action, KeyStroke }
+
+object UserAction {
+  /* Key denoting in which menu an action ought to be included */
+  val ActionCategoryKey    = "org.nlogo.swing.ActionCategoryKey"
+  /* Key for an action to denote what actions it should be grouped with. Allows like actions to be grouped together. */
+  val ActionGroupKey       = "org.nlogo.swing.ActionGroupKey"
+  /* Key for an action to share a submenu with */
+  val ActionSubcategoryKey = "org.nlogo.swing.ActionSubcategoryKey"
+  /* Key for an action to indicate it's rank within the group, expressed as a java.lang.Double.
```
Idloj: Why a Double?

Idloj:
Documentation about the new menu system, both generally and practically, like how to add a new action, how and when to offer it.

Documentation about some other design changes and responsibilities that moved around, e.g., the new FileManager. In general, some info about the new code structure & architecture.

Idloj:
A general question: is it planned to use the new editor area functionality to replace some of the custom written editor-related things (like line-numbers, code-completion, etc.)? (It makes sense to me to do that)

mrerrormessage:
> A general question: is it planned to use the new editor area functionality to replace some of the custom written editor-related things (like line-numbers, code-completion, etc.)? (It makes sense to me to do that)

At the moment, the new editor area functionality is used in the code tab while the old editor is used in editor fields. Over time, I think it would be great to replace basically all of the editor-related functionality, but I was a bit hesitant to do that in this PR because I wanted to see how the changes were received. Over time, it would be great to replace all editor functionality with RSyntaxTextArea.

> Documentation about the new menu system, both generally and practically, like how to add a new action, how and when to offer it.
> Documentation about some other design changes and responsibilities that moved around, e.g., the new FileManager. In general, some info about the new code structure & architecture.

This sounds like great documentation to have. I'll try to put together something on NetLogo internals with a focus on menus and `App` / startup in the next few days, probably while I build the next release

Idloj:
That would be great, thanks

Idloj:
The Import/Export submenus disappear after opening a library model, they do not come back after `File > New`

mrerrormessage:
> The Import/Export submenus disappear after opening a library model, they do not come back after File > New

This has now been fixed, thanks for testing!

mrerrormessage:
> The Import/Export submenus disappear after opening a library model, they do not come back after File > New

This has now been fixed, thanks for testing!

mrerrormessage:
> The Import/Export submenus disappear after opening a library model, they do not come back after File > New

This has now been fixed, thanks for testing!

mrerrormessage:
> The Import/Export submenus disappear after opening a library model, they do not come back after File > New

This has now been fixed, thanks for testing!

mrerrormessage:
> The Import/Export submenus disappear after opening a library model, they do not come back after File > New

This has now been fixed, thanks for testing!

mrerrormessage:
> The Import/Export submenus disappear after opening a library model, they do not come back after File > New

This has now been fixed, thanks for testing!

PR #1194, opened 2016-11-03
Implement range

### LINKED ISSUE #1121: Implement `range`
`range` is a common function in a number of programming languages that is used to generate sequences of numbers. It often has three forms:

```
range 10 => [0 1 2 3 4 5 6 7 8 9]
range 5 10 => [ 5 6 7 8 9 ]
range 5 10 2 => [ 5 7 9 ]
```

This came up today in a devel meeting. Currently, the way to implement `range` functionality is with `n-values`, which is essentially the first form of `range`, followed by a `map`. `n-values` works well for the basic case (with just a maximum: `n-values 10 [ ? ]`) but for more complicated cases, it gets trickier:

```
range start top => n-values (top - start) [ from + ? ]
range start top step => n-values ceiling ((top - start) / step) [ start + step * ? ]
```

To be honest, it took me an embarrassing amount of time to figure out how to correctly write the `n-values` version with `step`. A novice would likely have a very difficult time figuring out how to do this. This is fairly common task and should be simple to do.

As for what the range syntax should look like, the simplest is `range`, with one required argument and two optional. The three forms are then:

```
range top
(range start top)
(range start top step)
```

Note that `step` here can then be a decimal.

The problems with this are:
- Remembering whether `step` comes before or after `top` is always a pain. Different programming languages do this differently.
- The meaning of the arguments changes depending on how many there are.

I'm personally okay with these drawbacks.

This is definitely _not_ targeted for hexy.

qiemem:
Fixes #1121


mrerrormessage:
🤖 meep. morp. This change requires language tests. meep. morp. 🤖

SethTisue:
mo infix operators, mo weird parsing problems (and not just for implementers either, for users too). I liked the prefix version better. infix is only maybe worth it for super common things like `with`

qiemem:
Switched back to prefix form. As nice as the infix syntax read, we decided the weird context dependent `by` and introducing a new data type were too high a cost.

mrerrormessage:
Looks like the following tests are failing on Travis:

* org.nlogo.headless.TestChecksums
* org.nlogo.headless.TestCompileAll

PR #1195, opened 2016-11-07
Bugfix: Only make pen trail updates when turtle moves

TheBizzle:


TheBizzle:
Note: This only affects Headless

PR #1215, opened 2016-11-10
Improve external source files (.nls files) UX

### LINKED ISSUE #710: .nls improvements catchall bug
This ticket refers to multiple different problems that probably ought to be fixed, so it would not be unreasonable to break it up into multiple smaller issues—perhaps one issue for each circle 1-6?

[Here](https://groups.google.com/forum/#!topic/netlogo-devel/AJY9Y3b_-ow), Alan Isaac pointed out a couple of usability problems with NetLogo's support for `.nls` files.  His complaints (and several of mine) can be summarized as follows:
1. The "Includes" dropdown is hidden until an `__includes []` directive (with at least one string inside the list) is entered
2. Even if you have the tab selected for the `.nls` file you wish to save, choosing "Save as" after creating a new `.nls` file will not provide you with the option to save the `.nls` file; "Save as" only ever seems to save the `.nlogo` file
3. Even if you have the tab selected for the `.nls` file you wish to save, choosing "Save" on your fresh, new tab for the `.nls` file forces you to first "Save as" changes to the `.nlogo` file (if you got it from the Models Library)
4. An `.nls` file cannot be saved in NetLogo until something has been entered into the editor
5. Code tabs do not indicate when they have unsaved content
6. Rather than NetLogo being smart when you try to save an `.nls` file without having ".nls" in the path, it instead just complains to you, "You must choose a name ending with: .nls"

I think that the excuse till now has been that the `__includes` stuff is "experimental".  However, I can find references to it dating back to the manual for NetLogo 4.0 (back in the glorious September of 2007).  That's... pretty old.

As far as I'm concerned, the experiment's over—who's still gathering results on this experiment and doing anything with them?—and Alan's report is certainly not the first with significant complaints about the `__includes` support.  I think that it's about time that we either get serious about the functionality or remove it (and I really think that we shouldn't remove it, because I think it solves a serious problem for model authors).  The point is, though, that claiming it to still be experimental does not strike me as reasonable.

Regarding (1), I think the current behavior is pretty bogus.  Alan suggests that `__includes []` (without supplying any file name—real or fake) should show the dropdown.  Personally, I don't see what is really gained by hiding that dropdown at all ever.  It just serves to make things annoying for someone who wants to introduce a first `.nls` file.  It would actually be a very simple fix to [this file](https://github.com/NetLogo/NetLogo/blob/5.x/src/main/org/nlogo/app/IncludesMenu.scala) to remove all of the references to `isEmpty` and make the dropdown always appear.

Regarding (2), surely this can't be the intended behavior....

Regarding (3), this is nothing short of annoying.  "Save" is acting as a "Save all", and we provide no means through which to selectively save an `.nls` file.

Regarding (4), that's phenomenal!  When coupled with (3), it's actually reasonable to think that you're saving the new `.nls` file that you just created, when, in fact, you simply resaved the `.nlogo` file that you had to edit in order to get a chance to open the `.nls` pane.  (I say that it's reasonable to think that because I seriously just did it a few minutes ago when I was tinkering with the functionality.)  It's not even that NetLogo has an opposition to saving an empty `.nls` file, because you can open a new `.nls` pane, type a character, hit backspace, and _then_ save the new file; it's just that you can't save a 100% fresh `.nls` file.

Regarding (5), that's quite user unfriendly.  IntelliJ shows you when a tab contains unsaved changes simply by putting an asterisk at the end of the tab's title.  Even an indication as simple as that is a sufficient fix—just so people know approximately what changed.

Regarding (6), this is kind of baffling.  It knows exactly what the problem is.  Just add the missing extension to the filename!

Idloj:
Also closes #710


Idloj:
Travis fails on the gogo extension, it's a trivial fix for me to do, but how should I do it (from the aspect of it being a submodule)?

mrerrormessage:
Once you've made the change, make an issue on the gogo extension
referencing your change (doesn't need to be a PR) and I'll open a new
branch.

Idloj:
I'm not sure I understand what you mean by "Once you've made the change ... referencing your change", how do I make the change?

It's a one line change, I can even make it on Github itself, should I open a PR on gogo-extension?

mrerrormessage:
Yeah, open a PR for the gogo extension on GitHub targeting the new
includes-enhancement branch.

Idloj:
Done.

What was the decision about the visibility and name of the Includes menu?
Should I add an "always show" checkbox in the Preferences dialog? Should I change the name of the menu?

Idloj:
This PR makes it so that when a file is compiled, it is first saved. This change was based on the behavior before #1198. It needs to be changed now that there is no need to save a file in order to compile the current version of it.

mrerrormessage:
I'm in the process of reviewing this. I hope to finish reviewing and do an initial test on Monday. Thanks so much for the PR!

Idloj:
How do I update the gogo-extension commit that the submodule points to?

mrerrormessage:
The gogo extension can be updated by fetching from NetLogo's remote and checking out the `includes-enhancements` branch, then going back to the NetLogo root and running `git add extensions/gogo` and then `git commit`.

netlogo-gui/src/main/app/Tabs.scala
```diff
@@ -93,132 +105,154 @@ class Tabs(val workspace:    GUIWorkspace,
...
       val index = getIndexOfComponent(tab)
       remove(tab)
       removeMenuItem(index)
+      externalFileTabs -= tab
     }
-  }
 
   def forAllCodeTabs(fn: CodeTab => Unit) =
-    getComponents collect { case tab: CodeTab => tab } foreach (fn)
+    (externalFileTabs.asInstanceOf[Set[CodeTab]] + codeTab) foreach fn
```
Idloj: Yes, otherwise you cannot add `codeTab` (of type `MainCodeTab`) to a `Set[TemporaryCodeTab]`, so I convert it to a `Set[CodeTab]`

netlogo-gui/src/main/app/codetab/TemporaryCodeTab.scala
```diff
@@ -2,170 +2,120 @@
...
-import org.nlogo.swing.{ FileDialog => SwingFileDialog, OptionDialog, ToolBar }
+import org.nlogo.core.I18N
+import org.nlogo.swing.{ FileDialog => SwingFileDialog, OptionDialog, ToolBarActionButton }
 import org.nlogo.window.{ Events => WindowEvents, ExternalFileInterface }
 import org.nlogo.workspace.AbstractWorkspace
 
 object TemporaryCodeTab {
-  val NewFile = "New File"
+  private var _serialNum = 1
+  def serialNum = {
```
Idloj: No, just an oversight

netlogo-gui/src/main/app/codetab/CodeTab.scala
```diff
@@ -50,7 +54,20 @@ class CodeTab(val workspace: AbstractWorkspace) extends JPanel
   override def zoomTarget = text
 
   val errorLabel = new EditorAreaErrorLabel(text)
-  val toolBar = getToolBar
+  val toolBar = new ToolBar {
```
Idloj: Why having a `val` that just calls the function? What do you gain by putting it in a function?

mrerrormessage: Here are the gains:

1) Removes clutter from initializer
2) Overriding a `val`s is tricky, especially if the `val` is acted upon in the initializer.
3) Overriding `val` leads both the subclass and the superclass to run their definitions, resulting in extra computation and an object which is never used.

Idloj: But we never override it (if, in some point in the future we will, we can change it then).
Regarding the clutter, I don't see

```
val toolbar = getToolbar
def getToolbar = {
  ...
}
```

as more cluttered than

```
val toolbar = {
  ...
}
```

Maybe even a bit less cluttered

mrerrormessage: It this was the last statement in the initializer, it might not be more cluttered, but there are several statements which follow this. So the comparison is

```scala
val toolBar = new ToolBar {
  // ...
}
val scrollableEditor = // ...
```

```scala
val toolBar = getToolBar
val scrollableEditor = // ...

// ...
def getToolBar = new ToolBar {
  // ...
}
```

I find the second much more reasonable.

Idloj: Then, we disagree on it,  but I'll make the change

mrerrormessage:
> What was the decision about the visibility and name of the Includes menu?
> Should I add an "always show" checkbox in the Preferences dialog? Should I change the name of the menu?

I think a preference with an "always show" checkbox is best. I can't remember the names we discussed for the menu, did you have a preferred name other than "Includes"?

On gitter, you asked me to respond to another question, but the link didn't work for me for some reason. Please let me know if there are other questions you have to get this wrapped up.

Idloj:
> This PR makes it so that when a file is compiled, it is first saved. This change was based on the behavior before #1198. It needs to be changed now that there is no need to save a file in order to compile the current version of it.

What do you think about this?

mrerrormessage:
I think "Included Files" makes sense. I think the changes which don't require saving included files make the behavior of includes files more consistent with the rest of the application since you don't have to save after modifying the code tab to have the model reflect your code changes.

netlogo-gui/src/main/app/Tabs.scala
```diff
@@ -235,39 +269,88 @@ class Tabs(val workspace:    GUIWorkspace,
...
   }
+
+  object SaveAction extends ExceptionCatchingAction(I18N.gui.get("menu.file.save"), this)
+  with MenuAction {
+    category    = UserAction.FileCategory
+    group       = UserAction.FileSaveGroup
+    accelerator = UserAction.KeyBindings.keystroke('S', withMenu = true)
+
+    @throws(classOf[UserCancelException])
+    override def action(): Unit = currentTab match {
```
Idloj: I've done that. Just wanted to ask, what is the benefit of doing this over acting on the current tab? (taking into consideration, that swapping the actions will happen much more often than clicking "Save")

netlogo-gui/src/main/app/Tabs.scala
```diff
@@ -93,132 +105,154 @@ class Tabs(val workspace:    GUIWorkspace,
...
+          openExternalFile(filename)
+          tab = getTabWithFilename(Right(filename))
+        }
+        if (e.error != null) setSelectedComponent(tab.get)
+        recolorTab(tab.get, e.error != null)
+        requestFocus()
+      case null => // i'm assuming this is only true when we've deleted that last widget. not a great sol'n - AZS 5/16/05
+        if(e.error != null) setSelectedIndex(0)
+        recolorTab(interfaceTab, e.error != null)
+      case jobWidget: JobWidget if !jobWidget.isCommandCenter =>
```
Idloj: >  I would prefer to see these two same cases moved into one case clause (using if or a separate helper function)

Can you illustrate what you mean (I tried back then to do `case null | jobWidget: ...` but it Scala doesn't like it) ?

mrerrormessage: I think you can do it like this (although it looks very strange): `case jobWidget: JobWidget if jobWidget == null || ! jobWidget.isCommandCenter =>`

Or you can keep two separate cases and just add a helper function (either in this method or a `private` method outside of it) which is called in both cases.

Idloj:
@mrerrormessage Can you make an `includes-enhancement` for ls like you did for gogo so I can open a PR there?

Idloj:
I opened a PR on LevelSpace

Idloj:
What is the relation between `netlogo-gui/project/autogen/i18n/Errors_*` and `shared/i18n/Errors_*` ? And, having moved `Errors_ja.txt` to `shared/i18n/`, is it the right place or should it be in `netlogo-gui/project/autogen/i18n/`?

mrerrormessage:
`netlogo-gui/project/autogen/i18n/Errors_*` is used in netlogo-gui, `shared/i18n/Errors_*` are used in the parser (and NLW). In general, if a string is used only in netlogo-gui, it should go in `netlogo-gui/project/autogen`, if it's used in any of `core`, `parse`, or `lex`, it should go in `shared` and in `netlogo-gui`. I *am* planning to clean this up in the near future as I understand that it's currently a mess.

Idloj:
I don't see the bug you mentioned about the Command Center.
Maybe it was solved as a result of the changes. Let me know if you still see this bug and what command fails you.

netlogo-gui/src/main/app/Tabs.scala
```diff
@@ -57,168 +63,187 @@ class Tabs(val workspace:    GUIWorkspace,
...
-    (3 until getTabCount)
-      .map(getComponentAt)
-      .collect{case tab: TemporaryCodeTab => tab}
-      .find(_.filename == name)
+  def getTabWithFilename(filename: Filename): Option[TemporaryCodeTab] =
+    externalFileTabs find (_.filename == filename)
 
-  def newTemporaryFile() { addNewTab(TemporaryCodeTab.NewFile, false) }
+  private var _externalFileNum = 1
+  private def externalFileNum = {
```
mrerrormessage: This method has a side effect and so should be `externalFileNum(): Int`.

mrerrormessage:
Upon Retesting report:

* The "Save" and "Save As" menu items don't appear on the Interface, Info, or Code Tabs. They do appear for "Included" code tabs.
* When running "Check" on a model which includes a non-existent ".nls" file, a NullPointerException is raised.

I did test "Save All", which appeared to work properly. Definitely needs to be retested with "Save" and "Save As" available for other tabs.

Idloj:
I'm ready for another review, I just need to resolve the conflicts, most of them I already did, what left is the PRs I opened in ls and gogo. Once they're merged, I can make this PR ready-to-merge.

mrerrormessage:
LS and GoGo PRs have been merged. I'm re-reviewing the changes now.

netlogo-gui/src/main/app/Tabs.scala
```diff
@@ -235,39 +262,55 @@ class Tabs(val workspace:    GUIWorkspace,
...
-    saveExternalFiles()
-  }
-
   def handle(e: AfterLoadEvent) {
     requestFocus()
   }
+
+  private val saveModelActions = {
+    def saveAction(saveAs: Boolean) =
+      new ExceptionCatchingAction(if (saveAs) I18N.gui.get("menu.file.saveAs") + Ellipsis else I18N.gui.get("menu.file.save"), Tabs.this)
```
mrerrormessage: Can this class be defined in `FileManager` instead of `Tabs`? `Tabs` is already big enough and this is not central to what `Tabs` does. It can be instantiated in here passing in `this` / `Tabs.this` as a constructor parameter.

This line needs at least one line break, maybe 2. Lines *should* be 80 characters or less where possible and should *never* be longer than 120 characters except in extreme circumstances.

Idloj:
I addressed the comments

mrerrormessage:
Thanks for the persistence and hard work on this!

I would really prefer that we not show the "Save All" option unless `__includes` is active. It gives the perception that "Save" doesn't save some part of the model, which is only true when there are `__includes` files.

In testing, I found The "Check" button in the code tab is disabled after a compiler exception occurs until there are changes made to the file. In general, this *would* be a good thing, but there are many times when we want to recompile when the tab isn't dirty. The most common is probably when a user makes changes, saves, and then wants to run "Check." there are many others around included files (for instance, a new ".nlogo" file is saved to a directory, an `__includes` target is saved, an `__includes` target is updated). Perhaps in the future we could consider disabling the "Check" button if we had a list of all the reasons to re-enable it and responded accordingly. For now, we should just leave it enabled all the time.

I'm under some time pressure to get some things regarding `__includes` fixed today. I'm happy to merge this and make both of the above changes (I expect that they will be quite small). Or if you would prefer to fix them, I would be happy to have you do that. But to get this into NetLogo 6, I will need a reply one way or the other in the next 2 hours. Sorry to give so little warning in advance about this deadline, but things have shaped up that way as we're preparing to release.

Idloj:
I'll do the changes now.

P.S. I really like your observation:

> I would really prefer that we not show the "Save All" option unless __includes is active. It gives the perception that "Save" doesn't save some part of the model, which is only true when there are __includes files.

mrerrormessage:
Thanks for the speedy response!

PR #1221, opened 2016-11-15
Remove patch size default assignment on load begin

### LINKED ISSUE #1197: OOM with models library models with varying dimensions
Steps to reproduce:

1. Open Kicked Rotator (or any other models library model with a very large patch size, but a small number of patches)
2. Open Fire (or any other model with a large number of patches)

Result: OOM

The OOM occurs here: https://github.com/NetLogo/NetLogo/blob/b6bdee202500e3b1ec70867ee0edcfdd02d28ffc/netlogo-gui/src/main/window/View.java#L171

I think what's happening is that the patch size from the previous model is being used with the dimensions of the new model. Indeed, printlning `getWidth` and `getHeight` there gives:

```
[info] 357
[info] 357
[info] 27108
[info] 27108
# OOM happens
[info] 502
[info] 502
```

Note that this *only* happens with models library models.

qiemem:
Fixes #1197

### DIFF
--- a/netlogo-gui/src/main/window/View.java
+++ b/netlogo-gui/src/main/window/View.java
@@ -326,8 +326,6 @@ public strictfp class View
 
   public void handle(org.nlogo.window.Events.LoadBeginEvent e) {
     setVisible(false);
-    patchSize = 13;
-    zoom = 0;
     renderer = workspace.newRenderer();
   }
 
--- a/netlogo-gui/src/main/window/View.java
+++ b/netlogo-gui/src/main/window/View.java
@@ -379,15 +379,15 @@ public strictfp class View
   }
 
   protected double patchSize = 13.0;
-  private double zoom = 0.0;
+  private double zoom = 1.0;
 
   public double patchSize() {
-    return patchSize + zoom;
+    return patchSize * zoom;
   }
 
   public void visualPatchSize(double patchSize) {
     double oldZoom = zoom;
-    zoom = patchSize - this.patchSize;
+    zoom = patchSize / this.patchSize;
     if (zoom != oldZoom) {
       renderer.resetCache(patchSize());
     }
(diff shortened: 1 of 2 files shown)


mrerrormessage:
I'll want to test this a bit, but I think it sounds good on the basis of what you've posted. Are there ever cases where *this change* would cause NetLogo to allocate a much larger view than needed? What happens if a prior model has a ludicrously large patch size and/or zoom and the new model has a large number of patches?

qiemem:
Ah, it actually still breaks if you're super zoomed in. Man, having zoom be an additive constant on the patch size seems pretty silly to me...

mrerrormessage:
Agreed. I'm not 100% sure about how to make the fix here, but I have a few ideas:

1. It appears that we set the view to be not visible between when the model starts loading and when the model finishes loading. This doesn't (apparently) prevent it from being rendered. Can we find a way to prevent that?
2. We set the renderer on on load begin. Can we set it to some do-nothing renderer for the duration of the loading period, then set it to the "real" renderer when the load finishes?
3. Can / should we disable drawing the whole interface area for the duration of model load?

I don't know whether you've attempted any of these or looked into them, I might have some time to investigate further later today.

qiemem:
These problems all derive from the fact that the magnitude of the effect of zoom depends on the patch size. Thus, patch size changes trigger changes in zoom level (which was the first problem) and patch size changes change how zoom is applied (which is what's happening now). Zoom really should be independent of patch size, instead applied directly as a scalar to the dimensions. That is, the dimensions of the view are currently calculated as:

```
worldWidth * (patchSize + zoom)
worldHeight * (patchSize + zoom)
```

when a more robust approach would be

```
worldWidth * patchSize * zoom
worldHeight * patchSize * zoom
```

This way, patch size and zoom level won't depend on each other at all. I'm honestly surprised this isn't how zoom already works.

mrerrormessage:
This looks good to me. 

While testing this, I uncovered a corner case in which the zoom and model size are such that they cause the graphics system to error due to arithmetic wraparound (width*height = area: `105969 * 105969 = -1655472927`). This PR doesn't cause this bug, but we should probably cap the zoom level at some point to prevent problems like that (or throw our own error to indicate that the view is too large).

PR #1225, opened 2016-11-18
Upgrade scala to 2.12

mrerrormessage:
At the moment we've disabled the parser js tests depend on a snapshot build of scala-parser-combinators.


SethTisue:
I've made a note to bug Antoine about (and offer assistance with) that release after a bit more time has passed.

mrerrormessage:
Thanks @SethTisue ! Although it was a bit annoying to have to build a single dependency just for us, the upgrade to scala 2.12 was actually much easier than I was expecting in terms of finding appropriately versioned dependencies. My hat is off to typesafe and the scala community for the fairly rapid upgrades that everyone has made!

netlogo-gui/src/main/agent/Torus3D.java
```diff
@@ -8,51 +8,59 @@
 public final strictfp class Torus3D
     extends Torus
     implements Topology3D {
+
+  // this class *ought* to be able to see `world` in Topology, but
+  // it can't, probably due to the scala class `Torus` between this
+  // class and Topology. in the meantime, adding this field is good enough
+  // for now.
+  final World _world;
```
SethTisue: is this a regression in scalac?

netlogo-headless/src/main/mirror/Mirrorables.scala
```diff
@@ -13,34 +13,38 @@ import org.nlogo.core.AgentVariables.implicitTurtleVariables
 object Mirrorables {
 
   case object Patch extends Kind {
-    object Variables extends Enumeration {
+    object PatchVariables extends Enumeration {
```
SethTisue: curious what motivated this and if there might be a scalac regression involved

netlogo-headless/src/main/prim/etc/_shuffle.scala
```diff
@@ -4,20 +4,20 @@ package org.nlogo.prim.etc
...
-import scala.collection.JavaConversions._
+import scala.collection.JavaConverters._
 
 import org.nlogo.core.LogoList
 import org.nlogo.nvm.{Reporter, Context}
 
 class _shuffle extends Reporter {
   def report(context: Context): AnyRef = {
-    val result: JList[AnyRef] = scala.collection.mutable.MutableList(argEvalList(context, 0): _*)
+    val result: JList[AnyRef] = scala.collection.mutable.MutableList(argEvalList(context, 0).scalaIterator.toSeq: _*).asJava
```
SethTisue: is this just a defensive copy, or what?

mrerrormessage: Yeah, I think it's a copy so that when it's shuffled it doesn't change the value returned by `argEvalList`.

mrerrormessage: `LogoList` is immutable, but we need to copy to a mutable data structure here so we can shuffle in the line below.

netlogo-headless/src/main/mirror/Mirrorables.scala
SethTisue: `abstract class C { def x: AnyRef }; object O extends C { object x }` compiles fine in both 2.11.8 and 2.12.0, as does `abstract class Kind { val Variables: Enumeration } ; case object Patch extends Kind { object Variables extends Enumeration }`, so there must be something else/more here...?

netlogo-headless/src/main/prim/etc/_shuffle.scala
SethTisue: I think the entire sequence gets copied twice here: once from `LogoList` to `Stream`, then again from `Stream` to `MutableList`. Surely it's possible to get away with only one copy?

mrerrormessage: I see what you mean now, and yes, I think this is avoidable.

netlogo-gui/src/main/agent/Torus3D.java
mrerrormessage: So, the following breaks:
```java
abstract class A {
  final String s;

  A(String string) {
    s = string;
  }
}

class C extends B {
  C(String string) {
    super(string);
  }

  public String foo() {
    return s;
  }
}
```

```scala
class B(s: String) extends A(s) {
  def bar = s
}
```
But if the method `bar` is removed from `B` so that `B` never references `s`, it's fine. Alternatively, it's also fine if `B`'s constructor parameter is named `_s`. I'm not quite sure how this actually ever worked, as my minimal test case failed compilation in the same way in 2.11.7, 2.11.5, and 2.12.0. I've decided that rather than adding this field, it's easiest to rename the constructor parameter in `Torus`. I'm happy to open a bug for this if you would like.

netlogo-headless/src/main/mirror/Mirrorables.scala
mrerrormessage: Well, I changed it back and it worked....not sure what was going on before - maybe I hadn't cleaned the project thoroughly enough.

netlogo-gui/src/main/agent/Torus3D.java
SethTisue: ah. this seems confused/confusing enough that filing a bug certainly might be appropriate, but probably isn't a good use of your time.

netlogo-headless/src/main/mirror/Mirrorables.scala
mrerrormessage: After reverting this change, the following error shows up in tests:

`scala.UninitializedFieldError: Uninitialized field: /Users/rgg284/IdeaProjects/NetLogo/netlogo-headless/src/main/mirror/Mirrorables.scala: 43 (or 37)`

I'm unsure what's causing this, but it's definitely quite subtle and I can only assume has to do with `lazy val`-style initialization. I've spent about 15 minutes trying to find a minimal case and I've been unable to do so thus far. I'm going back to the old fix and we can revisit this at a later time (since this isn't in `api` or `core` we can introduce breaking changes later).

SethTisue:
🎉

PR #1231, opened 2016-11-23
Indent tweaks

qiemem:
- Indents `] [` correctly
~~- Matches indenting in cases of custom indenting after opener, e.g.:~~ *this resulted in weird corner cases*
- Makes all indent triggers behave consistently.


qiemem:
Shouldn't be merged yet. There's some funny cases with the custom indentation alignment.

Cases that need to be resolved:

```
let func [ [x] ->
  x + 1
]
```

```
let foo (list
  1
  2
  3
)
```

qiemem:
Should be ready to go!

PR #1244, opened 2016-12-09
Simplify and optimize link lookup

qiemem:
Moves link storage to turtles. This drastically simplifies the code while also speeding it up (by eliminating the hash map lookups in the link manager).

Memory overhead shouldn't be too bad: each turtle is initialized with a single, zero-capacity `ArrayList`. If we think that's too much, I could initialize this list to null, reducing the overhead to a pointer-per-turtle for models that don't use links.

### DIFF
--- a/netlogo-gui/src/main/agent/LinkManager.scala
+++ b/netlogo-gui/src/main/agent/LinkManager.scala
@@ -97,108 +97,15 @@ trait LinkManager {
 //
 
 
-class LinkMap {
-
-  // These should NEVER be iterated over if the order of iteration is
-  // important. It will result in non-determinism in simulations.
-  // -- BCH 10/22/2016
-  private val undirected = mutable.Map[Turtle, ArrayBuffer[Link]]()
-  private val outgoing = mutable.Map[Turtle, ArrayBuffer[Link]]()
-  private val incoming = mutable.Map[Turtle, ArrayBuffer[Link]]()
-
-  def addLink(link: Link) = {
-    if (link.isDirectedLink) {
-      outgoing.getOrElseUpdate(link.end1, ArrayBuffer.empty[Link]) += link
-      incoming.getOrElseUpdate(link.end2, ArrayBuffer.empty[Link]) += link
-    } else {
-      undirected.getOrElseUpdate(link.end1, ArrayBuffer.empty[Link]) += link
-      undirected.getOrElseUpdate(link.end2, ArrayBuffer.empty[Link]) += link
-    }
-  }
-
-  def removeLink(link: Link) = {
-    if (link.isDirectedLink) {
-      outgoing(link.end1) -= link
-      incoming(link.end2) -= link
-    } else {
-      undirected(link.end1) -= link
-      undirected(link.end2) -= link
-    }
-  }
-
-  /**
-    * Returns all links connected to the given turtle that match one of the
-    * given includes as an array.
-    * @return The array containing all matching links. This array is a *copy*
-    *         and may be safely mutated by the caller. An array is returned
-    *         for performance reasons, as well as for convenient consumption
-    *         by Java and convenient (and fast) conversion to an AgentSet.
-    *
-    */
-  def links(turtle: Turtle, includeUndirected: Boolean = false,
-                            includeOutgoing:   Boolean = false,
-                            includeIncoming:   Boolean = false): Array[Link] = {
-    var size = 0
-
-    val un = if (includeUndirected && undirected.nonEmpty)
-      undirected.get(turtle).map { ls => size += ls.size; ls }
-    else None
-
-    val out = if (includeOutgoing && outgoing.nonEmpty)
-      outgoing.get(turtle).map { ls => size += ls.size; ls }
-    else None
-
-    val in = if (includeIncoming && incoming.nonEmpty)
-      incoming.get(turtle).map { ls => size += ls.size; ls }
-    else None
-
-    val result = new Array[Link](size)
-    var i = 0
-
-    un.foreach { ls => ls.copyToArray(result, i); i += ls.size }
-    out.foreach { ls => ls.copyToArray(result, i); i += ls.size }
-    in.foreach { ls => ls.copyToArray(result, i); i += ls.size }
-
-    result
-  }
-
-  /**
-   * Removes the turtle from the maps. Note that this only removes that
-   * the given turtle's entries from the maps. It does NOT remove the turtle's
-   * links that appear in other turtles' entries.
-   */
-  def removeTurtle(turtle: Turtle) = {
-    undirected -= turtle
-    outgoing -= turtle
-    incoming -= turtle
-  }
-
-  def clear() = {
-    undirected.clear
-    outgoing.clear
-    incoming.clear
-  }
-
-}
-
-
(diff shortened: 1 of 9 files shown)


qiemem:
Related: I've added benchmarks for directed links here: https://github.com/NetLogo/models/pull/318

mrerrormessage:
@qiemem , can you update the models submodule to point to the branch with the benchmark, as well as checking in the appropriate benchdump for that model?

SethTisue:
> Memory overhead shouldn't be too bad: each turtle is initialized with a single, zero-capacity ArrayList

fwiw, we didn't do it that way originally exactly for fear of this overhead. (also, secondarily, for modularity of the code.)

netlogo-gui/src/main/agent/Turtle.java
```diff
@@ -1238,4 +1241,38 @@ public int alpha() {
...
+  public Link[] selectLinks(boolean undirected, boolean out, boolean in, AgentSet breed) {
+    Link[] result = new Link[links.size()];
+    int writeTo = 0;
+    for (Link link : links) {
+      boolean isDir = link.isDirectedLink();
+      boolean isOut = isDir && link.end1 == this;
+      boolean isIn = isDir && link.end2 == this;
+      boolean passesDirectedness = (undirected && !isDir) || (out && isOut) || (in && isIn);
+      boolean passesBreededness = breed == world.links() || breed == link.getBreed();
+      if (passesBreededness && passesDirectedness) {
```
mrerrormessage: Are we expecting this code to be in a tight loop? (Either the `for` loop here, or something calling `selectLinks` itself in a tight loop).

qiemem:
1,000,000 turtles, used heap after gc (some error due to delay between gc and reading the value)

```
beta2: 397,231,536 B
in-turtle link list (null): 404,567,264 B
in-turtle link list (empty): 428,176,736 B
```

So about a 7% memory hit over beta2. I'm going to initialize to null instead.

qiemem:
I don't understand the travis failure:

```
[info] - ANN *** FAILED ***
[info]   java.io.FileNotFoundException: /home/travis/build/NetLogo/NetLogo/models/test/benchmarks/ANN Benchmark.nlogo (No such file or directory)
[info]   at java.io.FileInputStream.open(Native Method)
[info]   at java.io.FileInputStream.<init>(FileInputStream.java:138)
[info]   at scala.io.Source$.fromFile(Source.scala:91)
[info]   at scala.io.Source$.fromFile(Source.scala:76)
[info]   at scala.io.Source$.fromURI(Source.scala:121)
[info]   at org.nlogo.fileformat.AbstractNLogoFormat.$anonfun$sections$1(NLogoFormat.scala:38)
[info]   at org.nlogo.fileformat.AbstractNLogoFormat$$Lambda$1802/2009012727.apply(Unknown Source)
[info]   at scala.util.Try$.apply(Try.scala:209)
[info]   at org.nlogo.fileformat.AbstractNLogoFormat.sections(NLogoFormat.scala:37)
[info]   at org.nlogo.fileformat.AbstractNLogoFormat.sections$(NLogoFormat.scala:35)
[info]   ...
```

but [it's definitely there](https://github.com/NetLogo/models/blob/f2a24ce85bd32c8256c121884af3217c4afab05d/test/benchmarks/ANN%20benchmark.nlogo). The test passes locally as well.

mrerrormessage:
The test appears to be looking for `ANN Benchmark.nlogo`, it looks like the file in the models library is `ANN benchmark.nlogo`. I admit to being very confused as to how this is passing locally for you, but the case-sensitivity of file resolution likely varies somewhat between java implementations or perhaps your working copy has a file that's uppercased?

qiemem:
Ah! Good catch. I should've noticed.

HFS+ is case insensitive by default, but is case preserving. For instance, when I changed the name of the model, git didn't realize anything had changed until I did `git mv`. http://apple.stackexchange.com/a/22304/67447

Most Linux filesystems are case sensitive, I believe.

netlogo-gui/src/main/agent/LinkManager.scala
```diff
@@ -216,23 +123,23 @@ class LinkManagerImpl(world: World, linkFactory: LinkFactory) extends LinkManage
   }
 
   def addLink(link: Link) = {
-    allLinks.addLink(link)
     if (link.getBreed eq world.links) {
+      // FIXME: Wait, does this make sense? `isDirectedLink` is calculated from the breed...
```
mrerrormessage: Can you explain this comment? It definitely needs to be removed before merge. Are you confident that this code is correct?

netlogo-gui/src/main/agent/LinkManager.scala
```diff
@@ -241,14 +148,7 @@ class LinkManagerImpl(world: World, linkFactory: LinkFactory) extends LinkManage
...
 
-  def cleanupTurtle(turtle: Turtle) {
-    links(turtle, world.links).foreach(_.die())
-    allLinks.removeTurtle(turtle)
-    breededLinks.valuesIterator.foreach(_.removeTurtle(turtle))
-  }
-
-  private def linkMapsForBreed(linkBreed: AgentSet): LinkMap =
-    if (linkBreed eq world.links) allLinks else breededLinks.getOrElseUpdate(linkBreed, new LinkMap())
+  def cleanupTurtle(turtle: Turtle) = turtle.links.foreach(_.die)
```
mrerrormessage: Since `die` is a side-effecting method, add parens.

netlogo-gui/src/main/agent/LinkManager.scala
```diff
@@ -269,25 +169,15 @@ class LinkManagerImpl(world: World, linkFactory: LinkFactory) extends LinkManage
...
-    val linkMaps = linkMapsForBreed(linkBreed)
-    linkMaps.links(target, includeUndirected = true, includeIncoming = true)
-  }
+  def inLinks(target: Turtle, linkBreed: AgentSet): Array[Link] = target.selectLinks(true, false, true, linkBreed)
 
-  def links(turtle: Turtle, linkBreed: AgentSet): Array[Link] = {
-    val linkMaps = linkMapsForBreed(linkBreed)
-    linkMaps.links(turtle, includeUndirected = true, includeOutgoing = true, includeIncoming = true)
-  }
+  def links(turtle: Turtle, linkBreed: AgentSet): Array[Link] = turtle.selectLinks(true, true, true, linkBreed)
```
mrerrormessage: Add line break here and 176, 178, and 192.

mrerrormessage:
Had two small formatting comments, once those are taken care of, this is ready to merge.

qiemem:
So, remember that code that I said didn't do anything? Specifically, this code: `world.links.setDirected(link.isDirectedLink)`. Turns out it was masking this NW bug in the NW language tests: NetLogo/NW-Extension#129. Anyway, trying to figure out a solution.

mrerrormessage:
Cool, let me know if you figure out anything. It's pretty late for radical changes, obviously, so small, conservative fixes are preferred (at least on my end).

qiemem:
Looks like travis is happy. If you're okay with the way I fixed that issue, this should be good to go.

mrerrormessage:
OK, this looks good to me. Thanks for all the hard work on this! Sorry that NetLogo's link model is such a pain.

Issue #1222, opened 2016-11-17
Jump to declaration enhancements


mrerrormessage:
Jump to declaration misses a few cases. For instance, trying to jump-to-definition of `countdown` doesn't work as expected. Some people have also requested that jump-to-definition work on procedures (when those are available). If there are other enhancements desired, please leave a comment on this issue.

nicolaspayette:
+1 for jumping to definition of procedures. That would be far more convenient than using the "procedures" menu in the toolbar. Especially if there was a keyboard shortcut for it. Can we please get a keyboard shortcut for it? :-)

nicolaspayette:
"Jump to declaration" is an awesome feature, and I use it all the time, but hardly a day goes by without me wishing there was a keyboard shortcut for it...

(see also #1399)

PR #1246, opened 2016-12-09
Enhan jump to function

Akarshit:
Fix for #1222

### DIFF
--- a/netlogo-gui/src/main/ide/JumpToDeclaration.scala
+++ b/netlogo-gui/src/main/ide/JumpToDeclaration.scala
@@ -32,7 +32,7 @@ object JumpToDeclaration {
 
   def getDeclaration(token: Token, source: String): Option[Token] = {
     val tokens = Femto.scalaSingleton[TokenizerInterface]("org.nlogo.lex.Tokenizer").tokenizeString(source).filter(t => t.tpe != TokenType.Comment).toSeq
-    val (globals, owns) = parseOwns(tokens)
+    val (globals, owns, functions) = parseOuterScope(tokens)
     val indexOfToken = tokens.indexWhere(t => t.start == token.start)
     var stack = 0
     for(i <- indexOfToken to 0 by -1) {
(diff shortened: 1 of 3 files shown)


mrerrormessage:
Thanks for the PR! I had one comment, but other than that, this looks terrific! Looking forward to having functions built into jump-to-declaration.

netlogo-gui/src/main/ide/JumpToDeclaration.scala
```diff
@@ -58,15 +58,22 @@ object JumpToDeclaration {
...
+    * @param tokens list of all the tokens in the editor
+    * @return tuple of globals, owns, functions. Each value is a map
+    *         from token text to token.
+    */
+  def parseOuterScope(tokens: Seq[Token]): (Map[String, Token], Map[String, Token], Map[String, Token]) = {
     val globals = Map[String, Token]()
     val owns = Map[String, Token]()
+    val functions = Map[String, Token]()
     var i = 0
     while(i < tokens.length) {
```
mrerrormessage: This is much better. Thanks for looking at it!

mrerrormessage:
I'd like this to be squashed and/or rebased down to a single commit, then this can be merged. Thanks again for the PR, we're very excited to have this change!

Akarshit:
@mrerrormessage There is something wrong while building the submodule. How do I fix that in this PR?

mrerrormessage:
It appears you have this tied to a revision of the `r` submodule which no longer exists. Hexy points to `a74e519`, so this branch should also point there.

Quick reference on updating submodule commit. Starting from NetLogo root directory:
```bash
> cd extensions/r
> git fetch
> git co a74e519
> cd ../../
> git add extensions/r
> git commit --amend # this attaches it to the prior commit, like we want for this PR. In other situations you can just use git commit normally
```

mrerrormessage:
Looks great. Thanks!

Issue #214, opened 2012-09-13
missing list primitive: `insert` (or `insert-item`)


SethTisue:
the syntax would be the same as `replace-item`, but it would insert instead of replacing. like this:

``` text
to-report insert [n xs x]
  report sentence (sublist xs 0 n) fput x (sublist xs n length xs)
end
```

mrerrormessage:
fixed in #1296

PR #1296, opened 2017-01-26
Add primitive `insert-item` [issue #214]

brotatotes:


parser-core/src/main/core/LogoList.scala
```diff
@@ -53,6 +53,8 @@ class LogoList private (private val v: Vector[AnyRef])
   override def reverse = new LogoList(v.reverse)
   def replaceItem(index: Int, obj: AnyRef) =
     new LogoList(v.updated(index, obj))
+  def insertItem(index: Int, obj: AnyRef, elt: AnyRef) =
+    new LogoList(v.slice(0, index) ++ Vector(elt) ++ v.slice(index, v.size))
```
SethTisue: would `Vector#patch` be helpful here?

mrerrormessage:
This looks good. There's a few tests still failing (not a bug, the tests just need to be updated). We'll look at the test failures together on Wednesday.

PR #1298, opened 2017-01-27
Speed indenter

mrerrormessage:
This substantially speeds the indenter (30x or more, by my calculations). The existing behavior has been maintained to the extent it was specified in the tests and new tests have been added clarifying behavior, although these should be considered descriptive.

@qiemem , would appreciate a quick look at the test cases if you have time.

### DIFF
--- a/netlogo-gui/src/main/app/codetab/SmartIndenter.scala
+++ b/netlogo-gui/src/main/app/codetab/SmartIndenter.scala
@@ -9,17 +9,68 @@ import org.nlogo.editor.Indenter
 class SmartIndenter(code: EditorAreaInterface, compiler: CompilerServices)
 extends Indenter {
 
-  /// first, the four handle* methods in IndenterInterface
+  private val TAB_WIDTH = 2
 
+  // TokenizedLine represents a line that has been tokenized and is ready to be evaluated
+  // for indentation.
+  case class TokenizedLine(lineNum: Int, lineStart: Int, lineEnd: Int, text: String, leadingSpaces: Int, tokens: Seq[Token]) {
+    // bracketDelta lists the change in number of brackets.
+    // "" = 0, "[" = 1, "]" = -1, "[]" = 0, "[[" = 2, &c
+    lazy val bracketDelta = tokens.map(_.tpe).foldLeft(0) {
+      case (i, (TokenType.OpenBracket | TokenType.OpenParen))  => i + 1
+      case (i, (TokenType.CloseBracket | TokenType.CloseParen)) => i - 1
+      case (i, _) => i
+    }
+    // bracketsClosed lists the number of unmatched closing brackets on this line.
+    // "" = 0, "[" = 0, "]" = 1, "[]" = 0, "]]" = 2, "[]]" = 1, &c
+    lazy val bracketsClosed: Int = {
+      val (_, closedCount) = tokens.map(_.tpe).foldLeft((0, 0)) {
+        case ((currentDelta, maxClosed), (TokenType.CloseBracket | TokenType.CloseParen)) =>
+          (currentDelta + 1, maxClosed max (currentDelta + 1))
+        case ((currentDelta, maxClosed), (TokenType.OpenBracket | TokenType.OpenParen)) =>
+          (currentDelta - 1, maxClosed)
+        case ((currentDelta, maxClosed), _) => (currentDelta, maxClosed)
+      }
+      closedCount
+    }
+    lazy val finalCommentStart: Option[Int] =
+      tokens.lastOption.filter(_.tpe == TokenType.Comment).map(_.start - lineStart)
+    lazy val isOnlyComment: Boolean = tokens.length == 1 && tokens.head.tpe == TokenType.Comment
+  }
+
+  // LineIndent and the classes that extend it track operations needing to be performed on
+  // a particular line.
+  sealed trait LineIndent {
+    def lineNum: Int
+    def lineStart: Int
+    def lineEnd: Int
+    def text: String
+    def delta: Int
+  }
+  case class AddIndent(lineNum: Int, lineStart: Int, lineEnd: Int, text: String, delta: Int) extends LineIndent
+  case class MaintainIndent(lineNum: Int, lineStart: Int, lineEnd: Int, text: String) extends LineIndent {
+    def delta: Int = 0
+  }
+  case class RemoveIndent(lineNum: Int, lineStart: Int, lineEnd: Int, text: String, delta: Int) extends LineIndent
+
+  /// first, the four handle* methods in IndenterInterface
   def handleTab() = {
-    val line1 = code.offsetToLine(code.getSelectionStart)
-    val line2 = code.offsetToLine(code.getSelectionEnd)
-    (line1 to line2).foreach(indentLine)
+    val indentations = lineIndentation(code.getSelectionStart, code.getSelectionEnd)
+    if (indentations.nonEmpty) {
+      executeIndentations(indentations)
+    }
   }
 
   def handleEnter() = {
     code.replaceSelection("\n")
-    indentLine(code.offsetToLine(code.getSelectionEnd))
+    val line = code.offsetToLine(code.getSelectionEnd)
+    val start = code.lineToStartOffset(line)
+    val end = code.lineToEndOffset(line)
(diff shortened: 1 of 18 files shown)


test/indent.txt
```diff
@@ -101,6 +163,13 @@ to foo ; comment
 + ask turtles [ fd 1 ]
 end
 
+Comments3
```
qiemem: Not crazy about this test case. Seems like the kind of thing you would *imagine* people would want rather than something people would *actually* want. If you need to do a multi-line comment, you do it before the line in question.

test/indent.txt
```diff
@@ -19,12 +19,74 @@ to go
...
+  if true [ set abc [
+    1 2 3 ]
+  ] [
+++  set abc [4 5 6]
+  ]
+end
+
+# I'm not certain this should be normative
+# This behavior wasn't specified prior to 6.0.1 and could probably change
+IfMultiBracket2
```
qiemem: Related problem. Currently, this happens:

```
to foo
  let abc 0
  if true [ set abc
    [
      1 2 3
--    ] ]
...
```

which is wrong. Also, adding an open bracket on the end then does this:

```
to foo
  let abc 0
  if true [ set abc
    [
      1 2 3
++    ] ] [
```

This also seems wrong, as an open bracket shouldn't change the current line's indentation level. `++` should happen whether or not the open bracket is there.

test/indent.txt
```diff
@@ -19,12 +19,74 @@ to go
...
+to foo
+  let abc 0
+  if true [ set abc
```
qiemem: This should be `ifelse`. While it doesn't technically matter right, if we ever get an argument-aware indenter, it will.

test/indent.txt
qiemem: Anyway, in response to your comment, this is a weird enough case that being normative about it seems odd. That said, it looks right to me. I don't know. Could go either way.

test/indent.txt
```diff
@@ -19,12 +19,74 @@ to go
...
+  ]
+end
+
 IfPreindented
 to go
   if true
     [show 1]
 end
 
+IfPreindentedWithPriorLineMessedUp
```
qiemem: Yes!

test/indent.txt
```diff
@@ -19,12 +19,74 @@ to go
...
+
+FunctionAfterGlobals
+globals [
+++abc
+]
+to foo [def]
+-  do-something
+end
+
+## Again, this behavior wasn't specified prior to 6.0.1 and could probably be changed
```
qiemem: This looks like user code. It should probably be scrubbed and simplified with `foo`s and `bar`s. But again, it's weird enough that being normative about it feels odd.

test/indent.txt
mrerrormessage: I can rewrite this case. This case is designed to test something that the prior comment tests do not test, which is:

* multi-line comment alignment
* bottom comment line starts out of alignment
* top comment line starts out of alignment

Because we now make all indentation changes in a single pass (rather than making changes, re-evaluating, making more changes) this case is particularly fragile and needs a test. Having said that, I agree that this case doesn't actually look like code someone would write.

test/indent.txt
mrerrormessage: Yeah, thanks for the feedback on this. I ended up breaking it down into several cases and discovered an even better indentation strategy where it looks through the indentation stack for the last bracket matched instead of the delta between openers and closers.

mrerrormessage:
Comments have been addressed. Have a look if you would like, otherwise I plan to merge this tomorrow afternoon.

test/indent.txt
qiemem: Were you going to modify this test case?

netlogo-gui/src/main/app/codetab/SmartIndenter.scala
```diff
@@ -9,17 +9,68 @@ import org.nlogo.editor.Indenter
...
+  // for indentation.
+  case class TokenizedLine(lineNum: Int, lineStart: Int, lineEnd: Int, text: String, leadingSpaces: Int, tokens: Seq[Token]) {
+    // bracketDelta lists the change in number of brackets.
+    // "" = 0, "[" = 1, "]" = -1, "[]" = 0, "[[" = 2, &c
+    lazy val bracketDelta = tokens.map(_.tpe).foldLeft(0) {
+      case (i, (TokenType.OpenBracket | TokenType.OpenParen))  => i + 1
+      case (i, (TokenType.CloseBracket | TokenType.CloseParen)) => i - 1
+      case (i, _) => i
+    }
+    // bracketDelta lists the number of brackets closed on this line.
```
Idloj: `bracketDelta` should be `bracketsClosed` and the statement might be changed to the following for clarification:

`bracketClosed lists the number of unmatched closing brackets on this line.`

netlogo-gui/src/main/app/codetab/SmartIndenter.scala
```diff
@@ -33,167 +84,151 @@ extends Indenter {
...
-        val lineStart = code.lineToStartOffset(lineNum)
-        if(newSpaces > oldSpaces)
-          code.insertString(lineStart, spaces(newSpaces - oldSpaces))
-        else
-          code.remove(lineStart, oldSpaces - newSpaces)
+  private def executeIndentations(indentations: Seq[LineIndent]): Unit = {
+    val start = indentations.head.lineStart
+    val finish = indentations.last.lineEnd
+    val builder = new StringBuilder(finish - start)
+    val offset = indentations.foldLeft(0) { (offset: Int, indent: LineIndent) =>
```
Idloj: Is `offset` (the `val`) used anywhere?

netlogo-gui/src/main/app/codetab/SmartIndenter.scala
```diff
@@ -33,167 +84,151 @@ extends Indenter {
...
+      val newOffset = (offset, indent) match {
+        case (offset, AddIndent(lineNum, lineStart, _, text, delta)) =>
```
Idloj: This and the subsequent matches can be matched with something like `AddIndent(_, _, _, text, delta)`, can't they?

netlogo-gui/src/main/app/codetab/SmartIndenter.scala
```diff
@@ -33,167 +84,151 @@ extends Indenter {
...
-      // see where we are in relation to the command if we're indented past the command move to 1
-      // tab stop past if we're before we move to be in line with the command
-      val cmd = findCommand(prevLine)
-      if (token != null && isOpener(token) && cmd.isDefined &&
-        countLeadingSpaces(currentLine) > result)
-        return Some(result + TAB_WIDTH)
-    }
-    Some(result)
+  private def lineIndentation(startOffset: Int, endOffset: Int): Seq[LineIndent] = {
+    val startLine = code.offsetToLine(startOffset)
```
Idloj: This doesn't seem to be used anywhere

test/indent.txt
```diff
@@ -19,12 +19,173 @@ to go
...
+  let abc 0
+  if true [
+    set abc
+    [
+      1 2 3
+    ]
+  ]
+end
+
+IfMultiBracket3
```
Idloj: Isn't this a duplicate of the one above?

mrerrormessage: Nice catch!

netlogo-gui/src/main/app/codetab/SmartIndenter.scala
mrerrormessage: nice catch

netlogo-gui/src/main/app/codetab/SmartIndenter.scala
mrerrormessage: Yep, changing.

netlogo-gui/src/main/app/codetab/SmartIndenter.scala
mrerrormessage: Nope. Removing.

netlogo-gui/src/main/app/codetab/SmartIndenter.scala
mrerrormessage: Nice catch!

mrerrormessage:
@Idloj , thanks for the review. I've incorporated the changes and I'll be merging as soon as Travis passes.

PR #1327, opened 2017-02-17
Speedup in radius

mrerrormessage:
The general structure of `InRadiusOrCone` has changed. Now `InRadiusOrCone` constructs several different potential strategies for determining the agents in the radius (or cone), decides which of those strategies apply and applies the cheapest applicable strategy to determining the agents. I'm not 100% finished with optimizations and tweaks, but thought I would post this to get feedback about the general approach.


netlogo-gui/src/main/agent/AgentSet.scala
```diff
@@ -34,6 +34,10 @@ abstract class AgentSet(
   def randomSubsetGeneral(resultSize: Int, precomputedCount: Int, rng: api.MersenneTwisterFast): Array[Agent]
   def toLogoList: LogoList
 
+  // this is an upper bound on the size of the agentset. It's used in heuristics, so it should
+  // be free or cheap to compute.
+  private[agent] def sizeBound: Int
```
qiemem: Why have this be private to `agent`? Seems like it could be useful in extensions (though shouldn't be in the API).

netlogo-gui/src/main/agent/InRadiusOrCone.scala
```diff
@@ -0,0 +1,466 @@
...
+      p.turtlesHere match {
+        // this optimization may seem minor, but actually results in ~ 5-10% speedup
+        // when finding any turtles in-radius
+        case arrayList: JArrayList[_] =>
+          var i = 0
+          while (i < arrayList.size) {
+            f(arrayList.get(i).asInstanceOf[Turtle])
+            i += 1
+          }
+        case turtles =>
```
qiemem: Just to be clear, this branch can never be taken because `turtlesHere` always returns a `JArrayList`; you're just branching because it's only declared to return `Iterable`, right?

mrerrormessage: That's correct. We never expect this branch to be taken.

netlogo-gui/src/main/agent/AgentSet.scala
mrerrormessage: My view is that this information is really internal information about the `AgentSet` and shouldn't be visible outside of the package. I don't want extensions to know about it because I think some extension authors would use it instead of `count` (for instance). While `count` always returns an accurate result, the result returned by this can be wildly inaccurate (and I consider the amount of inaccuracy likely to change in the future). 

If extension authors want access, they can use reflection to get access to this method. While I would discourage them from doing so, it puts a big warning label on top of this method saying "this is subject to change or removal" and forces them to account for that in their code.

mrerrormessage:
This change will alter model checksums and so needs to wait for 6.1.0. I will be closing this PR and merging in the initial version (which is a much smaller change) instead.

PR #1339, opened 2017-02-23
Put link primitves next to patch and turtle prims

Idloj:
- Add version info to `__includes`
- Remove "Release Notes" from the headings, it looked like
  a section heading for the four items beneath it.
- Rename "What's New?" to "Release Notes"

I changed `is-anonymous-...`  prims' "since version" to 6.0, because when I made the version marks, the rule was that the marked prim must be literally available at the mentioned version, not its meaning. (This rule can be argued of course).

(Sorry for the bulk of doc PRs, should've been one PR)

### DIFF
--- a/autogen/docs/headings.html.mustache
+++ b/autogen/docs/headings.html.mustache
@@ -69,9 +69,8 @@
   <span class="block offwhitetext bold nowrap">{{date}}</span>
 </div>
 <div style="margin-bottom: 0em; margin-top: .25em;">
-  <span class="heading-title xsmallfont">Release Notes</span>
   <ul class="small-items">
-    <li><a href="versions.html"     target="entry">What's New?</a></li>
+    <li><a href="versions.html"     target="entry">Release Notes</a></li>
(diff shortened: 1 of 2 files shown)


autogen/docs/headings.html.mustache
```diff
@@ -69,7 +69,6 @@
   <span class="block offwhitetext bold nowrap">{{date}}</span>
 </div>
 <div style="margin-bottom: 0em; margin-top: .25em;">
-  <span class="heading-title xsmallfont">Release Notes</span>
```
mrerrormessage: Why was this heading removed?

Idloj: It looked like a section heading for the four items beneath it (I recorded the reason in the commit message). "What's new?" can be changed to "Released Notes", of course, but the current position & look of "Release Notes" is *very* confusing.

mrerrormessage: Yes, let's rename "What's new?" to "Release Notes". I agree this was somewhat confusing.

PR #1349, opened 2017-02-27
Prevent error from showing in plot on halt

qiemem:


mrerrormessage:
So I think this approach is generally sound, but `plot` isn't allowed to depend on `nvm`, which contains `HaltException`. So we might have to rethink this a little. I'll comment again tomorrow if I have any good ideas.

qiemem:
Moving `HaltException` to `api` seems like a pretty reasonable thing. I'm a little surprised it wasn't there already.

qiemem:
Build log says

> The job exceeded the maximum time limit for jobs, and has been terminated.

netlogo-core/src/main/api/HaltSignal.scala
```diff
@@ -0,0 +1,3 @@
+package org.nlogo.api
```
mrerrormessage: 🤖 requires copyright notice 🤖

mrerrormessage:
Just needs a copyright notice in one file. Other than that it should be good. The push build failed with a timeout, but the PR build passed, so I'm going to chalk the failure up to Travis flakiness.

PR #1367, opened 2017-03-20
Add extname to manifest error messages (fix #662)

### LINKED ISSUE #662: Manifest Error Specification
A helpful addition to NetLogo would be to have the manifest warning messages for extensions explicitly state what extension's manifest has caused the warning to be created.  The Netlogo model I work with uses four extensions and determining which manifests are erroneous and which aren't can be a little frustrating.  Its not hard to debug mostly but this little piece of extra information would save a lot of time!

mister-walter:
- Added an extName parameter to several of the ExtensionManagerException subclasses, and updated their messages to include it.
- Updated test "extensionData raises an error when the jar manifest doesn't have the required fields" in JarLoaderTests to check that the exception's message contains the extension name.
- Updated JarLoader and InMemoryExtensionLoader to provide ExtensionManagerException subclasses with the extension name as appropriate.

### DIFF
--- a/netlogo-gui/src/main/workspace/ExtensionManagerException.scala
+++ b/netlogo-gui/src/main/workspace/ExtensionManagerException.scala
@@ -8,20 +8,20 @@ object ExtensionManagerException {
   case class ExtensionNotFound(extName: String) extends Cause {
     override def message: String = ExtensionManager.EXTENSION_NOT_FOUND + extName
   }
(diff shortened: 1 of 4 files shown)


mrerrormessage:
This looks awesome! I plan to test and merge this tomorrow. Thanks!

mrerrormessage:
Looks great. Thanks again!

PR #1373, opened 2017-04-02
Revising y-axis plot label's maximum and minimum value (fix #1105)   

### LINKED ISSUE #1105: Limit precision when using scientific notation in plot axis labels
In situation like these:

untitled

I'd much rather loose a few digits of precision and conserve plotting area real-estate.

We _could_ make the precision configurable (like for monitors) but I think that capping it to 3 or 4 would be entirely reasonable.

Shreyanand:


mrerrormessage:
In c13a71a I could configure a plot to display the following:
screen shot 2017-04-04 at 9 51 55 am

mrerrormessage:
No activity in 3 months, closing. @Shreyanand , if you feel like working on this again, please re-open.

PR #1380, opened 2017-04-09
Use awt FileDialog on Linux (in org.nlogo.swing.FileDialog). Closes #1375

### LINKED ISSUE #1375: Should file selection dialogs on Linux use awt's FileDialog instead of swing's JFileChooser?
Currently, `org.nlogo.swing.FileDialog.show` uses JFileChooser in two cases: Linux and directory choosing.

Directory choosing uses JFileChooser because (as I understand it, anyway) then we can do `chooser.setFileSelectionMode(JFileChooser.DIRECTORIES_ONLY)` to only show directories.

As for Linux, there is a [comment](https://github.com/NetLogo/NetLogo/blob/hexy/netlogo-gui/src/main/swing/FileDialog.java#L61) about it:

    // On Linux:
    // 1) java.awt.FileDialog has a weird bug where it returns OK even
    //       if the user clicks the X on the dialog to cancel it.
    // 2) The AWT dialog looks unbelievably ugly, shows all the hidden files
    //  and folders in my HOME directory (and yes, there are a million),
    //  won't autocomplete names, won't let me start typing names to select files,
    //    and generally hates me as much as I hate it.
    // I suspect my Linux-specific code could replace all the Mac/Windows code below,
    // but that code already seems strangely more complicated than expected, so I decided
    // not to mess with it.  Probably it's just left-over cruft from a pre-JFileChooser era,
    // but I'm not sure.
    //  ~Forrest (6/8/2009)

1) If "returns OK" means that `getFile` returns a file and not `null` than that's no longer true.

2) (a) The GTK file dialog looks quite good, and is what used for all other non-Java applications
    (b) hidden files can be filtered out forcibly. Hidden directories can't be forcibly filtered out and their
         visibility is whatever the user decides, Ctrl+H for toggling in file dialog/manager, the default is to
         hide them.
    (c) It does autocomplete names (JFileChooser doesn't)
    (d) It does let you select files by starting typing

Here is a screenshot of awt FileDialog:

screenshot from 2017-04-03 21-46-01

Here is a screenshot of swing JFileChooser with GTK l&f:

screenshot from 2017-04-03 21-48-37

and with Nimbus l&f (that's what used by NetLogo on Linux, for me anyway) :
screenshot from 2017-04-03 21-50-08

I want to change `org.nlogo.swing.FileDialog.show` to use native awt FileDialog for Linux too, and am opening this issue for discussion before making the change (there's going to be a PR anyway, because I converted the file to Scala and simplified the code - no behavior changes. I want to include this change in the PR if accepted)

Environment details:
Ubuntu 16.10 (I'm using LXDE and openbox as the de and wm, respectively)
No special configuration other than that, every Ubuntu uses gtk file dialog.

Idloj:
I took the responses on #1375 as a green light.

Another (important, I think) feature we get from switching to the native file dialog, which I somehow forgot to mention there, is that you're warned when trying to save over an already-existing file. This actually was my original motive to use the OS dialog instead of JFileChooser.

There are trivial updates to three submodules (ls & vid & gogo). @mrerrormessage, can we do the trick we've done on #1215, where you make a branch on the submodules so I can open a PR against them?


mrerrormessage:
Thanks for the PR! Yeah, I think we'll go ahead and do this. I haven't tested this yet - I'll do that once the submodules are updated. I've pushed branches to each of the submodules you listed with the name warn-dialog-update. Let me know when this (or the submodules) needs further review / pull request.

Idloj:
I opened PRs on the relevant submodules.

mrerrormessage:
Very nice work! I'm glad we have an improved experience for our linux users.

Idloj:
And for the tests to pass I need the PRs on the submodules to be accepted (so I can reference to their warn-dialog-update branch from here)...

mrerrormessage:
All submodule PRs have been approved, can you update the submodules to their appropriate commits and push that change? Once that's done, Travis should pass and I'll merge this, then I'll merge down each of the submodule branches.

PR #1403, opened 2017-05-23
The multiline field should exist for String input box types only, rem…

LaCuneta:
…oving from Numeric types.

This fix is related to the [Tortoise issue 182 about numeric and color inputs causing errors](https://github.com/NetLogo/Tortoise/issues/182).  These changes are not required to make the fix for 182 work, but I think it makes the purpose of multiline field clearer if it only exists on the input box type it is meant to apply to.


mrerrormessage:
This looks good. Thanks!

PR #1076, opened 2017-05-23
Autosuggest expanded primitive list

mrerrormessage:
Autosuggest provides suggestions for all built-in primitives, but it would be a nice enhancement if it also suggested agent variable names and extension primitives.


mrerrormessage:
Note: This only added support for extension primitives, but it should now be relatively easy to get access to built-in agent variables (via the dialect) if someone wants to tackle adding those. Getting access to user-defined procedures and variables is still hard and will require careful consideration, since we don't want to tie the ide too tightly to the workspace.

PR #1390, opened 2017-05-24
Browser for help does not open  using Windows 64 bit version

mrerrormessage:
OS:  Win64Pro x64 version 1703
Netlogo 6.0.1 x64 (only x64 version.  x32 version works correctly)

Action: Clicking  Help> Look up in Dictionary, Help>NetlLogo User Manual, or Help>NetLogo Dictionary
 
Result: fails to open browser or load page into open browser (Chrome, Edge)
The weblinks under Help (NetLogo User Group, Intro to agent based modeling) do work correctly to open a browser.


netlogo-gui/src/main/swing/BrowserLauncher.scala
```diff
@@ -0,0 +1,124 @@
...
+      Files.write(redirectFile, redirectHelper(uriWithAnchor).getBytes("UTF-8"))
+      openURI(comp, redirectFile.toUri)
+    }
+  }
+
+  def redirectHelper(targetUri: URI): String = {
+    s"""|<html>
+        |<head><meta http-equiv="refresh" content="0;url=${targetUri}" /></head>
+        |<body onload="window.location = '${targetUri}'"></body>
+        |</html>""".stripMargin
```
mrerrormessage: @TheBizzle , we have to create a temporary html page to redirect to a `file:` URL with an anchor tag (Win/Linux don't handle them correctly). Does this snippet match your understanding of best-practices for HTML redirects? Should I stick something else in here or remove something? This seems to work fine on FF in Win 10, Mac, and Ubuntu, as well as in IE8 (Windows 7).

TheBizzle: I'm not sure that there is any solution that's going to be pretty here, so I guess it's just a battle to find the least of all the evils.

I feel like it would be mildly cleaner to just drop the `window.location = ...` thing into a `script` tag (which will be immediately run when the page loads), rather than using an `onload` function on an element, but if that doesn't work... :man_shrugging:

mrerrormessage: Yeah, I can pull the JS into a separate script tag. I'm pretty sure that the the `meta` tag is actually doing the redirect in the majority of cases. One SO answer suggested including the `onload` code in case that tag wasn't picked up properly.

And yeah, it's definitely the least of all evils. I considered several different solutions here, and all of them sucked, this one just sucked the least.

PR #1408, opened 2017-05-28
Show error label on nls with compile errors. Fixes #1363

### LINKED ISSUE #1363: Error not displayed when opening model with compile error in includes file
Open a model which includes a compile-time error in one of its included ".nls" files.

nlogo:
```
__includes [ "compile-error.nls" ]
```

nls:
```
to compile-error
  run
end
```

NetLogo will open and switch to the compile tab as appropriate, but the error text (in this case "RUN expected at least 1 input...") will not be displayed at the top of the editor. This was a regression present in 6.0.

Idloj:


mrerrormessage:

PR #1410, opened 2017-05-28
Close temporary code tabs when loading a new model

Idloj:


mrerrormessage:
Awesome! Thanks for the PR!

PR #1412, opened 2017-05-31
Switching tabs compiles nls even after saving it. Fix #1355

### LINKED ISSUE #1355: Switching tabs from an includes file does not recompile it
To reproduce:

1. Open a new file
2. Add an includes file with a single reporter `foo`, that reports `1`.
3. Run `foo` in the command center. It will report 1.
4. Change the 1 to a 2, but don't save.
5. Run `foo` in the command center. It will report 2.
6. Change the 2 to a 3 and save the includes file before switching tabs.
7. Run `foo` in the command center. It will still report 2.

It should output 3 in step 7.

Idloj:


mrerrormessage:
Looks great. Thanks again!

PR #1413, opened 2017-06-01
Switch to Jenkins

mrerrormessage:
* Modify sbt launcher to avoid sbt colors cluttering the build
* Make sbt build friendlier on RHEL
* Add Jenkinsfile


SethTisue:
and the pendulum swings again...

mrerrormessage:
Yep. This change has largely been motivated by increasing build lengths on Travis and their enforcement of a fixed length for build times (this has been especially problematic for Tortoise). But I think there are potentially some benefits for NetLogo desktop as well, especially the potential to automate a number of extension-related tasks.

SethTisue:
makes sense. for Scala org stuff we use Travis when we can and Jenkins when we must

PR #1418, opened 2017-06-12
Make "Show Usage" high enough to match its content. Closes #1141

### LINKED ISSUE #1141: "Show usage" popup on Linux not quite high enough
For example:

untitled

It's only missing a few pixels, but the letters in the second line are truncated and the scroll bar unnecessarily appears.

Idloj:


mrerrormessage:
This is great, thanks for opening this PR!

PR #1420, opened 2017-06-13
Allow selection of individual optimizations

mrerrormessage:


mrerrormessage:
@TheBizzle , I plan to merge this as soon as tests pass. It will require modification of [DockingFixture](https://github.com/NetLogo/Tortoise/blob/master/jvm/src/test/scala/dock/DockingFixture.scala#L61) to include the appropriate optimizations. The optimizations implemented the prior quarter in Tortoise:

* Fd1, FdLessThan1 (encapsulated by `nvm.Optimizations.fdOptimizations`)
* CroFast, CrtFast, HatchFast, SproutFast (encapsulated by `nvm.Optimizations.emptyBlockCreateOptimizations`)
* With (can be added to an optimizations list as `... ++ Seq(nvm.Optimizations.Reporter -> "org.nlogo.compile.middle.optimize.With")`

I believe this will address the problem we discussed yesterday (let me know if there's anything else). something, once I've merged I'll publish the headless (and parserJS) artifacts and send the version to you. 

This also includes a builder implementation for core ASTNodes. While it's still a bit clunky, it's much easier than composing the nodes together and allows for re-use between different test cases. It's essentially a scaled-back copy of a builder I've made for NetLogo astNodes, which I use [in these tests](https://github.com/NetLogo/NetLogo/blob/hexy/netlogo-gui/src/test/compile/NvmTests.scala#L173-L185). The scaling-back is so that there's no dependencies on reflection, which gets tricky when scala-js is involved. If necessary, some of the reflective capabilities could be added back with a stern warning to avoid using them in javascript.

PR #1425, opened 2017-06-16
Errors in optimized withs are reported properly (fix #789)

### LINKED ISSUE #789: `other` confuses error messages thrown when `with` is given a non-predicate reporter
To reproduce:

```
observer> crt 10
turtles> show other turtles with [ color ]
```

You'll get an error like: `OTHER expected a true/false value from (turtle 0), but got 125 instead.`

The error should be: `WITH expected a true/false value from (turtle 0), but got 125 instead.`

I tested other primitives besides `other` (`any` and `one-of`) and they behaved correctly. My guess is that this has something to do with `other ... with ...` optimizations.

mrerrormessage:


mrerrormessage:
Sorry @TheBizzle , I accidentally added you as a reviewer. You're welcome to review if you would like, but I thought you already had left feedback on these commits and was clicking to see what it was. Nope. Just unintuitive GitHub UI.

PR #1436, opened 2017-06-29
Optimize argument lookup for extensions

qiemem:
This signficantly reduces the overhead of getting argument values in extensions. For instance:

6.0.1:
```
observer> let a array:from-list [ 0 ] reset-timer repeat 100000000 [ array:set a 0 0 ] show timer
observer: 17.658
```

This branch:
```
observer> let a array:from-list [ 0 ] reset-timer repeat 100000000 [ array:set a 0 0 ] show timer
observer: 14.87
```

Of course, the amount this matters is dependent on how long the primitive operations themselves take and how many arguments there are.


mrerrormessage:
Looks good. Thanks!

Issue #1346, opened 2017-02-27
Infotab - (un-)ordered lists


daimpi:
I've encountered some confusions around (un-) ordered lists in the documentation and the use of the lists:

1. the documentation seemingly gives two examples of ordered lists in the code section, but the output shown below is for unordered lists: https://ccl.northwestern.edu/netlogo/docs/infotab.html#ordered_lists

2. The link to the section of unordered lists doesn't seem to work (probably b/c the section doesn't exist) and instead throws back to the top of the infotab section : https://ccl.northwestern.edu/netlogo/docs/infotab.html#unordered_lists
 
3. When using lists in the infotab of NetLogo I figured that in order to go one list level deeper the indentation had to be increased by 4 spaces instead of two spaces as indicated in the documentation. Also: the first level seemingly doesn't need any indentation. So while it's true that the 2nd level needs 4 spaces indentation the first one seemingly can have two or none while the third would need 8.

4. The subitems in in ordered lists receive more indentation in the output than unordered lists. Level two in an ordered list seems to be corresponding to level three in an unordered list wrt indentation of the output.

5. The subitems in unordered lists don't receive different symbols for different levels anymore (which they did in 5.3.1).

Here is an example for code in the infotab: 

```
* hi

  * hi
    * hi
    * hi

  1. bla
    1. blu
    2. na
        1. hu

  * a
    * b
        * c

```

5.3.1: 
![grafik](https://cloud.githubusercontent.com/assets/22129968/23356964/1bf48e54-fcdc-11e6-90ff-9cf23d473654.png)

6.0:
![grafik](https://cloud.githubusercontent.com/assets/22129968/23356987/315c4660-fcdc-11e6-9c28-ad317e00b33e.png)

mrerrormessage:
Thanks for reporting:

Issues (1) and (2) will be addressed in https://github.com/NetLogo/models/issues/341 . We have a separate repository for our models library, so I've filed the bug there. 

Issue (3) and (4) are legitimate, but we're limited by Pegdown, our Markdown parsing library. Unfortunately, Pegdown is no longer maintained and has a number of tickets piling up (no doubt for things like inconsistent list indentation). I've created https://github.com/NetLogo/NetLogo/issues/1347 to address the transition issue.

[...]

PR #1347, opened 2017-06-30
Switch from Pegdown to flexmark for Markdown processing

mrerrormessage:
Pegdown is dead (long live Pegdown!). We should move to their suggested replacement, [flexmark-java](https://github.com/vsch/flexmark-java), which is (a) faster, (b) maintained, and (c) more extensible.

This will be a substantial undertaking as we have a decent amount of code which references Pegdown (NetLogo itself, NetLogo build code, the NetLogo documentation sbt plugin). 

As noted on #1346, different indentation levels of unordered lists should ideally have different styles. If this fix isn't handled as part of this ticket, please open a new ticket to track that part of the work.

- [x] Replace Pegdown in NetLogo
- [x] Replace Pegdown in NetLogo build
- [x] Apply distinct styles to differing indent levels of unordered lists


mrerrormessage:
The last commit in this PR also changes list indentation from 4-spaces to 2-spaces (per #1346).

mrerrormessage:
Correction: this *does not* change indent sizes. Doing so had weird ramifications for indented code blocks and would have required substantial changes in the models library. Instead, lists still use 4-space indents.

PR #1448, opened 2017-07-09
Update cf and ls

qiemem:


mrerrormessage:
Addressed in 5039ca7c167e93386cec00f72ff2ef21f318036e

PR #1442, opened 2017-07-10
Current-line editor highlight is too close to "error" color

mrerrormessage:
Should be switched back to same color as in 5.3.1


mrerrormessage:
screen shot 2017-07-10 at 1 06 37 pm

"before" is on the right, "after" is on the left.

fstonedahl:
Personally, I might prefer an even subtler highlight color...

PR #1440, opened 2017-07-17
Improve behaviorspace command-line documentation

mrerrormessage:
https://ccl.northwestern.edu/netlogo/docs/behaviorspace.html#advanced

A user notes that the jar is no longer at NetLogo.jar, it's at app/NetLogo.jar (on Windows and Linux, it's actually in a different location entirely on Mac OS X.

On Linux, we distribute a headless.sh script. It's probably worth distributing a similar script for other platforms so we can unburden users of the responsibility of locating the appropriate jars and configuring everything on their own.

### DIFF
--- a/autogen/docs/behaviorspace.html.mustache
+++ b/autogen/docs/behaviorspace.html.mustache
@@ -394,8 +394,11 @@ windows.-->
       How to use it
     </h4>
     <p>
-      Run Java with the org.nlogo.headless.Main class. The Main.main()
-      method supports these arguments:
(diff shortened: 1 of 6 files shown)


LaCuneta:
@colby-ham Sorry you're having trouble finding a helpful example.  I'm assuming you looked through the [BehaviorSpace docs](http://ccl.northwestern.edu/netlogo/docs/), and specifically the "Advanced Usage" section?  It has a couple of example command lines for running experiments with XML.  Is there a particular part of those docs that you are finding lacking?

PR #1465, opened 2017-07-18
Rework models library cross-referencing and indexing

mrerrormessage:
- Adds typesafe's "Config" as a dependency
- Converts ModelsLibrary, ModelsLibraryDialog to scala
- Uses config to CrossReference when displaying models library
  instead of copying files around
- Sorts out sbt dependencies so tasks related to models
  library packaging are performed in correct order
- Updates models library


SethTisue:
yay HOCON! ❤️

PR #1441, opened 2017-08-02
Editor should stick to end-of-line when navigating whitespace lines

mrerrormessage:
Instead of staying on the current column, when navigating to the a blank line, the editor should move to the rightmost column.

If a user enters a block, then closes that block, then navigates backward, they end up 1 space from the end of the line instead of at the very end of the line.


mrerrormessage:
@qiemem , you may want to give this a spin to see if you approve of the behavior. Changes were inspired by @cbradyatinquire, who identified a weakness in the indenter when opening a block, adding a few newlines, then going back into the block to add content. Should not affect the indenter as such, but only arrow key navigation (and only up/down navigation for that).

I need to have another look at the code before I'm ready to merge it in, but the behavior should be pretty much the same.

mrerrormessage:
Closing for now. Work is archived at tag `archive-caret-follows-line-end`.

PR #1484, opened 2017-08-23
Make scrolling in the Info tab slower. Fixes #1483

### LINKED ISSUE #1483: Smoother scrolling in Info Tab
One of our users has written that scrolling in the info tab appears jerky. If possible, this behavior should be adjusted so that it is smooth (scrolls in smaller increments). This is particularly noticeable when using Mac Trackpad scrolling, which tends to be extremely smooth.

Idloj:


mrerrormessage:
Awesome! This is a nice improvement.

PR #1485, opened 2017-09-05
`export-world` has different `"OUTPUT"` behavior in Headless vs. GUI

mrerrormessage:
In certain models (working on a comprehensive list right now) Headless will put `"OUTPUT"` in the export-world string right before `"PLOTS"`

Ex:
```
[...links...]

"OUTPUT"
"PLOTS"
"Population"
[...plotting...]
```

Whereas GUI will omit the line with the `"OUTPUT"`. This happens on models without any output.

This is a small bug but is causing some issues when testing `export-world` and `import-world` in Galapagos, and I'm not sure why it would be different. GUI will still import the Headless export version with `"OUTPUT"` in it, but the Web `import-world` parser doesn't handle it correctly.

Thanks!


netlogo-gui/src/main/headless/HeadlessWorkspace.scala
```diff
@@ -117,7 +117,9 @@ with org.nlogo.api.ViewSettings {
   /**
    * Has a model been opened in this workspace?
    */
-  var modelOpened = false
+  private[this] var _openModel = Option.empty[Model]
+  def modelOpened = _openModel.nonEmpty
+  def setModelOpened(model: Model): Unit = { _openModel = Some(model) }
```
Idloj: Shouldn't it be `setOpenModel`?

(The comment above is not consistent with the code anymore)

mrerrormessage: That's a better name. I can also rearrange is so that `modelOpened` directly follows the comment.

CamdenClark:
Thanks for tending to this bug!

mrerrormessage:
headless artifact now published - version 6.0.2-7803691

PR #1497, opened 2017-09-18
Show `import-world` errors in a scrollable text area. Fix #196

### LINKED ISSUE #196: import-world error dialog should be scrollable
all too often there are multiple errors and you can only see the beginning because the dialog is of a fixed, small size

Idloj:


mrerrormessage:
Looks good. Thanks!

PR #1507, opened 2017-10-07
Agent monitors enhancement

Idloj:
Before & After (left is After):
image

Other than the look, there are two behavior enhancements:
1. The "Watch" button is a toggle button. When turned off, it resets the perspective (thus cancels the watching).
2. The whole panel can be clicked to expand/collapse, not just the arrow.


mrerrormessage:
Overall, this PR looks really good (tests are passing on Jenkins, even though the status on the PR says otherwise). Testing on Mac and windows has revealed one minor issue which I'm wondering whether you've seen or would have ideas on how to resolve (as always, I'm happy to program the potential fixes on Mac/Windows since you run Linux). The issue is that agent monitors open with the variable display in scroll mode (so not all variables are visible). Here's a before / after on Windows:

screen shot 2017-10-24 at 3 15 23 pm

If this was the only instance, it would probably be fine (after all, you can see part of the last variable). However, the situation on Windows for turtles is worse:

screen shot 2017-10-24 at 3 35 53 pm

As you can see, several variables are totally invisible. On Mac, this problem also occurs:

screen shot 2017-10-24 at 3 40 24 pm

Obviously, it's totally fine if the scroll bar appears after the window is resized, but the window should open showing all variables without it, if possible. Have you seen this problem? If not, what file(s) should I edit to fix it?

Thanks again for working on this, it's a big improvement to agent inspector windows. I'll want to go through and re-screenshot tutorial shots of the inspector windows at some point when this is merged (I'm leaving this as a note to myself, you don't need to do that for the PR to be merged).

Idloj:
I haven't seen this problem on Linux, and looking at `CollapsiblePane`'s code, I don't know why it happens.

mrerrormessage:
OK, I've fixed the problem on Mac (haven't yet tested on Windows, but I suspect it's also fine there) with the following change:

```diff
diff --git a/netlogo-gui/src/main/app/tools/AgentMonitor.scala b/netlogo-gui/src/main/app/tools/AgentMonitor.scala
index 27b9b70..4f49621 100644
--- a/netlogo-gui/src/main/app/tools/AgentMonitor.scala
+++ b/netlogo-gui/src/main/app/tools/AgentMonitor.scala
@@ -2,9 +2,9 @@

 package org.nlogo.app.tools

-import java.awt.{ Dimension, GridBagConstraints, GridBagLayout, Insets }
+import java.awt.{ BorderLayout, Dimension, GridBagConstraints, GridBagLayout, Insets }
 import java.util.{ List => JList }
-import javax.swing.{ BoxLayout, JPanel, JScrollPane, JWindow, ScrollPaneConstants }
+import javax.swing.{ JPanel, JScrollPane, JWindow, ScrollPaneConstants }

 import org.nlogo.agent.Agent
 import org.nlogo.app.common.{ CommandLine, HistoryPrompt, LinePrompt }
@@ -48,7 +48,7 @@ extends JPanel with CommandCenterInterface // lets us embed CommandLine
   private val historyPrompt = new HistoryPrompt(commandLine)
   private val agentEditor = new AgentMonitorEditor(this)

-  setLayout(new BoxLayout(this, BoxLayout.PAGE_AXIS))
+  setLayout(new BorderLayout)

   val scrollPane = new JScrollPane(agentEditor,
     ScrollPaneConstants.VERTICAL_SCROLLBAR_AS_NEEDED,
@@ -62,14 +62,14 @@ extends JPanel with CommandCenterInterface // lets us embed CommandLine
   val viewPanel: Option[AgentMonitorViewPanel] =
     if(agentKind == AgentKind.Observer) {
       // the observer monitor doesn't have a view or the command center. ev 6/4/08
-      add(scrollPane)
+      add(scrollPane, BorderLayout.CENTER)
       None
     }
     else {
       implicit val i18nPrefix = I18N.Prefix("tools.agentMonitor")
       val panel = new AgentMonitorViewPanel(workspace)
-      add(new CollapsiblePane(I18N.gui("view"), panel, window))
-      add(new CollapsiblePane(I18N.gui("properties"), scrollPane, window))
+      add(new CollapsiblePane(I18N.gui("view"), panel, window), BorderLayout.NORTH)
+      add(new CollapsiblePane(I18N.gui("properties"), scrollPane, window), BorderLayout.CENTER)
       commandLine.setEnabled(agent != null && agent.id != -1)
       historyPrompt.setEnabled(agent != null && agent.id != -1)
       commandLine.agentKind(agentKind)
@@ -90,7 +90,7 @@ extends JPanel with CommandCenterInterface // lets us embed CommandLine
       c.insets = new Insets(1, 1, 1, 1)
       gridBag.setConstraints(historyPrompt, c)
       commandPanel.add(historyPrompt)
-      add(commandPanel)
+      add(commandPanel, BorderLayout.SOUTH)
       Some(panel)
     }

```

Was there a reason you were switched from using `BorderLayout` to `BoxLayout` that would prevent this from working? Can you verify that this change works for you?

Idloj:
I've applied your changes, and I don't see any problem -- I can't remember why I switched from `BorderLayout` to `BoxLayout`.

mrerrormessage:
Looks good. Thanks!

PR #1509, opened 2017-10-24
Align trail drawing with NLW while also drawing back pen trails

mrerrormessage:


netlogo-core/src/main/agent/Turtle.java
```diff
@@ -871,34 +871,34 @@ public void ycor(Double ycor)
...
+
     Patch originalPatch = getPatchHere();
 
     double newX = _world.wrapX(xcor);
     double newY = _world.wrapY(ycor);
 
-    drawLine(this.xcor, this.ycor, xcor, ycor);
+    if (! isJump) {
+      drawLine(this.xcor, this.ycor, xcor, ycor);
+    }
```
LaCuneta: If I'm reading this right, this skips drawing if we're jumping, but my understanding is that we're supposed to draw a line whenever we move (per the dictionary).  If that's true, I don't think we need any of the `isJump`-specific logic.

mrerrormessage: The arrangement of things here is a bit strange, but we do still draw a line even when jumping. Whereas this method uses [drawLine with start and dest](https://github.com/NetLogo/NetLogo/pull/1509/files#diff-e44dd5e0b96bd9fd2b33db4c98d47d26R110) to draw the line, jumping uses [drawLine with start and distance](https://github.com/NetLogo/NetLogo/pull/1509/files#diff-e44dd5e0b96bd9fd2b33db4c98d47d26R94). We need the separate `drawLine` methods for exact reproducibility as `PenLineMaker` needs to know distance and heading to do wrapping appropriately. Distance and heading are, unfortunately unavailable in the `xandycorHelper` method.

LaCuneta: Thank you for the explanation, I think I'm starting to understand.  One more question, is [the `isJump` passed in here used](https://github.com/NetLogo/NetLogo/pull/1509/files#diff-5c1f0f72bba41ac7187d99ade4e7e672R879)? I don't see that value actually getting passed to the `xandycorHelper`, and I'm not clear if it's supposed to be ignored.

mrerrormessage: Nice catch! That method is called by `TieManager`, so the `isJump` parameter should probably be removed because tie-based movement is never considered a jump.

LaCuneta: I've run the drawing tests in Tortoise against this branch and we are aligned, no errors!  Thank you again for this work.

mrerrormessage:
Published under `0.2.0-38544d4` (parserJS/JVM) and `6.0.2-38544d4` (headless).

PR #1515, opened 2017-11-20
Disconnected flows and buggy behavior in System Dynamics Modeler

mrerrormessage:
Greetings, several students in my class have reported strange behaviors while using the System Dynamics Modeler using NetLogo 6.0.2 (although similar problems were observed last year on version 5.3.1 as well).

During the course of creating a SD model, some of the flows seem to "disconnect" from the attached stocks and are not computed in the simulation. If the user re-attaches the flows to the stocks, they are _still_ not considered in the simulation. The flows must be deleted and re-created to fix the problem.

I have attached two sample files here exhibiting the behavior and one file with expected results. While I can verify the problems with these files, I cannot reproduce the steps to break a SD model in this way which leads me to believe it is a bug in the GUI related to redefining flow variables as a model takes shape.

[examples.zip](https://github.com/NetLogo/NetLogo/files/1476971/examples.zip)

### DIFF
--- a/netlogo-gui/src/main/sdm/gui/AggregateDrawing.java
+++ b/netlogo-gui/src/main/sdm/gui/AggregateDrawing.java
@@ -17,12 +17,13 @@ public strictfp class AggregateDrawing
   }
 
   public void synchronizeModel() {
+    model.elements().clear();
     FigureEnumeration figs = figures();
     while (figs.hasNextFigure()) {
       Figure fig = figs.nextFigure();
       if (fig instanceof ModelElementFigure &&
           ((ModelElementFigure) fig).getModelElement() != null) {
-        getModel().addElement(((ModelElementFigure) fig).getModelElement());
+        model.addElement(((ModelElementFigure) fig).getModelElement());
           }
     }
   }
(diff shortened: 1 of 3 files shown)


mrerrormessage:
@ptgrogan. Thank you so much for reporting this issue. This fix will be merged once tests pass and should be included in the next release of NetLogo.

The problem was that we weren't reloading certain model elements before recompiling based on them. It seems that this didn't affect certain types of changes (like text edits) while it did affect others (like rate connections). This fix rebuilds the model based on what is shown in the system dynamics editor. The upshot is that rates that before were unable to make effective connections between stocks will now be updated when the system dynamics model is recompiled. So in the attached "sample1" model, the "Reworkcompletion" and "Workcompletion" flows seem to be going between Reworktodo->Workdone and Worktodo->Workdone respectively, but internally they were represented as going from Workdone->Workdone (and thus making no difference). Even worse, changing them in the GUI didn't affect the code generated by the model! After making this change, the model still opened these connections incorrectly as Workdone->Workdone *however* it is now possible to edit these connections to properly reflect the model, which should fix the problem of flows "disconnecting."

PS. It appears that github goofed on this and put my name on it when I opened a pull request for this. I don't know exactly why that happened, but thanks again for submitting this issue and please let us know if there are problems you run into in the future.

ptgrogan:
Thanks for your quick response and I am glad to hear you were able to track down the problem! Your technical description seems to match the behavior I was observing.

PR #1541, opened 2018-01-22
Make repeat, while, and loop scope at runtime

### LINKED ISSUE #1540: Anonymous commands created inside a new scope in a repeat share scope they shouldn't
Example:

```
set lambdas []
repeat 5 [
  let bar 0
  ask patch 0 0 [
    set lambdas lput [ -> set bar bar + 1 set glob1 bar ] lambdas
  ]
]
foreach lambdas [ l -> run l ] show glob1
```

where `lambdas` and `glob1` are globals. This code shows `5` when it should show `1`. 

This is highly related to this test: https://github.com/NetLogo/NetLogo/blob/520e9a7d28c8646b720d62e860fafb8060da75e6/test/reporters/CommandLambda.txt#L295-L299

qiemem:
Fixes #1540 

Note that this does result in a minor performance hit to these primitives.

### DIFF
--- a/netlogo-gui/src/main/prim/etc/_scope.scala
+++ b/netlogo-gui/src/main/prim/etc/_scope.scala
@@ -1,0 +1,17 @@
+package org.nlogo.prim.etc
+
+import org.nlogo.nvm.{Command, Context}
+
+class _enterscope extends Command {
+  override def perform(context: Context): Unit = {
+    context.activation.binding = context.activation.binding.enterScope
+    context.ip += 1
+  }
+}
+
+class _exitscope extends Command {
+  override def perform(context: Context): Unit = {
+    context.activation.binding = context.activation.binding.exitScope
+    context.ip += 1
+  }
+}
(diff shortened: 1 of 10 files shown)


SethTisue:
is there some fundamental reason why the compiler can't sort this out at compile time? the performance implications here seem alarming to me.

netlogo-gui/src/main/prim/etc/_scope.scala
```diff
@@ -0,0 +1,17 @@
+package org.nlogo.prim.etc
+
+import org.nlogo.nvm.{Command, Context}
+
+class _enterscope extends Command {
```
mrerrormessage: This method should have a `perform_1` so it's eligible for inlining.

netlogo-gui/src/main/prim/etc/_scope.scala
```diff
@@ -0,0 +1,17 @@
...
+  override def perform(context: Context): Unit = {
+    context.activation.binding = context.activation.binding.enterScope
+    context.ip += 1
+  }
+}
+
+class _exitscope extends Command {
```
mrerrormessage: This method should also have a `perform_1`

netlogo-gui/src/main/prim/etc/_scope.scala
mrerrormessage: This prim should have a `perform_1`

mrerrormessage:
Superseded by #1581 .

Issue #92, opened 2012-02-24
NetLogo crashed while HubNet clients are logging in...?


davidatnie:
Hello, 

We use NetLogo in a school in Singapore (particularly HubNet), and I think we may have run into a bug...
So we used a modified Disease model in a class of about 20 students. Just shortly after the activity was started (when roughly 3/4 of the students have logged in), it suddenly crashed, giving the error message below. 

We closed and then restarted the host, and asked the students to log back in, but again after about 3/4 of them have logged in, it crashed again (we had this for 4 times before switching to another model). The error messages are as follows:

Crash 1

```
error (NullPointerException)
 while observer running ASK
  called by procedure LISTEN-CLIENTS
  called by procedure GO
  called by Button 'GO'

NetLogo is unable to supply you with more details about this error.  Please report the problem
to bugs@ccl.northwestern.edu, or at https://github.com/NetLogo/NetLogo/issues, and paste the
contents of this window into your report.

java.lang.NullPointerException
 at org.nlogo.prim._asm_procedurelistenclients_equal_8.report(:1)
 at org.nlogo.nvm.Context.evaluateReporter(Context.java:228)
 at org.nlogo.prim._asm_procedurelistenclients_ask_7.perform(:2)
 at org.nlogo.nvm.Context.stepConcurrent(Context.java:91)
 at org.nlogo.nvm.ConcurrentJob.step(ConcurrentJob.java:82)
 at org.nlogo.job.JobThread.org$nlogo$job$JobThread$$runPrimaryJobs(JobThread.scala:143)
 at org.nlogo.job.JobThread$$anonfun$run$1.apply$mcV$sp(JobThread.scala:78)
 at org.nlogo.job.JobThread$$anonfun$run$1.apply(JobThread.scala:76)
 at org.nlogo.job.JobThread$$anonfun$run$1.apply(JobThread.scala:76)
 at scala.util.control.Exception$Catch.apply(Exception.scala:88)
 at org.nlogo.util.Exceptions$.handling(Exceptions.scala:41)
 at org.nlogo.job.JobThread.run(JobThread.scala:75)

NetLogo 5.0RC10
main: org.nlogo.app.AppFrame
thread: JobThread
Java HotSpot(TM) 64-Bit Server VM 1.6.0_29 (Apple Inc.; 1.6.0_29-b11-402-10M3527)
operating system: Mac OS X 10.6.8 (x86_64 processor)
Scala version 2.9.1.final
[...]

PR #1543, opened 2018-01-25
Fix HubNet NullPointerException

mrerrormessage:
Fixes the infamous #92 . 

The main fix here was to the server, but I spent a substantial amount of time investigating the "double login" bug that afflicted the client and often seemed to lead to this bug. The behavior to create this was to have a client login, kick that client, and have them login again. Their first login would fail, their second would succeed. This odd client behavior correlated strongly with the `NullPointerException` and was a reflection of the client being out-of-state.

A basic sketch of several of the issues contributing to this problem:

* Client were being added to the server's client map when they had a `clientId` of `null` (`EnterMessage` sent or received out-of-order).
* Client enter, activity, and exit messages were processed by the server regardless of whether the client had sent a handshake.
* Calls to `disconnect` would fail for clients which had not completed a client handshake, since they were routed through the connection manager, which looked up clients based on client id.
* Client disconnections for clients with `clientId = null` would generate a hubnet exit message with a `null` `hubnet-message-source`.
* If a client sent multiple handshakes, all of them were processed as enter messages. This lead to the following problems:
  * "ghost" clients appearing in the HubNet control center
  * The HubNet server retained `ServerSideConnection`s after they had been closed because it had them in the client map under two distinct `clientId`s
* The double login bug was fixed by making the `connected` variable (which is accessed at roughly the same time by both the UI thread and the `AbstractConnection` read thread) atomic so as to prevent repeated calls to `ClientApp.handleDisconnect` with `connected == true`. When called with `connected == true`, `ClientApp.handleDisconnect` performed *two* logins.

As is often the case when changing HubNet code, I feel there is a decent chance I have introduced two or three new bugs (although these will likely be rarer and less severe than #92). I am taking the time to write this up in hopes that when those bugs manifest, they can be fixed in such a way that the bug fixed here does not re-emerge.

### DIFF
--- a/netlogo-gui/src/main/hubnet/client/ClientApp.scala
+++ b/netlogo-gui/src/main/hubnet/client/ClientApp.scala
@@ -119,7 +119,9 @@ class ClientApp extends JFrame("HubNet") with ErrorHandler with ClientAppInterfa
     }
   }
 
-  def completeLogin() { setVisible(true) }
+  def completeLogin() {
+    setVisible(true)
+  }
 
   private def login(userid: String, hostip: String, port: Int) {
     var exs: Option[String] = None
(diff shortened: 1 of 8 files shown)


SethTisue:
😲🎉

PR #1547, opened 2018-02-02
Add nlogox file format

mrerrormessage:
It's been a long road getting here. See #849 for some background on this task. Documentation on large parts of this PR are also available [here on the wiki](https://github.com/NetLogo/NetLogo/wiki/The-nlogox-format). Summary of a few basic achievements of this branch (aside from the headlining file format addition), in no particular order:

* Remove `comeUpForAir` flag on `World`
* Consolidate `AbstractWorkspace` and `AbstractWorkspaceScala` into `AbstractWorkspace`
* Rationalize workspace initialization and dependencies
* Change `ExtensionManager` implementation (note: the API here should be the same, but you won't find `workspace` on extension manager any more)
* Fix 3D inspector selection on HiDPI displays (untracked bug)
* Adds ability to switch (albeit somewhat awkwardly) between NetLogo 2D and NetLogo 3D.

Some notes to various people regarding getting this ready for release to users:

* @LaCuneta , the code generator we have here should work on Javascript, we will just need to write Javascript wrapper code equivalent to [our JVM xml adapter](https://github.com/NetLogo/xml-format-library/blob/master/xml-lib/jvm/src/main/scala/ScalaXmlAdapter.scala).
* @qiemem , I've resaved the models in various extensions and don't think most extensions will require major attention. However, it would probably be best for you to look over the changes to `ls` and take it for a spin when possible.
* @bainco , For now I've left the .nlogo/.nlogo3d models in the models library alongside .nlogox, just as a reference. I have tools to convert quickly between the two and plan to improve that tooling even further before release.
* @bainco , @TheBizzle this branch includes the [ModelInfo](https://github.com/NetLogo/NetLogo/blob/wip-add-nlogox/parser-core/src/main/core/ModelInfo.scala) metadata we've talked about in the past.
* To anyone who is interested in reviewing this (or even a portion of it) I would be happy to provide a brief walkthrough of various sections or changesets and discuss what the most effective ways to review are.
* To anyone who is interested in testing this, you would have my undying gratitude. I have tried to test things thoroughly, but NetLogo does a lot, so it's sometimes hard to get everything in one go.

Just so everyone is aware, my anticipated timeline for merging this is the afternoon of February 6 or the morning of February 7, assuming all tests are passing and no major defects are discovered. I'm going to take time this afternoon to get any failing tests passing as well as to condense the commit history (although I expect to make no major code changes). If you would prefer a delay in merging or have reservations, please comment on this PR and the timeline can be adjusted. Otherwise, problems discovered will be fixed after this is merged into hexy 😁  .


SethTisue:
cool! a decade-old dream!

arthurhjorth:
Quick (or maybe not!) question: what do the changes to `workspace` mean? is it private or hidden now, so that you can't access it from an extension?

mrerrormessage:
@arthurhjorth  - good questions. Here are the basic bullet points of what has changed:

* `org.nlogo.api.ExtensionManager` (an interface) is the same as it was before, with the addition of a single method: `activeModel: org.nlogo.core.Model`. `ClassManager`s which relied only on [the methods found on `org.nlogo.api.ExtensionManager`](https://ccl.northwestern.edu/netlogo/docs/scaladoc/org/nlogo/api/ExtensionManager.html) will not need to change.
* Similarly, [`org.nlogo.api.Context`](https://ccl.northwestern.edu/netlogo/docs/scaladoc/org/nlogo/api/Context.html) has not changed. Primitives which relied on the [`org.nlogo.api.Workspace`](https://ccl.northwestern.edu/netlogo/docs/scaladoc/org/nlogo/api/Workspace.html) obtained through the context should not need to change.
* Regrettably, many extensions take `org.nlogo.api.ExtensionManager` and cast it to [`org.nlogo.workspace.ExtensionManager`](https://ccl.northwestern.edu/netlogo/docs/scaladoc/org/nlogo/workspace/ExtensionManager.html). Those extensions are liable to break as `workspace.ExtensionManager` no longer has a `workspace` method. In certain cases, this `workspace` was saved in primitives. The preferred approach for primitives which need workspace access should *always* be to obtain the workspace through the context.
* Also, many extensions take the `org.nlogo.api.Workspace` obtained through `context` and cast it to a `GUIWorkspace` or `AbstractWorkspace` or `AbstractWorkspaceScala`. These extensions may or may not break depending on what they cast to and which methods they called. Extensions which cast a workspace to `AbstractWorkspaceScala` will definitely break (the class doesn't exist any more), extensions which called methods on `AbstractWorkspace` or `GUIWorkspace` may break if those methods were removed or their signatures changed.

A more thorough treatment of compatibility differences can be found [on the wiki](https://github.com/NetLogo/NetLogo/wiki/6.0-Extension-and-Controlling-API-Transition-Guide#addendum-proposed-api-and-non-api-changes-after-602) (note that I'm going to continue to update and refine this section of the wiki).

arthurhjorth:
Thanks for clarifying, @mrerrormessage!

Yeah, wrapping my head around how many Workspace classes there are, and casting between them to be able to get the right stuff was super confusing when I first got into building extensions.  It is great that it is being cleaned up. I think I get it, but I might ask more questions next time I start building an extension. Thanks either way :)

mrerrormessage:
This branch is being put on the blocks for now. The tag lives at `archive-add-nlogox` if anyone wants to update it or take it over at some point in the future.

Issue #1550, opened 2018-02-07
pcolor LogoList cast error on clear-all in NetLogo 3D 6.0.2


gdower:
Hello,

I'm trying to convert a model from NetLogo3D 5.3.1 to 6.0.2 and I get the following Java exception when running clear-all after the model has been setup more than once:

```
error (ClassCastException)
 while observer running CLEAR-ALL
  called by Command Center

NetLogo is unable to supply you with more details about this error.  Please report the problem
 at https://github.com/NetLogo/NetLogo/issues, or to bugs@ccl.northwestern.edu, and paste the
contents of this window into your report.

java.lang.ClassCastException: org.nlogo.core.LogoList cannot be cast to java.lang.Double
 at org.nlogo.agent.Patch3D.pcolorDoubleUnchecked(Patch3D.java:243)
 at org.nlogo.agent.AgentManagement.clearPatches(AgentManagement.scala:133)
 at org.nlogo.agent.AgentManagement.clearPatches$(AgentManagement.scala:129)
 at org.nlogo.agent.World.clearPatches(World.scala:109)
 at org.nlogo.agent.AgentManagement.clearAll(AgentManagement.scala:80)
 at org.nlogo.agent.AgentManagement.clearAll$(AgentManagement.scala:78)
 at org.nlogo.agent.World3D.clearAll(World3D.scala:362)
 at org.nlogo.workspace.AbstractWorkspaceScala.clearAll(AbstractWorkspaceScala.scala:74)
 at org.nlogo.window.GUIWorkspace.clearAll(GUIWorkspace.java:240)
 at org.nlogo.prim.etc._clearall.perform(_clearall.scala:12)
 at org.nlogo.nvm.Context.stepConcurrent(Context.java:107)
 at org.nlogo.nvm.ConcurrentJob.step(ConcurrentJob.scala:65)
 at org.nlogo.job.JobThread.runPrimaryJobs(JobThread.scala:133)
 at org.nlogo.job.JobThread.$anonfun$run$1(JobThread.scala:68)
 at scala.runtime.java8.JFunction0$mcV$sp.apply(JFunction0$mcV$sp.java:12)
 at scala.util.control.Exception$Catch.apply(Exception.scala:224)
 at org.nlogo.api.Exceptions$.handling(Exceptions.scala:41)
 at org.nlogo.job.JobThread.run(JobThread.scala:66)

NetLogo 3D 6.0.2
main: org.nlogo.app.AppFrame
thread: JobThread
[...]

PR #1557, opened 2018-02-10
Fixes the pcolor RGB list cast to double bug

gdower:
Fixes the bug described in issue #1550 and includes unit testing


mrerrormessage:
This looks awesome! Thanks so much!

PR #1558, opened 2018-02-11
Make diffusion symmetric and faster

qiemem:
As floating point summation is commutative but not associative, our diffusion primitive are not symmetric. These tests demonstrate the problem:

https://github.com/NetLogo/NetLogo/blob/edb0ae5934a91216d88549a25e3d45c64e269bc4/test/commands/Diffuse.txt#L62-L108

While this is often inconsequential, in, for instance, diffusion reaction models, these small errors can be quickly magnified and completely change the behavior of a system.

The manner of implementation also dramatically reduces the amount of duplicated code between topologies (though there is still definitely some) and results in a measurable performance improvement . Both of these accomplished by having a single function in `Topology` which performs the diffusion for non-border patches, which means we don't have to constantly mod values in tight loops.


netlogo-core/src/main/agent/Box.scala
```diff
@@ -92,224 +93,33 @@ extends Topology(world2d, xWraps = false, yWraps = false) {
...
-      x = 0
-      while (x < xx) {
-        scratch(x)(y) =
-          world.fastGetPatchAt(x + minx, y + miny)
-            .getPatchVariable(vn)
-            .asInstanceOf[java.lang.Double].doubleValue
-        scratch2(x)(y) = 0
-        x += 1
+  override protected def diffuseCorners
+  (amount: Double, vn: Int, fourWay: Boolean, scratch: Array[Array[Double]]): Unit = {
```
mrerrormessage: I'm not sure what idiomatic scala style is here, but the prevailing convention I have used in NetLogo always puts the open-paren immediately following the method name. Do you think this is better stylistically, or was this simply a typo?

SethTisue:
nice catch — we should have tested for this years ago

mrerrormessage:
This looks good, just needs to be squashed down to a single commit. If you have time, great; otherwise let me know and I'll take care of it.

PR #1559, opened 2018-02-12
Upgrade to Scala 2.12.4

mrerrormessage:
Final Benchmarking will be done tonight to confirm this meets or exceeds the speed of Java 8 + scala 2.12.2. A brief rundown of some of the changes:

* `-XX:+UseParallelGC` - this was needed because the default GC changed in Java 9 to G1GC which prioritizes latency over throughput. Shoutout to the sbt folks for [discussing this issue](https://github.com/sbt/sbt/issues/3559), which put me on the track for this fix.
* For `def foo(): Unit`, Scala 2.12.4 shows the error "Eta-expansion of zero-argument method values is deprecated" for things like `(1 to 9).foreach(foo _)`, so I converted the instances of this to `(1 to 9).foreach { () => foo() }`. Yes, this seems a little weird, but I kind of see where they're coming from.
* Tighten up `ExclusiveJob` and `Context` code a bit.
* Make a bunch of `AgentSet` stuff `private[this]`. The reason for that is calls to non-`private[this]` Scala `val`s and `var`s are actually much more expensive than a single field access. "But what about the JIT?" - prior to these changes, various methods in `ArrayAgentSet$Iterator` actually called up the parent object (`$outer` in scala's terminology) and called a method on it. This meant that something which looked like a simple method call was actually 2. On the BZ benchmark, this lead to a ~5% speedup (prior to `UseParallelGC`) and I initially thought it was enough to get over the perceived performance loss between J8/S2.12.2 and J9/S2.12.4. 
* EDIT: This change was removed because it broke RNG on TabYagHyb and benchmarks showed it wasn't necessary. ~~Introduce `SingleAgentSet`. This was done for a 4% speedup on CA1D (prior to `UseParallelGC`). This primarily affects the performance of `ask`s done on a single agent, but may incidentally improve performance for various one-element agentsets created elsewhere. This change has introduced benchmark differences in the Tabonuco Yagrumo Hybrid model (I haven't compared the nature of the changes, but I believe them to be RNG-related), so we should talk about whether this change gets cut or not on that basis.~~

UPDATE: Benchmarks were run last night on the first 4 changes + scala 2.12.4 + Java 9 (w/`-X:+UseParallelGC`) and all models were faster or within 2% (the only models which measured slower were CA1D, by <1%, and ImportWorld, by 2%). Some models were significantly faster (Termites: 18% faster, GasLabOld: 12% faster, Ising: 11% faster, ANN: 10% faster).

UPDATE 2: As a note, prior tests were carried out using the Java 9 JVM, but the code was compiled using Java 8. Attempting to upgrade the whole project to Java 9 seemed like it would be promising, but ends up being blocked by [this JOGL bug](https://jogamp.org/bugzilla/show_bug.cgi?id=1317) (see also [this forum thread](http://forum.jogamp.org/Jogl-Jogamp-on-Java-9-td4038012i100.html)) which will fix a total JVM crash when an OpenGL window is closed (mac only, I believe). Once this bug is fixed (hopefully in [JOGL 2.3.3](https://jogamp.org/wiki/index.php/SW_Tracking_Report_Objectives_for_the_release_2.3.3)) we can add those changes and have a fully Java 9 project. For the time being, these fixes will get merged into the "ising" branch, which will be a long-running branch targeting some version of NetLogo past 6.0.3. This branch will also include nlogox and various other fixes which we don't typically put in patch-level releases.


SethTisue:
very exciting!

SethTisue:
the eta-expansion thing doesn't look right to me. I'm looking into it

SethTisue:
@mrerrormessage the eta-expansion thing is a bug and should be fixed in 2.12.5 and 2.13.0-M4 by https://github.com/scala/scala/pull/6177

mrerrormessage:
Thanks for the heads-up @SethTisue . Is there a scheduled release date for 2.12.5?

SethTisue:
https://github.com/scala/scala/milestone/68 says March 2. we might hit that, or it might slip a little

EwoutH:
Do you guys want to give an update to Java 13 and Scala 2.13 a try?

Issue #1369, opened 2017-03-27
improve BehaviorSpace dialog location


alan-isaac:
The location of the BehaviorSpace dialog box is apparently not tied to the location of the NetLogo GUI.  It would be helpful if the dialog box were to initially open on top of the GUI.

mrerrormessage:
Perfect fix for a new contributor. [This SO Post](https://stackoverflow.com/questions/1677318/java-swing-positioning-dialog-on-top-of-existing-window) covers how to do this in Swing. The fix should be able to be applied to [LabManager.show](https://github.com/NetLogo/NetLogo/blob/hexy/netlogo-gui/src/main/lab/gui/LabManager.scala#L39).

svr8:
I was trying to do the fix. I'm not noticing any significant GUI changes, am I choosing the right parent object ? I edited the line
https://github.com/NetLogo/NetLogo/blob/0f797196eadb86a27fd0f2f5b4c93b273cc73133/netlogo-gui/src/main/lab/gui/LabManager.scala#L39 to this:

def show() { 
dialog.update(); 
dialog.setVisible(true); 
dialog.setLocationRelativeTo(workspace.viewWidget)
}

mrerrormessage:
The `LabManager` dialog is opened when choosing "Behaviorspace" from the "Tools" menu. The location should be set relative to the `Frame` contained in the workspace.

Idloj:
We also have [this function](https://github.com/NetLogo/NetLogo/blob/hexy/netlogo-gui/src/main/awt/Positioning.scala#L11) which is used in several locations.

svr8:
Okay! I think adding one more statement should do the fix ?
Previous:
  ##https://github.com/NetLogo/NetLogo/blob/hexy/netlogo-gui/src/main/lab/gui/LabManager.scala#L39
New:
`def show() { 
     dialog.update(); 
     dialog.setVisible(true);
     dialog.setLocationRelativeTo(workspace.getFrame)
}`

mrerrormessage:
If you have solved the problem, test your work and open a pull request for review.

mrerrormessage:
Fixed in #1584

PR #1564, opened 2018-02-15
Set location of BehaviorSpace dialog box to centre

svr8:
BehaviorSpace dialog box to center is now bound to center of NetLogo Window.
Fix for issue: improve BehaviorSpace dialog location #1369

### DIFF
--- a/build.sbt
+++ b/build.sbt
@@ -125,8 +125,8 @@ lazy val netlogo = project.in(file("netlogo-gui")).
            Depend.dependTask: _*).
   settings(
     name := "NetLogo",
-    version := "6.0.3-M3",
-    isSnapshot := false,
+    version := "6.0.3-M4",
+    isSnapshot := true,
     mainClass in Compile := Some("org.nlogo.app.App"),
     modelsDirectory := baseDirectory.value.getParentFile / "models",
     extensionRoot   := (baseDirectory.value.getParentFile / "extensions").getAbsoluteFile,
@@ -194,8 +194,8 @@ lazy val headless = (project in file ("netlogo-headless")).
   settings(ChecksumsAndPreviews.settings: _*).
   settings(
     name          := "NetLogoHeadless",
-    version       := "6.0.3-M3",
-    isSnapshot    := false,
+    version       := "6.0.3-M4",
+    isSnapshot    := true,
     autogenRoot   := (baseDirectory.value.getParentFile / "autogen").getAbsoluteFile,
     extensionRoot := baseDirectory.value.getParentFile / "extensions",
     mainClass in Compile         := Some("org.nlogo.headless.Main"),


netlogo-gui/src/main/lab/gui/LabManager.scala
```diff
@@ -36,7 +36,7 @@ class LabManager(val workspace:        GUIWorkspace,
     protocols.clear()
   }
   private lazy val dialog = new ManagerDialog(this, dialogFactory, menuFactory)
-  def show() { dialog.update(); dialog.setVisible(true) }
+  def show() { dialog.update(); dialog.setVisible(true); dialog.setLocationRelativeTo(workspace.getFrame) }
```
mrerrormessage: Please split this method out into multiple lines, as it's getting a bit too long for one line. When splitting into multiple lines, you can also remove the semicolons.

netlogo-gui/src/main/lab/gui/LabManager.scala
```diff
@@ -36,7 +39,20 @@ class LabManager(val workspace:        GUIWorkspace,
...
   }
   private lazy val dialog = new ManagerDialog(this, dialogFactory, menuFactory)
-  def show() { dialog.update(); dialog.setVisible(true) }
+  def updateDialogCoordinates(): Unit = {
+    //If parent window lies in the same place, do not update coordinates
+    val parentLocation = workspace.getFrame.getLocationOnScreen
+    if(currentParentLocation!=null && currentParentLocation.x==parentLocation.x && parentLocation.y==parentLocation.y)
+      return
+    //Set new coordinates
+    currentParentLocation = parentLocation
```
mrerrormessage: If I comment out L43-L48, the dialog appears in the right place without "flickering." Can you test whether commenting them out removes the flicker for you? If it does, let's consider them unnecessary and remove them.

svr8: Ohkay, right! L49 was sufficient. I've removed the function, and added the needed code to show() function

netlogo-gui/src/main/lab/gui/LabManager.scala
```diff
@@ -27,7 +28,7 @@ class LabManager(val workspace:        GUIWorkspace,
 
   def getComponent: Seq[LabProtocol] = protocols.toSeq
   def defaultComponent: Seq[LabProtocol] = Seq()
-
+
```
mrerrormessage: Remove the whitespace on the end of this line.

netlogo-gui/src/main/lab/gui/LabManager.scala
```diff
@@ -36,7 +37,11 @@ class LabManager(val workspace:        GUIWorkspace,
     protocols.clear()
   }
   private lazy val dialog = new ManagerDialog(this, dialogFactory, menuFactory)
-  def show() { dialog.update(); dialog.setVisible(true) }
+  def show() { 
+    Positioning.center(dialog, workspace.getFrame) //Set dialog position to center of main window
```
mrerrormessage: This comment isn't necessary and should be removed. We want our comments to talk about *why* code is doing what it's doing, not just to describe behavior that is already clear from the code itself ([see this CodingHorror blog post](https://blog.codinghorror.com/code-tells-you-how-comments-tell-you-why/)). Here, notice that this comment (which is a good description of what is going on) has most of the the same words as the line of code. This is a good sign! It means the code indicates clearly what it is doing without needing comments. If we were doing something unusual here, we might want a comment explaining why we were doing that, but in this case it seems obvious that we want to center the window because that offers a better experience for our users.

mrerrormessage:
I see you have rebased, but in the process you've rewritten existing history, so this branch can't be merged in as-is. Before final approval, your changes need to appear on this pull-request page as a single commit (the current PR shows 15 commits). To do that it may be easiest to simply copy the work you have onto a new branch (use `git checkout -b backup-branch-name` to do this) change your `hexy` branch to match NetLogo:hexy exactly, then apply your changes to the new branch that matches NetLogo:hexy and add a single commit. To learn more about git, I recommend [gitready](http://gitready.com/) and [the git manual](https://git-scm.com/). I'm happy to provide some guidance on how to do this, but you should use this as an opportunity to learn about git and to experiment with how to use it effectively. When this PR is a single commit, with just the changes you've made, it is ready to be merged.

svr8:
I've opened a new PR for it since I had repeated the same mistake again(multiple commits). Please review. 
Thanks!

svr8:
I had messed up a lot of code during "rebase" somehow. I've opened a new PR (PR #1584 ) with a single commit as needed.

mrerrormessage:
Sounds good. Thank you!

PR #1576, opened 2018-02-22
Prevent plot & pen case sensitivity

mrerrormessage:
This both prevents the UI from creating plots and pens with the same-name-modulo-case AND converts existing models to ensure they have no plots or pens with the same-name-modulo-case.

Conversion is accomplished with the following technique:

* If duplicate names are found, they are sorted in reverse-string order (lowercase-first), then a number is appended to each name except for the first one. So `"abc" "aBC" "ABC"` become `"abc" "aBC_1", "ABC_2"`.
* The widgets are altered by renaming plots and pens as appropriate according to the renaming scheme. Plot and pen names are processed separately, so having a plot named "abc" and a pen named "ABC" won't convert either of those names.
* If any code in the model (including widget code) uses `set-current-plot` or `set-current-plot-pen`, two changes are made:
  1. As needed, procedures `_clarify-duplicate-{plot,plot-pen}` are added which disambiguate case-sensitive (old) names to case-insensitive (new) names. (see example code on L267 of ModelConverterTests).
  2. Code in the code tab and any widgets is altered from using `set-current-plot` to using `set-current-plot _clarify-duplicate-plot <original argument>`

I plan to do a bit more code-tidying, but would appreciate any feedback on both the general approach taken as well as on the renaming-and-conversion scheme that people (@TheBizzle, perhaps?) have to offer.

### DIFF
--- a/netlogo-core/src/main/api/CompilerServices.scala
+++ b/netlogo-core/src/main/api/CompilerServices.scala
@@ -8,8 +8,6 @@ trait CompilerServices extends LiteralParser {
   def dialect: Dialect
   def isConstant(s: String): Boolean
   @throws(classOf[CompilerException])
-  def readFromString(s: String): AnyRef
-  @throws(classOf[CompilerException])
   def readNumberFromString(source: String): java.lang.Double
   @throws(classOf[CompilerException])
   def checkReporterSyntax(source: String)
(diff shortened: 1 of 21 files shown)


TheBizzle:
### This pull request is :two: LEGIT :two: OMIT

:tada: :trophy: :1st_place_medal: :wine_glass: :tada:

mrerrormessage:
Thanks for the review!

PR #1578, opened 2018-02-23
Make Shape Editor Dialog Resize and Scale

TheFoolishMan:
There are 2 files which have been changed to make Shape Editor resize and scale.

## ShapeView.java

- Fixed one bug for crosshair.
- Changed color of crosshair from dark gray to white to distinguish it.
- Increased size from 300 to 400.

## EditorDialog.java

- Deleted drawingPanel and graphicPanel
- Created new panel called bottomPanel which contain previewPanel and buttonPanel
- Changed the layout of EditorDialog from BoxLayout to BorderLayout.
- Make set resizable value of Editor Dialog to true

### DIFF
--- a/netlogo-gui/src/main/shape/editor/EditorDialog.java
+++ b/netlogo-gui/src/main/shape/editor/EditorDialog.java
@@ -69,7 +69,7 @@ strictfp class EditorDialog
     shape.addObserver(this);
     shapeView = new ShapeView(this, shape);
     shape.addObserver(shapeView);
-    setResizable(false);
+    setResizable(true);
     AbstractAction closingAction = new AbstractAction() {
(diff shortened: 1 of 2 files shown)


mrerrormessage:
These are looking like some good first steps! There are a few changes I would like to see:

1. Some of your files have a mix of tab and space indentation. Switch your indenter two use two-spaces for tabs. You may need to reindent certain sections.
2. The shape of the pane with the crosshairs and grid lines needs to stay a square, so that width == height.  Google around and see if you can find a way to accomplish this.
3. The shape needs to stay centered relative to the crosshairs as the pane is resized.

Please reach out if you have further questions about how to accomplish any of these changes.

PR #1580, opened 2018-02-26
Ensure string outputs in BehaviorSpace results are only CSV-escaped

nicolaspayette:
...instead of being "NetLogo escaped" *and then* CSV-escaped.

Thus, the simple string:
```
x
```

is now written in the CSV result file as:

```
"x"
```

instead of:

```
"""x"""
```

And the following string, where the quotes are part of the string:

```
"x"
```

is now written in the CSV file as:

```
"""x"""
```

instead of:

```
"""\""x\"""""
```

The new behavior follows the RFC 4180 specification and matches the current behavior of the `csv` extension.

***

The justification for this is that BehaviorSpace output is mostly meant to be handled by external programs (like Excel, Calc, R, Python, etc.) rather than read back into NetLogo. Such external programs are usually equipped to handle one level of CSV escaping (as defined in RFC 4180), but then ignore (or choke on) the NetLogo-style escaping.

I suspect that outputting strings that contain quotes or other special characters through BehaviorSpace is a relatively uncommon occurrence, so the impact of this change should be limited, but it should be a big help for researchers that do need to externally handle such outputs.


nicolaspayette:
> Do you know what part of the Behaviorspace code in particular is passing boxed integers to CSV?

At least this:

https://github.com/NetLogo/NetLogo/blob/77360c28e281e20097f078081683741be2c134d7/netlogo-core/src/main/lab/Exporter.scala#L41-L43

I wondered if I should change it instead of handling it in the CSV object, but I figured the prohibition against boxed ints only made sense in the context NetLogo's engine and didn't see why BehaviorSpace should be prevented from writing these values.

mrerrormessage:
> I figured the prohibition against boxed ints only made sense in the context NetLogo's engine and didn't see why BehaviorSpace should be prevented from writing these values.

I agree. (As a bit of trivia, the world boundaries themselves are actually stored on `agent.World` as `int`s).

mrerrormessage:
Thanks @nicolaspayette !

PR #1581, opened 2018-02-26
loop, repeat, while introduce scope as needed. Fix #1540

### LINKED ISSUE #1540: Anonymous commands created inside a new scope in a repeat share scope they shouldn't

mrerrormessage:
Supersedes / replaces #1541 . 

The approach taken is to have the compiler check whether any let-variables introduced by a primitive are closed-over in a lambda contained inside that primitive. If so, `_enterscope` and `_exitscope` "synthetic" primitives (not directly traceable to user code) are inserted at the beginning and end of the primitive's command-block. 

Because the conditions here are fairly narrow, none of the benchdumps are affected. The transformation is tested in `MiddleEndTest` (part of this PR).


netlogo-gui/src/main/api/TokenHolder.scala
```diff
@@ -4,6 +4,7 @@ package org.nlogo.api
 
 import org.nlogo.core.Token
 
+@deprecated("6.0.3", "User core.TokenHolder instead")
```
Idloj: Typo, should be "Use core.TokenHolder instead" (Use, not User)

mrerrormessage: Nice catch!

qiemem:
This does solve NetLogo/LevelSpace#121!

I really like the approach with `SelfScoping` and `CompilerScoping`. Should those be moved to `api` so extensions can use them? Off the top of my head, `nw`'s generators should probably use `SelfScoping`.

What does `Scoping` do? Should the other two traits inherit from it?

mrerrormessage:
Thanks for the review!

> I really like the approach with SelfScoping and CompilerScoping. Should those be moved to api so extensions can use them? 

The compiler sees extension primitives as only `_extern` and `_externreport`, so it doesn't matter whether or not extension primitives have access to these traits. If we wanted to give extension primitives access to these traits, I would recommend modifying `core.Syntax`, but I think the number of extensions which require access to this feature is quite small and they can work around the problem by setting the scope before and after executing a block or implementing `CustomAssembled`.

> What does Scoping do? Should the other two traits inherit from it?

Nice catch, I intended for the other traits to inherit from it, I just forgot! 

`Scoping` is to indicate that any let-variables introduced within that primitive are not the responsibility of the parent primitive. To provide two concrete examples:

```
to scoping-for-ask
  repeat 5 [
    ask turtles [
      let a 5
      run [ -> print a ]
    ]
end

to scoping-for-if
  repeat 5 [
    if true [
      let a 5
      run [ -> print a ]
    ]
  ]
end
```

In this instance, `ask` is `Scoping` because lets inside an ask are not the responsibility of a containing loop construct to reset scope for. `if`, on the other hand, is not scoping because `let` inside the `if` uses the same binding as if it were directly inside `repeat`. I'm happy to clarify this with additional examples if needed and will be adding a test to cover this behavior.

qiemem:
Thanks for the clarification!

> I think the number of extensions which require access to this feature is quite small and they can work around the problem by setting the scope before and after executing a block or implementing CustomAssembled.

So, to clarify when this is needed:

- prims that just call lambdas *don't* need this as the creation of the lambdas already takes care of it. That's a pretty nice workaround.
- prims that use blocks *do* need this. So, `nw` definitely needs to be updated, yes?

mrerrormessage:
> prims that use blocks do need this. So, `nw` definitely needs to be updated, yes?

Prims that use blocks will need this only if their blocks run repeatedly. From what I can tell, `nw` doesn't need to be modified because all of its command-block prims use `runExclusiveJob`. I'm not aware of any other extensions which introduce iterative control structures, but my guess is that they would also have to use `runExclusiveJob` to evaluate the command argument supplied.

PR #1585, opened 2018-02-28
Miscellaneous source, doc fixes suggested by project scripts

mrerrormessage:
I ran some of the `bin` scripts and added the changes here.

I've gone ahead and added `strictfp` to Java classes as suggested. I'm skeptical that `strictfp` is needed at this point. [This SO post](https://stackoverflow.com/questions/22562510/does-java-strictfp-modifier-have-any-effect-on-modern-cpus) suggests that on CPUs supporting SSE2 (so pretty much every x86 Intel/AMD chip manufactured in the last 15 years) `strictfp` is redundant. This PR shows that `strictfp` hasn't been on a number of key classes (including agent.World) and we haven't seen any problems from that in 6.0.1/6.0.2. Not to mention that we don't use Scala's `strictfp` annotation on classes/methods anywhere and we've replaced increasingly large sections of NetLogo's computational core with Scala code. @SethTisue , do you have thoughts?


SethTisue:
@SethTisue no idea... last time I researched this was about 15 years ago

Issue #1568, opened 2018-02-17
Internal Error.


Achaean:
Going Help -> About and clicking the web link, returns this error (on Debian Jessie x64 KDE):

```
java.lang.UnsupportedOperationException: The BROWSE action is not supported on the current platform!
 at java.awt.Desktop.checkActionSupport(Desktop.java:225)
 at java.awt.Desktop.browse(Desktop.java:381)
 at org.nlogo.swing.BrowserLauncher$.$anonfun$openURI$1(BrowserLauncher.scala:26)
 at org.nlogo.swing.BrowserLauncher$.$anonfun$openURI$1$adapted(BrowserLauncher.scala:26)
 at scala.Option.map(Option.scala:146)
 at org.nlogo.swing.BrowserLauncher$.openURI(BrowserLauncher.scala:26)
 at org.nlogo.app.AboutWindow$$anon$4$$anon$8.mouseClicked(AboutWindow.scala:57)
 at java.awt.Component.processMouseEvent(Component.java:6536)
 at javax.swing.JComponent.processMouseEvent(JComponent.java:3324)
 at java.awt.Component.processEvent(Component.java:6298)
 at java.awt.Container.processEvent(Container.java:2236)
 at java.awt.Component.dispatchEventImpl(Component.java:4889)
 at java.awt.Container.dispatchEventImpl(Container.java:2294)
 at java.awt.Component.dispatchEvent(Component.java:4711)
 at java.awt.LightweightDispatcher.retargetMouseEvent(Container.java:4888)
 at java.awt.LightweightDispatcher.processMouseEvent(Container.java:4534)
 at java.awt.LightweightDispatcher.dispatchEvent(Container.java:4466)
 at java.awt.Container.dispatchEventImpl(Container.java:2280)
 at java.awt.Window.dispatchEventImpl(Window.java:2746)
 at java.awt.Component.dispatchEvent(Component.java:4711)
 at java.awt.EventQueue.dispatchEventImpl(EventQueue.java:758)
 at java.awt.EventQueue.access$500(EventQueue.java:97)
 at java.awt.EventQueue$3.run(EventQueue.java:709)
 at java.awt.EventQueue$3.run(EventQueue.java:703)
 at java.security.AccessController.doPrivileged(Native Method)
 at java.security.ProtectionDomain$JavaSecurityAccessImpl.doIntersectionPrivilege(ProtectionDomain.java:80)
[...]

PR #1600, opened 2018-04-02
Updated BrowserLauncher.scala to catch UnsupportedOperationException

amro-ashmeik:
This pull request is to resolve the issue (#1568)

### DIFF
--- a/netlogo-gui/src/main/swing/BrowserLauncher.scala
+++ b/netlogo-gui/src/main/swing/BrowserLauncher.scala
@@ -27,11 +27,12 @@ object BrowserLauncher {
       if (! opened)
         osSpecificBrowserRunner.getOrElse(browserNotFound()).apply(uri.toString)
     } catch {
-      val exceptionmsg = ex.getLocalizedMessage
+      val unableToOpenBrowserError = "We were unable to open a browser on your system.\n" +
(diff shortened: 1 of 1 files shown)


mrerrormessage:
I'm sorry, my advice on the issue was a bit unclear. Pull out the string in L113-114 to a `val` - call it `unableToOpenBrowserError` perhaps? Then have `UnsupportedOperationException` show a message dialog with `unableToOpenBrowserError`. The catch clause for `BrowserNotFoundException` can / should remain the same as before.

netlogo-gui/src/main/swing/BrowserLauncher.scala
```diff
@@ -27,6 +27,10 @@ object BrowserLauncher {
       if (! opened)
         osSpecificBrowserRunner.getOrElse(browserNotFound()).apply(uri.toString)
     } catch {
+      val unableToOpenBrowserError = "We were unable to open a browser on your system.\n" +
+      "This error can be reported to bugs@ccl.northwestern.edu"
```
mrerrormessage: This looks good. You can pull this out of the method and into the class and replace the string on L114 with `unableToOpenBrowserError`.

amro-ashmeik: Done!

PR #1612, opened 2018-04-18
Make link breed order comparison actually work

### LINKED ISSUE #1611: 6.0.3 : `link a b` fails to look up some directed links with at least three link breeds and three turtles
Dear Netlogo team,

NetLogo 6.0.3 and 6.0.2 reporters are not finding directed links even though they exist and can be viewed on screen and inspected. The same code works perfectly well in 6.0.1.

The code breaks down all the time when 3 or more directed links are used. With 2 directed breads it sometimes works.

A sample code file is attached.

[example_report_directednetwork_error.zip](https://github.com/NetLogo/NetLogo/files/1924097/example_report_directednetwork_error.zip)

qiemem:
Fixes #1611

### DIFF
--- a/netlogo-core/src/main/agent/LinkManagement.scala
+++ b/netlogo-core/src/main/agent/LinkManagement.scala
@@ -5,6 +5,7 @@ package org.nlogo.agent
 import org.nlogo.core.{ AgentKind, Program, Shape, ShapeListTracker }
 
 import java.util.{ HashMap => JHashMap, Map => JMap }
+import scala.collection.JavaConverters._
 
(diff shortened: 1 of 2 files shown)


SethTisue:
is this code performance-sensitive enough that dropping down to `while`-style code, like in the old version, might be preferable?

qiemem:
We were discussing that on gitter. This code is only hit when two links have the exact same endpoints, so I don't think so. Might be worth testing though.

SethTisue:
> This code is only hit when two links have the exact same endpoints

ah, ok, I see

PR #1620, opened 2018-05-02
Make `histogram` set plot x range and pen mode

### LINKED ISSUE #1274: beginner histogram
A common beginner experience with `histogram` is that they see nothing (because the data is outside the default x range).  It would be useful to have a `histogram-all` command that does a `set-plot-x-range` to the range in the data before histogramming it.  It should also reset the pen mode.  Doing something clever with the y range (after binning) would also help.

qiemem:
Fixes #1274 

This is not ready for merge yet. Things that are not addressed:

- Tests are not updated
- With this solution, users cannot use a fixed x range but expanding y range with histogram. The best solution to that problem would be to have independent x and y auto scaling options, but that requires a file format change, so may not happen for a bit

An alternative to this would be to add this as a new simple histogram primitive (say, `histogram-all` as suggested in #1274), but having such non-orthogonal primitives is unfortunate.

Mostly, I'm opening this PR so people can experiment with behavior.

### DIFF
--- a/netlogo-gui/src/main/plot/Plot.scala
+++ b/netlogo-gui/src/main/plot/Plot.scala
@@ -243,21 +243,19 @@ class Plot private[nlogo] (var name:String) extends PlotInterface with JSerializ
     plotListener.foreach(_.setHistogramNumBars(numBars))
   }
 
-  var histogram: Option[Histogram] = None
-
-  def beginHistogram(pen:PlotPen) {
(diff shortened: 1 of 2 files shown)


alan-isaac:
Closing this and https://github.com/NetLogo/NetLogo/issues/1274 is leaving teachers in the lurch.  We are left with no good, simple way to introduce histograms in a basic undergrad course.

Issue #1525, opened 2017-12-05
`mean` should include the name of the primitive responsible in their failure message


mrerrormessage:
`mean` fails due to an empty list, it currently display an error message like "List is empty.". It should display an error consistent with the other list primitives: "Can't find the mean of a list with no numbers: []"

This is an ideal fix for a new contributor, since it involves changing only `_mean.java` and the `Errors_en.txt` file.

ERSUCC:
This was fixed by pull #1623.

PR #1623, opened 2018-05-05
Update error message for mean of empty list

fenekku:
This includes changes for Netlogo-Gui and Netlogo-Headless as well as translations. Also reformats some code to better match [formatting guidelines](https://github.com/NetLogo/NetLogo/wiki/Code-formatting). I like whitespaces. Attempts to address  #1525 .

### DIFF
--- a/netlogo-gui/project/autogen/i18n/Errors_en.txt
+++ b/netlogo-gui/project/autogen/i18n/Errors_en.txt
@@ -84,6 +84,7 @@ org.nlogo.prim._greaterthan.cannotCompareParameters = The > operator can only be
 org.nlogo.prim._max.cantFindMaxOfListWithNoNumbers = Can''t find the maximum of a list with no numbers: {0}
 org.nlogo.prim._min.cantFindMinOfListWithNoNumbers = Can''t find the minimum of a list with no numbers: {0}
 org.nlogo.prim._mean.cantFindMeanOfNonNumbers= Can''t find the mean of a list that contains non-numbers : {0} is a {1}.
+org.nlogo.prim.etc._mean.cantFindMeanOfListWithNoNumbers = Can''t find the mean of a list with no numbers: {0}.
 
 org.nlogo.prim.etc.$common.emptyList = List is empty.
 org.nlogo.prim.etc.$common.emptyString = String is empty.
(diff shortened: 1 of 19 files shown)


SethTisue:
did this get forgotten about...?

fenekku:
Sorry about this. I was out of town for a while. I don't think the translations were verbatim copies from Google Translate. I am learning Japanese and I am familiar enough with Latin languages to be reasonably confident about the Japanese and Spanish translations. I don't know about the Chinese one.

I don't understand the steps to fix the "benchdumps" by reading through the wiki. If you provide me with more details I could do that too (might take some more time).
cheers,

LaCuneta:
@fenekku Thanks for looking back into this one for us!

Understood on the translations, I think our one for-sure concern was the use of "medio" instead of "promedio" in the Spanish translation.  If you want to change that one over and remove the Chinese one (if you're not sure on it), that'd be enough for me.

Since there has been a bit of work on `hexy` since this was submitted, it would be best to rebase to the latest to make sure to get the bench dumps correct.

The bench dumps aren't too hard to update, it should just be `netlogo/dump bench` at an `sbt` console, and then committing the changed text files in `test/benchdumps/`.  You can check them (before and after) with `netlogo/testOnly *TestCompileBenchmarks`.  

Feel free to amend your commits and force push.

fenekku:
Ok that was less complex than I thought! I have changed the Spanish translation and committed the `test/benchdumps` files. A couple of the `extensions` files also changed. Should those be committed as well?

LaCuneta:
Don't worry about the extensions, although you can get them reset to the correct commits using `git submodule update` if you want.

I actually forgot that `headless` also has separate dumps that need to be updated; I'm very sorry about that!  They are just as easy to update, though.  `headless/dump bench` and commit the changes (should be about the same as the other ones).  You can use `headless/testOnly *TestCompileBenchmarks` to check that things are good.  I'm going to update our docs so I don't forget these next time. 

Thank you, again for your help!

fenekku:
Thanks for the explanations. `headless/dump bench` seems to have done the trick. It is passing now :+1:

LaCuneta:
Nice, thank you for being patient!

PR #1624, opened 2018-05-06
make issues script run on latest sbt, Scala

SethTisue:
not sure if this script is being used anymore. (I found it very useful,
back in the day.)

if it hasn't been in use, perhaps it's because of all the scary stuff
about Conscript. none of that is necessary anymore -- the script should
work just fine without any special setup.


LaCuneta:
I had no idea this existed, it looks like it could save a lot of time!  Thanks for the updates!

PR #1626, opened 2018-05-13
Make multiline input widgets directly editable

### LINKED ISSUE #1593: Make the inline textbox in mult-line inputs editable
I totally get why the pop-out editor is necessary. But it seems like it would be quite useful for the inline textbox to be editable as well. These are not mutually exclusive things. Indeed, they're pretty complementary for quick edits vs. large edits.

qiemem:
Fixes #1593 

A few notes:

- Enter will still change focus (and thus apply changes) on single-line inputs
- Tab changes focus on both multi-line and single-line. I figure this was the safer change. People can just use spaces.
- The "change" button seems a little mislabeled now. I don't think it's a big deal, but if anyone has anyone ideas (something along the lines of "open editor") I'm happy to change it.

I needed this for a model I'm working on, so just implemented it...

### DIFF
--- a/netlogo-gui/src/main/window/InputBox.scala
+++ b/netlogo-gui/src/main/window/InputBox.scala
@@ -87,9 +87,18 @@ abstract class InputBox(textArea:AbstractEditorArea, editDialogTextArea:Abstract
   protected var multiline = false
   def multiline(multiline: Boolean) {
     this.multiline = multiline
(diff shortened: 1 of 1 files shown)


LaCuneta:
This looks good to me, just waiting on the 6.0.4 release to go out before merging to hexy.

PR #1636, opened 2018-06-07
Extensions Manager

Idloj:


qiemem:
This is so awesome! The interface is great. A few issues I ran into:

- When I first opened the dialog, when no extensions are installed, no extension was selected and the "Install" button was blank. Clicking on it causes an NPE.
- It would be nice to have a way to uninstall extensions, though I suppose I mostly want the for testing.
- Extension auto-complete appears to not work, even on built-ins

Also, mind if I add more extensions to the list of available extensions?

Finally, are includes files working yet? If so, how do you use that part?

TheBizzle:
> Finally, are includes files working yet? If so, how do you use that part?

`.nls` files aren't supported yet.  Adding extension support is phase 1.

qiemem:
Another issue. When I first open the dialog, the initially selected description shows fine:

image

but then selecting any other extension makes it blank out:

image

There's no way to get it back except by reopening the dialog.

Idloj:
@qiemem Can you try and reproduce the black-rectangle bug now? I've added 2 lines that I think will solve the issue.

qiemem:
Fixed! Thanks @Idloj !

Idloj:
Extension auto-complete works now. The problem was actually that the code wasn't automatically compiled when switching tabs. It is described [here](https://github.com/NetLogo/NetLogo/pull/1636/files#diff-742452b5e85fc1a4b5ece8f40ff757dfR429). I have no idea why this is the behavior, if anybody knows please explain to me.

Idloj:
@TheBizzle I'd like to resolve the following list of issues before this is merged in:
- Idloj/NetLogo#1
- Idloj/NetLogo#3
- Idloj/NetLogo#4
- Idloj/NetLogo#8

Sounds good?

TheBizzle:
Sure.  I'd say that the first issue is optional, but the other three almost certainly should go in before we proceed.

Idloj:
This is pretty much ready for merge, @TheBizzle. One thing that's left is updating the extensions, which I've already opened PRs for on the R extension and on LevelSpace.

qiemem:
@Idloj I'm a little hesitant to accept a PR that won't compile against hexy. How about update the submodules to point to the extensions' branches?

Idloj:
That's what I intended. After the PRs for R extension and LevelSpace are accepted, I'll update the submodules, the PR will compile successfully, and only then will the PR be merged.

qiemem:
I was actually suggesting doing it the other way. Update *this* repository to point to your branches on the extension repositories to get it compiling. Once this PR is merged, then we merge the PRs on the other repositories (though that doesn't even need to happen right away). That way, we won't have any commits on any of the repositories that don't build.

Personally, I don't think there is anything wrong with having the submodule pointers refer to branches. We've done it before plenty of times. I think it's more important to avoid mainline commits that don't compile to preserve easy bisecting and such.

Idloj:
> Update this repository to point to your branches on the extension repositories to get it compiling.

Which is why I'm suggesting to merge the PRs to a new branch (not the main one). I can't change the submodules to point to my repositories (I can technically, but you get the point), they need to point to the NetLogo repositories, not my forks of them.

qiemem:
Ack! Sorry, I completely misread your comment. My apologies!

qiemem:
Done! https://github.com/NetLogo/LevelSpace/tree/Idloj-extension-manager

Idloj:
Great, thanks! Can you do the same for the R extension?

Idloj:
Submodules updated. Ready for merge.

PR #1644, opened 2018-07-16
 [Libraries Dialog] Only enable "Update All" if there can be updates

### LINKED ISSUE #1643: "Update All" produces unidiomatic update text when there are no updates to run
_From @TheBizzle on July 5, 2018 20:27_

@LaCuneta reports:

> When I press "Update All" after installing a couple of things it says "Installing 0 extensions...", which never goes away. Maybe better to give a message like "No extensions need updates."?

_Copied from original issue: Idloj/NetLogo#16_

Idloj:
This resolves #1643 simply by making it impossible to click the button in cases where it won't do anything.


Idloj:
@LaCuneta,I've rebased

PR #1648, opened 2018-07-16
 Have a separate sub-dir inside the per-user dir for each minor version

Idloj:


netlogo-core/src/main/api/FileIO.scala
```diff
@@ -124,6 +124,33 @@ object FileIO {
...
+      if (os.contains("WIN"))
+        System.getenv("APPDATA") + "\\NetLogo"
+      else if (os.contains("MAC"))
+        System.getProperty("user.home") + "/Library/Application Support/NetLogo"
+      else
+        System.getProperty("user.home") + "/.netlogo"
+
+    appData + File.separator + path
+  }
+
```
TheBizzle: Why is all of this being exposed in the NetLogo API?  That doesn't seem right to me.

Idloj: I agree, but this package already contain a lot of classes (or traits) that aren't really part of the NetLogo API, but are instead internal to NetLogo, and I couldn't think of a better place.

TheBizzle: Well, if there are a bunch of things in `api` that aren't really a part of the API, that seems like a negative thing to me, and I feel that we shouldn't actively make the problem worse.  However, I _can_ be convinced that this is our best reasonable move.

Can you explain why it's better for this to be in `api` than in `util.Utils`, where you had it before?

Idloj: Per the Package Guide in the wiki:
>  nothing in here is supposed to be NetLogo-specific

Looking at the other functions there, this rule is being adhered to.

TheBizzle: Okay.  I don't really like it, but the argument against putting it in the `util` package is clear, and `RemoteFile` and `StringUtils` _do_ set a precedent for dumping this sort of stuff in `api` (much as I might disagree).

LaCuneta:
While checking this I noticed that when there are no extensions available, as with a testing NetLogo 6.1.0 build, the extensions manager was (understandably) blank.  While I don't think it's likely we'll ever release a NetLogo build without any extensions, it might be worth adding a "No extensions for this version found..." message in that case, even if just for NetLogo builders/maintainers.

PR #1650, opened 2018-07-17
ExtensionAssistant now offers the Libraries Dialog instead of the wiki

Idloj:


Idloj:
Good point! I did not think about that. What do you suggest instead?

TheBizzle:
If an extension with that name exists, open the Extensions & Include Files window.  If there is no known extension with that name, pop up a dialog that communicates these ideas:

  * Dear user, we aren't aware of any extension with that name
  * Please check to make sure that you spelled the name of your desired extension correctly
  * If you _don't know_ the exact name of the extension that you're looking for, feel free to go to "Tools > Extensions & Include Files" and browse through the extensions listed there
  * If you are certain that you provided the correct name of the extension, you'll need to manually download and install it in order for NetLogo to recognize it.  Here is a link to a page that explains how to do that.  Also, dear user, you might even consider submitting this extension to the NetLogo team, so that you and others in the NetLogo community can see it in the Extensions & Include Files window for automatic installation.

TheBizzle:
A version of this PR with the requested changes is implemented in 0fd46e4c2bb74a719ffd23b4d2189805ed52bd30 (not yet merged).

PR #1656, opened 2018-07-21
Add up-to-n-of

### LINKED ISSUE #1594: Add a primitive for getting at most n of something: `at-most`
A common problem in NetLogo is dealing with getting a `nobody` from a `one-of` or calling `n-of` when there aren't `n` agents in the agentset. While I think it's good that, for instance,

```
ask one-of other turtles-here [ set color red ]
```

errors when there aren't any other turtles here (it forces the users to think about a situation they may not have thought of), the following pattern is common enough that I think it may be worth adding a primitive:

```
if any? other turtles-here [ 
  ask one-of other turtles-here [ set color red ]
]
```

I propose `at-most` to handle these types of situations:

```
to-report at-most [ n agents ]
  report ifelse-value (n <= count agents) [ agents ] [ n-of n agents ]
end
```

That is, `at-most 5 turtles` will give you 5 turtles if there are 5 turtles, otherwise it will give you however many turtles there are. Thus, for the above situation, you would do:

```
ask at-most 1 other turtles-here [ set color red ]
```

and it would turn another turtle red if there are any other turtles here, and otherwise do nothing.

While this primitive overlaps a lot with `n-of`, I think this is a common enough use case that a new primitive is warranted. Furthermore, I think the name makes the distinction pretty clear: it says up front "you might not get all of the agents you are asking for".

qiemem:
Fixes #1594 .

### DIFF
--- a/netlogo-headless/src/main/prim/etc/_uptonof.scala
+++ b/netlogo-headless/src/main/prim/etc/_uptonof.scala
@@ -1,0 +1,26 @@
+package org.nlogo.prim.etc
+
+import org.nlogo.agent.AgentSet
+import org.nlogo.core.{I18N, LogoList, Syntax}
+import org.nlogo.nvm.{ArgumentTypeException, Context, Reporter, RuntimePrimitiveException}
+
+class _uptonof extends Reporter {
+  override def report(ctx: Context): AnyRef = {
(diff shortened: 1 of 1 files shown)


SethTisue:
shouldn't this include an addition the documentation...?

PR #1657, opened 2018-07-25
[Libraries Dialog] Handle bundled extensions correctly

Idloj:
I didn't implemented it exactly as I've written in https://github.com/NetLogo/NetLogo-Libraries/issues/3#issuecomment-406624443. Instead of copying the bundled resource, we use it as a Config and merge it with installed-libraries.conf (which can override values). See the [commit message](https://github.com/NetLogo/NetLogo/commit/bc7ae2970dfda11c619626010c59bdd3dea71341).

@TheBizzle, please review.

### DIFF
--- a/netlogo-gui/resources/system/bundled-libraries.conf
+++ b/netlogo-gui/resources/system/bundled-libraries.conf
@@ -1,0 +1,64 @@
+{
+    "extensions" : {
+        "arduino" : {
+            "installedVersion" : "3.0.0"
+        },
+        "array" : {
+            "installedVersion" : "1.1.0"
+        },
+        "bitmap" : {
+            "installedVersion" : "1.1.0"
+        },
+        "cf" : {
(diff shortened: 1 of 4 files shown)


TheBizzle:
I'd do this myself, if I could, but could you update the version numbers for the extensions to all have a patch version (e.g. "1.0.0" instead of "1.0")?

Also, please update a few of the version numbers as follows:

    gogo     - 2.0.0
    ls       - 2.2.3
    nw       - 3.7.5
    view2.5d - 1.1.0

Idloj:
Updated

TheBizzle:
@Idloj, hopefully this will be the last set of version numbers that I'll need you to update:

```
arduino - 3.0.0 (as was already stated in its `build.sbt`)
rnd     - 3.0.0 (as indicated by the repo's Git tags)
```

PR #1658, opened 2018-07-27
Fix: Runtime logo excpetions were causing an NPE to show up

Idloj:
GUIWorkspace was getting a null as an ErrorDialogManager.

@TheBizzle please review.

### DIFF
--- a/netlogo-gui/src/main/app/App.scala
+++ b/netlogo-gui/src/main/app/App.scala
@@ -288,6 +288,9 @@ class App extends
 
     org.nlogo.swing.Utils.setSystemLookAndFeel()
 
+    errorDialogManager = new ErrorDialogManager(frame,
+      Map(classOf[MetadataLoadingException] -> new LibraryManagerErrorDialog(frame)))
+
     org.nlogo.api.Exceptions.setHandler(this)
     Thread.setDefaultUncaughtExceptionHandler(new Thread.UncaughtExceptionHandler() {
       def uncaughtException(t: Thread, e: Throwable) { org.nlogo.api.Exceptions.handle(e) }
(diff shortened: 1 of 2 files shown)


TheBizzle:
Could you say a bit more about this?  Under what conditions was `errorDialogManager` `null`?  Why does adding laziness (through a call-by-name param and a `lazy val`) fix the problem?  If you're fixing an NPE by adding laziness, my suspicion is that that isn't going to solve the core problem, but I'd be interested in hearing a fuller explanation before making a judgment on that.

Also, while you at it, could you fix up the commit message?  "exceptions" is spelled incorrectly.

Idloj:
`errorDialogManager` is initially declared as `null` (like all the other `var`s there; https://github.com/Idloj/NetLogo/blob/hexy/netlogo-gui/src/main/app/App.scala#L267).
When passing it to GUIWorkspace (https://github.com/Idloj/NetLogo/blob/hexy/netlogo-gui/src/main/app/App.scala#L322), its value is still `null`. `errorDialogManager` is only initialized [after `tabs.init`](https://github.com/Idloj/NetLogo/blob/hexy/netlogo-gui/src/main/app/App.scala#L433) because, as described in the comment, the dialogs `ErrorDialogManager` contains have to be initialized after `tabs.init`. The NPE was triggered when getting a runtime error while running a piece of NetLogo code, because GUIWorkspace manager tries to invoke `show` on `errorDialogManager` which is `null` (https://github.com/Idloj/NetLogo/blob/hexy/netlogo-gui/src/main/window/GUIWorkspace.java#L1045).

So to solve this bug, `errorDialogManager` has to be initialized before creating GUIWorkspace, while the dialogs it contains need to be initialized after the call to `tabs.init`.
The solution was to initialize `errorDialogManager` [here](https://github.com/NetLogo/NetLogo/pull/1658/files#diff-742452b5e85fc1a4b5ece8f40ff757dfR291), and have its dialogs created lazily, so they are not created as soon as `ErrorDialogManager` is created (which is before `tabs.init`).

PR #1665, opened 2018-08-21
Handle offline conditions. Closes #1660

### LINKED ISSUE #1660: Handle offline situations in Library Manager
It should simply use the local metadata file, and somehow communicate to the user that the computer is offline. Currently it pops up an error.

Idloj:
@TheBizzle please review.


LaCuneta:
Looks good to me, much more user friendly than before.  Jenkins reported a test failure, but I cannot reproduce locally and a re-run gave no errors on Jenkins, either.

PR #1668, opened 2018-09-14
Make ifelse and ifelse-value variadic

### LINKED ISSUE #344: add `switch` or `case` statement, and/or `cond`, to language
Uri writes:

> in the past weeks, several colleagues hve complained to me about
> netlogo not having a case or switch. I agree that nested if-elses
> are a drag.  I know we've had this on our list for years.  How
> hard would it be to do?

We had a devel meeting about this circa 2005 and nobody liked any of the syntaxes that were proposed.  They all involved an awful lot of square brackets that nobody could see how to get rid of.

We could do it again and then just bite the bullet and pick the least bad syntax we can think of. The actual coding on this would not be a huge project, just a medium sized one (for someone who knows the compiler and engine).

Berkeley Logo has both `case` and `cond`.  (We might decide that if we support `cond`, we don't need `case`.)

`case` example from Berkeley Logo manual:

```
to vowelp :letter
output case :letter [ [[a e i o u] "true] [else "false] ]
end
```

`cond` example from Berkeley Logo manual:

```
to evens :numbers   ; select even numbers from a list
op cond [ [[emptyp :numbers] []]
          [[evenp first :numbers]  ; assuming EVENP is defined
           fput first :numbers evens butfirst :numbers]
          [else evens butfirst :numbers] ]
end
```

Lots of brackets, very Lisp-y.

An example of when you want `cond`, from @reuven's Autumn model:

```
to set-shape-for-value [value]
  ifelse value > 75 [ set shape "default" ]
  [ ifelse value <= 25 [ set shape "default one-quarter" ]
    [ ifelse value <= 50 [ set shape "default half" ]
                         [ set shape "default three-quarter" ]]]
end
```

If we adopted the Berkeley `cond` syntax this would be:

``` text
to set-shape-for-value [value]
  cond [ [[value >  75] set shape "default" ]
         [[value <= 25] set shape "default one-quarter" ]
         [[value <= 50] set shape "default half" ]
         [else          set shape "default three-quarter" ] ]
end
```

This reads OK, but is hard to remember exactly how to write.

An example of when you want `case`, from Heat Diffusion:

``` text
to update-alpha
  ifelse (material-type = "wood") ;; wood
  [ set alpha 0.00128 ]
  [ ifelse (material-type = "stone") ;; stone
    [ set alpha 0.012 ]
    [ ifelse (material-type = "iron")  ;; iron
      [ set alpha 0.2034 ]
      [ ifelse (material-type = "aluminum") ;; aluminum
        [ set alpha 0.8418 ]
        [ ifelse (material-type = "silver")  ;; silver
          [ set alpha 1.7004 ]
          [ user-message "Choose your own value for alpha!" ] ] ] ] ]
end
```

If we adopted the Berkeley `case` syntax this would become:

``` text
to update-alpha
  case material-type
   [ [ ["wood"]     set alpha 0.00128 ]
     [ ["stone"]    set alpha 0.012   ]
     [ ["iron"]     set alpha 0.2034  ]
     [ ["aluminum"] set alpha 0.8418  ]
     [ ["silver"]   set alpha 1.7004  ]
     [ else user-message "Choose your own value for alpha!" ] ]
end
```

Again, this reads OK, but is hard to remember exactly how to write.

qiemem:
This PR makes `ifelse` and `ifelse-value` variadic, following through with the plan laid out in this comment: https://github.com/NetLogo/NetLogo/issues/344#issuecomment-360206584. Thus, this PR fixes #344.

Furthermore it:

- Lowers the precedence of `ifelse-value`, thus fixing #588. Note that this is a breaking change. For instance, the Robby the Robot model must be changed (@bainco).
- Generalizes the assembler's goTos/comeFroms to use labels. I needed to use multiple gotos, and it was just as easy to generalize to full labels, so I just did that. This should make it easy to implement, e.g., and efficient switch statement in the future.

While CF used anonymous procedures to implement this functionality, I found that the performance hit was unacceptably high. This PR uses blocks instead and consequently has no performance hit for `ifelse` and only a very tiny hit for `ifelse-value` (we could probably eliminate it as well with the assembler, but it's almost immeasurable as it is). 

There are several tasks to be done before this PR is ready:

- [x] Update benchdumps.
- [x] Make sure headless works ok.
- [ ] Make sure `ifelse` throws type errors when conditions/blocks are mixed up. The problem is that we don't know something is wrong until the assembler stage. The different strategies I see here are:
    - Expand syntax definitions to allow more complex variadic types so we can differentiate blocks and booleans in the type signature. This sounds like a ton of work.
    - Just throw the type errors in the assembler. The downsides to this are that all compiler type errors are deeply embedded in the expression parser right now, and are not easy to reuse, and that the assembler can only differentiate between reporters and command blocks. That is, it won't know what type of reporter its looking for.
    - Have `_ifelse.assemble` catch the current `ClassCastException` that is being thrown and create an appropriate error. This is pretty hacky.
- [x] Remove the cf extension
- ~[ ] Auto-convert cf extension calls to core calls~
- ~[ ] Auto-convert old `ifelse-value` calls to new ones. Not familiar enough with the auto-converter to know if this is possible, but when infix operators have an `ifelse-value` expression as an argument, the `ifelse-value` expression  needs parens around it. That said, this may not be possible, since, given the new precedence, we won't actually know that the `ifelse-value` expression would be an argument to the infix operator. I suppose we could also include a `_ifelse-value-old` and then just do token rewriting. Or we could just not worry about it, since this is a very particular case of not super commonly used primitive.~

The models library will also have to be updated.


qiemem:
So the byte code of the bench dumps looks fine, but some of the string representations of the code changed. I'm not sure how to judge those. For example: https://github.com/NetLogo/NetLogo/pull/1668/commits/ad4b45bf211fd9a262b2326fdb76c0246b9755d3#diff-31b7be7092fc3297e693ac5e2c6f13aeL1966

@SethTisue @mrerrormessage, if you guys have a moment, do those look okay?

bainco:
Note that the breaking change resulting from the fix to #588 has been addressed in the 1 appropriate model (`Robby the Robot`) in NetLogo/models@de34035. Other updates to the Models Library resulting from the larger changes to `ifelse` and `ifelse-value` will be tracked in NetLogo/models#374 and merged after this PR has been merged.

qiemem:
Note, the checksum changes in fea84aa are entirely due to the models library update.

qiemem:
@Idloj You mean just to remove CF, correct?

Idloj:
Correct

SethTisue:
> @SethTisue @mrerrormessage, if you guys have a moment, do those look okay?

seems like a regression, but I'm not sure. the best I can do is say it looks worth digging into

Issue #1606, opened 2018-04-13
Ctrl+ shortcuts don't work on some keyboard layouts


miguelppais:
This seems to happen because semicolon and brackets need an additional key to access on some layouts.

In my case (Portuguese layout), semicolon needs a shift key and brackets need Alt Gr

Other layouts with this issue may be: ru/es/fi

Similar issues on other software: [here](https://github.com/processing/processing/issues/2199) and [here](https://bugs.documentfoundation.org/show_bug.cgi?id=88027)

It would be nice to have a quick way to comment/uncomment, but I can't seem to get it to work.

nicolaspayette:
> Other layouts with this issue may be: ru/es/fi

<kbd>Ctrl</kbd>+<kbd>[</kbd> and <kbd>Ctrl</kbd>+<kbd>]</kbd> don't work on the French Canadian keyboard either. Italian colleagues also complained to me that some other shortcuts don't work (I think it's <kbd>Ctrl</kbd>+<kbd>+</kbd> and <kbd>Ctrl</kbd>+<kbd>-</kbd> for zooming, but I'm not 100% sure.)

Basically, anything that's not a letter isn't a safe bet for keyboard shortcuts.

ERSUCC:
Unfortunately I don't think there's a good way for this to be solved internally. The best solution would be to allow users to rebind keys as they wish. This idea has already been floated around a bit due to a keybind conflict on Mac, so it may well make its way into the app in the future.

PR #1672, opened 2018-10-03
Add a few hotkeys - jump-to declaration, show usage, and save-all

LaCuneta:
New hotkeys:

- Command-Alt-S: Save-all (for multiple open+dirty includes files)
- Command-U: Show usage
- Command-E: Jump to declaration

If those are bad choices for some reason, I'm open to suggestions for easier-to-use keys.

### DIFF
--- a/netlogo-gui/project/autogen/i18n/GUI_Strings_en.txt
+++ b/netlogo-gui/project/autogen/i18n/GUI_Strings_en.txt
@@ -196,6 +196,7 @@ tabs.info.help = Help
 
 tabs.code = Code
 tabs.code.rightclick.jumptodeclaration = Jump To Declaration
+tabs.code.rightclick.showusage = Show Usage
 tabs.code.rightclick.quickhelp = Quick Help
 tabs.code.rightclick.quickhelp.notfound = {0} could not be found in the NetLogo Dictionary.\n \
    Would you like to open the full NetLogo Dictionary?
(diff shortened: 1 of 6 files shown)


nicolaspayette:
> I think command-period is kind of a standard for "jump to definition"?

I don't know about command-period in particular, but any shortcut not involving plain old letters has the potential to be problematic on international keyboard layouts. My Italian colleagues can't use <kbd>Ctrl</kbd>+<kbd>+</kbd> to zoom in, for example, and I can't use <kbd>Ctrl</kbd>+<kbd>/</kbd> to hide the command centre on my French-Canadian keyboard (though the latter is not much of a loss...)

LaCuneta:
So <kbd>Ctrl</kbd>+<kbd>J</kbd> in NetLogo does a "remove end-of-line-break" or "combine with next line" operation.  I can't find a definition for this in the code, so I assume it's a default for the text editor control.  <kbd>Ctrl</kbd>+<kbd>.</kbd> looks to be open, though, if that's better than 'E'.

nicolaspayette:
Related: #1606.

> So <kbd>Ctrl</kbd>+<kbd>J</kbd> in NetLogo does a "remove end-of-line-break" or "combine with next line" operation.

Woah. Never knew that. Does anyone actually use that? Still, I wouldn't want to take it away from big <kbd>Ctrl</kbd>+<kbd>J</kbd> fans out there if there are any.

I think <kbd>Ctrl</kbd>+<kbd>.</kbd> *might* be a problem with the French AZERTY keyboard (which weirdly requires pressing <kbd>Shift</kbd>+<kbd>;</kbd> to get a period).

Keeping <kbd>Ctrl</kbd>+<kbd>E</kbd> seems like the safe option...

LaCuneta:
I did a little more research and didn't find any consistent key for "Jump to..." among editors and IDEs, so I think we'll start with <kbd>Ctrl</kbd>+<kbd>E</kbd> and change later if needed (or maybe move specification to a config file?)

PR #1674, opened 2018-10-10
Make lambda arrow checking take into account depth

qiemem:
Fixes this:

`ifelse-value true [ [ -> true ] ] [ [ -> false ] ]`

This was reported by @nicolaspayette.

### DIFF
--- a/parser-core/src/main/parse/ArrowLambdaScoper.scala
+++ b/parser-core/src/main/parse/ArrowLambdaScoper.scala
@@ -2,16 +2,16 @@
 
 package org.nlogo.parse
 
-import org.nlogo.core.{ Fail, Token, TokenType }, Fail.exception
-import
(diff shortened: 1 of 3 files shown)


qiemem:
One behavioral change (well, there's several, but this captures most of them I believe): `[ [ -> ] ]` no longer causes the lambda parser to error. Instead, it causes the list literal parser to error. I would argue that this new behavior is more correct.

Previously, the compiler would parse  the outer block of `[ [ -> ] ]` as a lambda and the inner block as an argument list, and thus would throw an error indicating that `->` is not valid variable name. Now, it correctly parses the inner block as the lambda (because it is a valid lambda) and the outer block as a list literal. It then errors since lambdas are not allowed in list literals. The tests have been updated to reflect this.

LaCuneta:
LGTM, thanks for the extra explanation for the test changes.

PR #1684, opened 2018-11-02
Fix : __includes accepts comments and throws with no open bracket

charlesres:


parser-core/src/main/parse/StructureParser.scala
```diff
@@ -138,18 +138,19 @@ object StructureParser {
...
       tokens.dropWhile(! _.text.equalsIgnoreCase("__includes"))
     if (includesPositionedTokens.isEmpty)
       Seq()
     else {
       includesPositionedTokens.next
-      if (includesPositionedTokens.next.tpe != core.TokenType.OpenBracket)
-        Seq()
+      val includesWithoutComments = includesPositionedTokens.filter(_.tpe != core.TokenType.Comment)
+      if (includesWithoutComments.next.tpe != core.TokenType.OpenBracket)
+        exception("Missing open bracket in structureParser", tokens.next)
```
LaCuneta: Should this be something like "Did not find expected open bracket for __includes declaration"?  We'll know which file it's in from the stack trace, so it should be something that might help out a user who managers to trigger the error.

charlesres: Yea, I'll update the exception to better reflect the error.

Issue #1618, opened 2018-04-30
Runtime error "no such plot" if plot was deleted and then delete was reverted


irmingard:
If I delete a plot and then undo the widget deletion, I get a runtime error when using the "set-current-plot" method that says "no such plot" even though it clearly exists.

Internal details:

org.nlogo.nvm.RuntimePrimitiveException: no such plot: "My test plot"
 at org.nlogo.nvm.AnonymousCommand.perform(AnonymousProcedure.scala:181)
 at org.nlogo.nvm.AnonymousCommand.perform(AnonymousProcedure.scala:158)
 at org.nlogo.prim.etc._foreach.perform(_foreach.scala:33)
 at org.nlogo.nvm.Context.stepConcurrent(Context.java:107)
 at org.nlogo.nvm.ConcurrentJob.step(ConcurrentJob.scala:65)
 at org.nlogo.job.JobThread.runPrimaryJobs(JobThread.scala:133)
 at org.nlogo.job.JobThread.$anonfun$run$1(JobThread.scala:68)
 at scala.runtime.java8.JFunction0$mcV$sp.apply(JFunction0$mcV$sp.java:12)
 at scala.util.control.Exception$Catch.apply(Exception.scala:224)
 at org.nlogo.api.Exceptions$.handling(Exceptions.scala:41)
 at org.nlogo.job.JobThread.run(JobThread.scala:66)

NetLogo 6.0.2
main: org.nlogo.app.AppFrame
thread: JobThread
Java HotSpot(TM) 64-Bit Server VM 1.8.0_141 (Oracle Corporation; 1.8.0_141-b15)
operating system: Linux 3.13.0-117-generic (amd64 processor)
Scala version 2.12.2
JOGL: (3D View not initialized)
OpenGL Graphics: (3D View not initialized)
model: bidirectional

04:01:13.507 RuntimeErrorEvent (org.nlogo.app.App$$anon$1 (org.nlogo.window.GUIWorkspace)) AWT-EventQueue-0
04:01:13.496 JobRemovedEvent (org.nlogo.app.App$$anon$1 (org.nlogo.window.GUIWorkspace)) JobThread
04:01:13.482 InterfaceGlobalEvent (org.nlogo.app.interfacetab.InterfacePanel$$anon$1 (org.nlogo.window.SliderWidget)) AWT-EventQueue-0
[...]

PR #1688, opened 2018-11-16
Fix: #1618 Plot is readded to plot list, extensionPath changed to extensionsPath, and new checksums

charlesres:
After updating plotManager and plotWidget, Jenkins failed at building the new changes. After some fixes, extensionManager and JarLoader have been updated, and a new checksum was added with model changes.


netlogo-gui/src/test/plot/PlotManagerTests.scala
```diff
@@ -44,6 +44,36 @@ class PlotManagerTests extends SimplePlotTest {
     assertResult(None)(manager.currentPlot)
   }
 
+    test("More Current Plot") {
```
LaCuneta: Spacing is off here.

charlesres: I've fixed the whitespace issue and rebased the commit messages.  I tried adding more context and a reason for the commit.

LaCuneta:
The messages look better, thanks.  Note that the character count limit for git is 72 before it auto-breaks, which can look a bit messy.  We [try to stay to 72](https://github.com/NetLogo/NetLogo/wiki/Code-formatting) but I often forget to trim them.

PR #1691, opened 2018-12-07
Slider no longer reverts to 0

charlesres:
The slider should no longer revert back to 0 after opening a model when saved with a compile error.
A testsuite was added to keep tabs on the series of events that caused the issue in the first place.

### DIFF
--- a/netlogo-gui/src/main/window/CompilerManager.scala
+++ b/netlogo-gui/src/main/window/CompilerManager.scala
@@ -171,12 +171,12 @@ class CompilerManager(val workspace: AbstractWorkspace,
     // this method again. - ST 7/7/06
     if (!isLoading) {
       val proceed = compileProcedures()
-      if (proceed) {
-        world.realloc()
-        world.rememberOldProgram()
-        setGlobalVariables() // also updates constraints
-        compileWidgets()
-      } else {
+      world.realloc()            // reallocates widgets
+      world.rememberOldProgram() // updates world dialect, which is
+                                 // important for 3D to 2D and vice versa
+      setGlobalVariables() // updates constraints
+      compileWidgets()     // update the widget once the compile is done
+      if (!proceed) {
         // even if compilation of the procedure tab fails, we still want to mark our
         // constraints as out of date, so that any existing dynamic constraints are
         // thrown away since they're compiled against the old program -- CLB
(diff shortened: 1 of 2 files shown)


netlogo-gui/src/test/window/CompilerManagerTests.scala
```diff
@@ -245,3 +334,23 @@ class DummyIGWidget(source: String, var value: AnyRef = Double.box(0)) extends D
   }
   def valueObject: AnyRef = value
 }
+
+class DummyTeSliderWidget(source: String, var value: AnyRef = Double.box(0)) extends DummyWidget(source) with InterfaceGlobalWidget {
+//  override def classDisplayName =  I18N.gui.get("tabs.run.widgets.slider")
```
LaCuneta: If this isn't needed, go ahead and remove it

netlogo-gui/src/test/window/CompilerManagerTests.scala
```diff
@@ -245,3 +334,23 @@ class DummyIGWidget(source: String, var value: AnyRef = Double.box(0)) extends D
...
+  def valueObject(x: AnyRef): Unit = x match {
+    case a: AnyRef => value = a
+  }
+  def valueObject: AnyRef = value
+  var increment = 1.0
+  var maximum = 100.0
+  var minimum = 0.0
+
+  // this sets the value in the current constraint and then ensures
+  // cached values are updated -- CLB
```
LaCuneta: Make sure to add a date (at least month/year) for comments.

netlogo-gui/src/main/window/CompilerManager.scala
```diff
@@ -171,12 +171,12 @@ class CompilerManager(val workspace: AbstractWorkspace,
...
-        world.realloc()
-        world.rememberOldProgram()
-        setGlobalVariables() // also updates constraints
-        compileWidgets()
-      } else {
+      world.realloc()            // reallocates widgets
+      world.rememberOldProgram() // updates world dialect, which is
+                                 // important for 3D to 2D and vice versa
+      setGlobalVariables() // updates constraints
+      compileWidgets()     // update the widget once the compile is done
```
LaCuneta: If these comments are helpful for the people reading the code, maybe pull them out into a separate chunk explaining things more generally, then sign and date them.

netlogo-gui/src/test/window/CompilerManagerTests.scala
LaCuneta: Double-checking, this might be a comment from a PR, in which case you can delete it if not needed.

SethTisue: Craig L. Brozefsky

charlesres: Yea, it was copied straight from DummySliderWidget.

Issue #1679, opened 2018-10-15
Plot Pen Case Sensitivity Auto-Conversion Too Sensitive


bainco:
In #1576 the auto-converter was updated to prevent plot & pen case sensitivity by implementing a renaming scheme for pens that were identical up to case insensitivity. However, it appears the auto-converter is _too_ sensitive as it will auto-convert pen names that are identical (up to case insensitivity) _*across*_ plots. 

This means that if you create a model (see attached) with two plots where:

1. Plot 1 has a single pen named `default`
2. Plot 2 has a single pen named `DEFAULT`

The model will save and run just fine, but when the model is reopened in NetLogo post 3ae13aa the auto-converter, following the rules from #1576, will rename the `DEFAULT` pen to `DEFAULT_1`. While this is not a breaking change, it is unnecessary.
 
[plot_pen_demo.nlogo.zip](https://github.com/NetLogo/NetLogo/files/2480316/plot_pen_demo.nlogo.zip)

LaCuneta:
Attaching a different sample model:

[Duplicate pen names.nlogo.zip](https://github.com/NetLogo/NetLogo/files/2672465/Duplicate.pen.names.nlogo.zip)

PR #1693, opened 2018-12-07
Fix: #1679 Added AllLocalPenNames to reduce the scope of allPenName

charlesres:
AllLocalPenNames limits which names are renamed through grouping names to their plot and checking each group instead of the whole list.

### DIFF
--- a/netlogo-core/src/main/fileformat/PlotConverter.scala
+++ b/netlogo-core/src/main/fileformat/PlotConverter.scala
@@ -18,7 +18,7 @@ object PlotConverter {
       .filterNot(_._1.isEmpty)
   }
 
-  private[fileformat] def mapRenameDecision(groupedNames: Seq[(String, Seq[String])]):  Seq[(String, String)] = {
+  private[fileformat] def generateRenameMappings(groupedNames: Seq[(String, Seq[String])]):  Seq[(String, String)] = {
     groupedNames
       .map {
         case (upper, originals) => upper -> originals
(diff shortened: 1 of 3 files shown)


netlogo-core/src/main/fileformat/PlotConverter.scala
```diff
@@ -5,36 +5,53 @@ package org.nlogo.fileformat
...
+                new_name  = s"${original}_${new_index}"
+              }
+              (original, s"${original}_${new_index}")
+            }
           }
         case _ => Seq.empty[(String, String)]
       }
   }
 
+  private[fileformat] def determineMapRenames(names: Seq[(String, Seq[String])]): Seq[(String, Seq[(String, String)])] = {
```
LaCuneta: The names for this and `determineRenames` are slightly confusing to me - would `determinePenRenames` and `determinePlotRenames` be any better?

netlogo-core/src/main/fileformat/PlotConverter.scala
```diff
@@ -100,30 +116,36 @@ class PlotConverter(
...
         val plotRenames = determineRenames(allPlotNames(model))
-        val penRenames = determineRenames(allPenNames(model))
-        if (plotRenames.isEmpty && penRenames.isEmpty) {
-          super.apply(model, modelPath)
-        } else {
-          val conversion = super.apply(model, modelPath)
-          conversion.updateModel(convertPlotAndPenNames(conversion.model, plotRenames.toMap, penRenames.toMap))
+        val penRenames: Seq[(String, Seq[(String,String)])] = determineMapRenames(allKeyedPenNames(model))
+        (plotRenames, penRenames) match {
+          case (Seq(), Seq()) => super.apply(model,modelPath)
```
LaCuneta: Careful with spacing: `super.apply(model, modelPath)`

netlogo-core/src/main/fileformat/PlotConverter.scala
```diff
@@ -100,30 +116,36 @@ class PlotConverter(
...
-              p.copy(display = newDisplay, pens = newPens)
-            case w => w
-          }
-          model.copy(widgets = updatedWidgets)
+   private def convertPlotAndPenNames( // you need to create a pipeline
+        model:       Model,               // Seq[Seq[String]] -> Seq[(String, String)] -> Complete
+        plotRenames: Map[String, String], // Apply each set of changes to its corresponding list
+        penRenames:  Map[String, Seq[(String, String)]]): Model = { // I need penRenames with list of list
+                                                     // List[Map[String,String]], possibly
+                                                     // this would need to be generated as we go
```
LaCuneta: I think the comments on this method should be moved before it into a more readable paragraph.  Also, sign and date.

netlogo-core/src/main/fileformat/PlotConverter.scala
```diff
@@ -100,30 +116,36 @@ class PlotConverter(
...
+       val updatedWidgets = model.widgets.map{
+         case p: Plot => {
+           val newDisplay = p.display.map(currentName => plotRenames.getOrElse(currentName, currentName))
+           val localRenames = penRenames.getOrElse((p.display.getOrElse("")), Seq()).toMap
+           val newPens = p.pens.map {
+             pen =>
+             val newPenName = localRenames.getOrElse(pen.display, pen.display)
+             pen.copy(display = newPenName)
```
LaCuneta: Spacing and whitespace look a little off here.

netlogo-core/src/test/fileformat/PlotConverterTest.scala
```diff
@@ -0,0 +1,163 @@
...
+
+  /*************
+   * Plot Name *
+   *************/
+
+  test("if the plot names are not initialized"){
+    val model = Model(widgets = Seq( View()
+                                   , Plot(display = None, pens = List(Pen(display = "dep", setupCode = "plot e 10")))
+                                   , Plot(display = None, pens = List(Pen(display = "DeP", setupCode = "plot e 10")))
+                                   , Plot(display = None, pens = List(Pen(display = "Dep", setupCode = "plot e 10")))))
```
LaCuneta: It would be nice (but not critical) if the rest of the test setups were nicely aligned with whitespace like this one is.

netlogo-core/src/main/fileformat/PlotConverter.scala
```diff
@@ -5,36 +5,53 @@ package org.nlogo.fileformat
...
+  def allKeyedPenNames(model: Model): Seq[(String, Seq[String])] = {
+    model.plots
+      .map((plot: Plot) => (plot.display.getOrElse(""), plot.pens.toSeq.map(_.display)))
+      .filterNot(_._1.isEmpty)
+  }
 
-  private def determineRenames(names: Seq[String]): Seq[(String, String)] = {
-    names.groupBy(_.toUpperCase)
-      .toSeq
+  private[fileformat] def mapRenameDecision(groupedNames: Seq[(String, Seq[String])]):  Seq[(String, String)] = {
```
LaCuneta: The name here is ambiguous to me, since `map` might be a verb, but it might be a noun.  Either way, I think we can clarify to something more along the [naming guidelines](
https://github.com/NetLogo/NetLogo/wiki/Code-formatting#naming), like `generateRenames`?

netlogo-core/src/main/fileformat/PlotConverter.scala
```diff
@@ -5,45 +5,85 @@ package org.nlogo.fileformat
...
+
+
+  private def clarifyPensProcedureBody(
+    renamePlotPairs: Seq[(String,String)],
+    renamePairs: Seq[(String, Seq[(String, String)])]): String = {
+    val nameMap = renamePlotPairs.map(p => s"""["${p._1}" "${p._2}"]""").mkString("[", " ", "]")
+    val nameReMap = renamePairs.flatMap( renamePair =>
+      renamePair._2.map(p => s"""[ "${renamePair._1}" "${p._1}" "${p._2}" ]""")).mkString("[", " ", "]")
+
+    s"""    |  let plot-name-map ${nameMap}
```
LaCuneta: It seems like we should be able to check if `renamePlotPairs` is empty - and if it is just use `plot-name` as the `reported-plot`, and otherwise if there were plot renames, just call the plot rename procedure `_clarify-duplicate-plot-name` instead of re-inserting the code for it here.  

If we do keep the code this way, then I think `nameMap` and `nameReMap` might need better names.

netlogo-core/src/main/fileformat/PlotConverter.scala
```diff
@@ -5,45 +5,85 @@ package org.nlogo.fileformat
...
-        |    set reported-name item 1 replacement
-        |  ]
-        |  report reported-name""".stripMargin
+    s"""    |  let name-map ${nameMap}
+    |  let replacement filter [ rename -> first rename = name ] name-map
+    |  let reported-name name
+    |  if not empty? replacement [
+    |    set reported-name item 1 (item 0 replacement)
+    |  ]
+    |  report reported-name""".stripMargin
```
LaCuneta: The pipes for the whitespace alignment should line up (as they did before the changes).

PR #1707, opened 2019-02-04
Scala, sbt version bumps

SethTisue:


SethTisue:
(not sure if I should be targeting hexy or ising?)

charlesres:
Targeting Ising would be best since it will have the latest batch of fixes for the 6.1 release.

SethTisue:
well this is interesting, upgrading to 2.12.7 is fine, but then going to 2.12.8 breaks org.nlogo.fileformat.ThreeDViewReaderTest . I'll look into it

SethTisue:
even a test such as

```
  test("sanity") {
    assert(ThreeDViewReader.validate(List(
      "GRAPHICS-WINDOW", "0", "0", "439", "460", "-1", "-1", "13.0", "1", "10", "1", "1", "1", "0", "1", "1", "1", "-16", "16", "-16", "16", "-16", "16", "1", "0", "1", "ticks", "30.0"
    )))
  }
```

fails on 2.12.8. some weird interaction with https://github.com/scala/scala/pull/7340 and parboiled? idk

SethTisue:
@mrerrormessage any insight?

SethTisue:
sorry but I don't have time to dig into this further

Issue #1312, opened 2017-02-11
Negative number of runs in BehaviorSpace


ToonTalk:
If one naively creates way too many runs for an experiment it can show a negative number of runs:

![image](https://cloud.githubusercontent.com/assets/2138696/22853207/f96a2b3a-f08a-11e6-8b41-099412ef7279.png)

mrerrormessage:
Thanks for reporting this. I think to fix this we will need to cap the number of runs at or below the range in which Java can use a `double` to represent an integer exactly.

PR #1710, opened 2019-02-21
Fixes: #1312 Added guards against integer oveflows

charlesres:
This minor fix adds another condition to EditPanel's valid() 
to determine if the product of totalRuns is less than the 
number of representable positive integers.


netlogo-core/src/main/api/LabProtocol.scala
```diff
@@ -14,7 +14,7 @@ case class LabProtocol(name: String,
                     metrics: List[String],
                     valueSets: List[RefValueSet])
 {
-  def countRuns = repetitions * valueSets.map(_.toList.size).product
+  def countRuns: Int = repetitions * valueSets.map(_.toList.size: Int).foldLeft(1)((accum, n) => accum * n)
```
SethTisue: I don't understand the intent behind this change. Why is this preferable to the old code?

charlesres: The change is, mostly, for style, preferring foldLeft over product. There is a small advantage in this context, such as providing space to add error handling for integer overflow for each multiplication. But it is not necessary for this fix.

SethTisue: what's wrong, style-wise, with `product`?

charlesres: The issue I had with product is that .product can apply to any type T with an implicit Numeric[T] implementation. So it would apply for Long, Float, and BigInt, which could "change" the type/behavior. With foldLeft, there are two type constraints to satisfy. The accumulator and the result must have the same type.
```
 foldLeft[B](z: B)(op: (B, A) ⇒ B): B 
```
This would, also, allow adding Options for error handling.

TheBizzle: @charlesres, the `foldLeft` isn't giving you any added protection against type coercion.  The type ascription (of `Int`) on `def countRuns` already guarantees that you'll get a type error if `product` is run on non-`Int`s.

Regarding the `Option` stuff, do we have any plans of expanding this to more-general monoids?  If we do that, can't we handle that when it comes up?  I don't see any problem with the old use of `product`.  It seems a lot more stylistically pleasant to me than manually writing out the fold.

charlesres: @TheBizzle, I agree that product would be better suited in that case. As for plans, I don't think there are plans to expand that case, so I'll keep product in LabProtocol.scala instead of foldLeft.

netlogo-gui/src/main/api/Editable.scala
```diff
@@ -18,6 +18,7 @@ trait Editable {
   // it's kind of lame to put this here but it'll require a bunch of changes all over the properties
   // package otherwise it seems not worth the effort ev 6/10/08
   def sourceOffset: Int
+  def fieldCheck(v:AnyRef): Boolean = true
```
LaCuneta: I don't see this used anywhere, I think it can be removed.

netlogo-gui/src/test/properties/CodeEditorTests.scala
```diff
@@ -0,0 +1,79 @@
+package org.nlogo.properties
+
+import org.scalatest.FunSuite
+//import org.nlogo.api.CompilerServices
```
LaCuneta: Commented line can be removed.

netlogo-gui/src/main/lab/gui/ProtocolEditable.scala
```diff
@@ -1,12 +1,11 @@
 // (C) Uri Wilensky. https://github.com/NetLogo/NetLogo
-
```
LaCuneta: Prefer to keep a line after the copyright notice.

netlogo-core/src/main/api/LabProtocol.scala
```diff
@@ -14,7 +14,7 @@ case class LabProtocol(name: String,
                     metrics: List[String],
                     valueSets: List[RefValueSet])
 {
-  def countRuns = repetitions * valueSets.map(_.toList.size).product
+  val countRuns = repetitions * valueSets.foldLeft(1)((accum,elt) => accum * elt.length.toInt)
```
SethTisue: this got back in? nbd I guess

charlesres: Hi @SethTisue. That's my bad since this is leftover from fixing the SteppedValueSets issue with space. This should have been:
```
.map(_.length.toInt).product
```

PR #1711, opened 2019-02-22
Added condition for mixed NetLogo files

charlesres:
This fix adds another condition for opening files with NetLogo. The additional condition
checks if the NetLogo version and source file are 3D while the file extension is 2D. If the
version match up in this case, NetLogo 3D will happily open a .nlogo file with a mislabeled
NetLogo section ("NetLogo 3D").

### DIFF
--- a/netlogo-core/src/main/fileformat/NLogoFormat.scala
+++ b/netlogo-core/src/main/fileformat/NLogoFormat.scala
@@ -149,8 +149,10 @@ trait AbstractNLogoFormat[A <: ModelFormat[Array[String], A]] extends ModelForma
     def validationErrors(m: Model): Option[String] = None
     override def deserialize(s: Array[String]) = {(m: Model) =>
       val versionString = s.mkString.trim
-      if(versionString.startsWith("NetLogo 3D") && Version.is3D(versionString) && !is3DFormat)
-        Success(m.copy(version = versionString.replace("3D ","")))
+      if(!versionString.startsWith("NetLogo 3D") && versionString.startsWith("NetLogo") && is3DFormat)
+        Success(m.copy(version = versionString))
+      else if(versionString.startsWith("NetLogo 3D") && Version.is3D(versionString) && !is3DFormat)
+        Success(m.copy(version = versionString))
       else if (versionString.startsWith("NetLogo") && Version.is3D(versionString) == is3DFormat)
         Success(m.copy(version = versionString))
       else {
--- a/netlogo-core/src/main/fileformat/NLogoFormat.scala
+++ b/netlogo-core/src/main/fileformat/NLogoFormat.scala
@@ -79,6 +79,10 @@ trait AbstractNLogoFormat[A <: ModelFormat[Array[String], A]] extends ModelForma
         val sectionLines = sections.getOrElse(name, Array[String]())
         if (sectionLines.isEmpty)
           "\n"
+        else if (name == "org.nlogo.modelsection.version"){
+          if(!is3DFormat) sectionLines.map(_.replace("3D ", "")).mkString("\n", "\n", "\n")
+          else sectionLines.mkString("\n", "\n", "\n")
+        }
         else if (sectionLines.head.isEmpty || sectionLines.head.startsWith("\n") || name == "org.nlogo.modelsection.code")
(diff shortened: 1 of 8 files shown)


netlogo-core/src/test/fileformat/ModelSectionTest.scala
```diff
@@ -51,6 +51,15 @@ trait ModelSectionTest[A, B <: ModelFormat[A, B], C] extends FunSuite {
...
 
+  def testErrorsOnDeserializationAndWrongArity(
+    description: String, serializedVersion: A, error: String): Unit = {
+    test(s"errors when deserializing $description") {
+      val s = subject
+      val m = s.deserialize(serializedVersion)(new Model())
+      assert(!m.isFailure)
+    }
+  }
+
```
LaCuneta: This is called `testErrorsOnDeserializationAndWrongArity`, so the `testErrors...` part makes me think the result should be an error, but the `assert(!m.isFailure)` is checking success.  Would a better name be something like `testNoErrors...`?

netlogo-core/src/test/fileformat/NLogoFormatTests.scala
```diff
@@ -111,6 +111,42 @@ class NLogoFormatConversionTest extends FunSuite with ConversionHelper {
       assert(rereadModel.code.contains("extensions [vid]"))
     }
 
+    test("performs tests for sectionToSource conversion conversion between files") {
+      val sectionsMap = collection.immutable.Map("org.nlogo.modelsection.version" -> Array[String]("NetLogo 3D 6.0.4"))
+      val nlogoformat = new NLogoFormat
```
LaCuneta: It's confusing to see the `nlogoformat` here, when the `nlogoformatThreed` is declared below.  I think they should be together, probably below the expected results.

netlogo-core/src/test/fileformat/NLogoFormatTests.scala
```diff
@@ -111,6 +111,42 @@ class NLogoFormatConversionTest extends FunSuite with ConversionHelper {
       assert(rereadModel.code.contains("extensions [vid]"))
     }
 
+    test("performs tests for sectionToSource conversion conversion between files") {
```
LaCuneta: The string param here gives the test its name, so this test will be called `performs tests for...`, but we already know this will perform tests because it's a test.  A better name should be more of the form `NOUN SHOULD VERB WHEN CONDITION`.

netlogo-core/src/test/fileformat/NLogoFormatTests.scala
```diff
@@ -111,6 +111,42 @@ class NLogoFormatConversionTest extends FunSuite with ConversionHelper {
...
+          |@#$#@#$#@
+          |@#$#@#$#@
+          |@#$#@#$#@
+          |@#$#@#$#@
+          |@#$#@#$#@
+          |@#$#@#$#@""".trim.stripMargin
+
+      assert(nlogoformat.sectionsToSource(sectionsMap).get.trim == expectedResult)
+      val nlogoformatThreed = new NLogoThreeDFormat
+      assert(nlogoformatThreed.sectionsToSource(sectionsMap).get.trim != expectedResult)
```
LaCuneta: It's jarring to see a test check that `someVal != expectedResult`, normally you want to see your expected result as the value.  Maybe `expectedResult` could be renamed to match `expectedThreedResult` so it's more clear why it should be different?

netlogo-core/src/main/fileformat/NLogoFormat.scala
```diff
@@ -79,6 +79,10 @@ trait AbstractNLogoFormat[A <: ModelFormat[Array[String], A]] extends ModelForma
         val sectionLines = sections.getOrElse(name, Array[String]())
         if (sectionLines.isEmpty)
           "\n"
+        else if (name == "org.nlogo.modelsection.version"){
+          if(!is3DFormat) sectionLines.map(_.replace("3D ", "")).mkString("\n", "\n", "\n")
```
charlesres: This line seems to only check if the format is not 3D and convert the version to 2D, but doesn't convert in the other way (2D -> 3D).

PR #1716, opened 2019-03-09
Beef Up Extension Manager

TheBizzle:
Here's what's been accomplished in this PR:

  * Provided much clearer messaging about when ExtMan extensions are being overridden by manually-installed extensions (e.g. the ones in `extensions/`), and the path where you can find the overriding extension
  * Provided a mini-wizard for finding/installing extensions when a non-installed extension is referenced in the Code tab
  * Added a button for inserting an extension into the Code tab from the ExtMan UI
  * In the ExtMan UI, explicitly state the latest and the currently-installed versions of the selected extension
  * Made it so that, in the release, bundled extensions are put into `extensions/.bundled` (which has lower priority than ExtMan, while `extensions/` has higher priority)
  * Got `TestCompileAll` working again (lel)
  * Disabled the damned linting in the SBT Scala console(!)
  * Various UX improvements
  * Various bugfixes
  * Improved documentation around extensions and the ExtMan

I've only tested this on Linux.  Testing on Windows and Mac OS is welcome.


SethTisue:
"beef up" 😁

TheBizzle:
> I think the `bundled-libraries.conf` version of the Python extension is off - it says it can update from 1.0.0 to 0.3.0 in the list.

As per a discussion that @LaCuneta and I had, this one was broken before this PR, and he's going to update the version number outside of this PR, anyway, so I'm not going to touch it.

> On macOS the extension paths in the "View Conflicting Libraries" window are raw

Good call.  [Fixed!](https://github.com/NetLogo/NetLogo/pull/1716/files#diff-fa81e86f6c2e0a7b8067343a4b878528R331)

> If I have a comment in my extensions declaration like `extensions ; ha-ha!\n[ array ]`, then the "Add to Code Tab" button adds a new extensions declaration instead of updating the one that is already there.

..."Ha-ha"?  **I don't find this funny at all!**

Do you find something comical about my code when I am writing my regular expressions?  Every programmer *needs* to write regular expressions, even the very thoughtless.  This was the most accurate regular expression that I could devise.  Should *I*, therefore, be made the subject of fun?  Would you like it if *I* laughed at *your* regular expressions?!

That scene from The Simpsons when Nelson laughs at the really tall guy, and then the really tall guy humiliates him

Nah, you got me good on this one.  It never even occurred to me that someone would or could put a comment between `extensions` and `[`.  Now I know.  [Now it's fixed.](https://github.com/NetLogo/NetLogo/pull/1716/files#diff-fdb522f4f2ef368191dcb65f63cb83ceR389)

Thanks!

TheBizzle:
Thanks, @charlesres.  I believe that that error is now fixed, and I pushed two other UI bugfixes to the branch, as well.

Issue #1402, opened 2017-05-22
Numeric widgets can be made multiline when changed from String type


LaCuneta:
- Add a new String input, mark it Multi-Line and OK.
- Edit it and set the Type to Number.  "Multi-Line" goes grey but retains the check.  OK.
- The input now acts like a multi-line input, but still only accepts a single number value.  

I think the expected behavior would be that the numeric input continues to act as the "normal" single-line version even when the box is checked.  Or possibly that the multi-line value should auto-clear when type is changed to numeric (when saved and reloaded, the field is cleared for numerics).

mrerrormessage:
Fixes for this will live in [InputBox.scala](https://github.com/NetLogo/NetLogo/blob/hexy/netlogo-gui/src/main/window/InputBox.scala). Whether or not the `changeButton` is visible, which is controlled (in part) by whether the current `inputType` has `changeVisible`. The relevant lines are [multiline set](https://github.com/NetLogo/NetLogo/blob/hexy/netlogo-gui/src/main/window/InputBox.scala#L91), [typeOptions set](https://github.com/NetLogo/NetLogo/blob/hexy/netlogo-gui/src/main/window/InputBox.scala#L289) and overriding [`changeVisible`](https://github.com/NetLogo/NetLogo/blob/hexy/netlogo-gui/src/main/window/InputBox.scala#L500) as appropriate.

TheBizzle:
Fixed by [853af3a](https://github.com/NetLogo/NetLogo/pull/1724/commits/853af3a6f9700c2281d71a66f600251f6adbabd0)

PR #1724, opened 2019-03-19
Fix for numeric widgets accepting multiline input when changed from String type

SaiPratyush:
This is a fix for issue #1402 

The multiline value is auto cleared once the type is changed from string to numeric type.


charlesres:
Hi @SaiPratyush. The updates look good to me. The input widget does update the interface behavior of numbers and multiline strings.

PR #1738, opened 2019-04-15
Remove cf primitives and extension via autoconverter

qiemem:
Also add general extension removal support to the AST rewriter, along with unit tests.

### DIFF
--- a/netlogo-core/src/main/fileformat/AutoConversionList.scala
+++ b/netlogo-core/src/main/fileformat/AutoConversionList.scala
@@ -66,6 +66,16 @@ object AutoConversionList {
     },
     "NetLogo 6.0-BETA2" -> {
       changeAllCode("rename range to _range", Seq(_.replaceToken("range", "_range")), Seq("range"))
+    },
+    "NetLogo 6.1" -> {
+      changeAllCode("remove CF extension; replace with build-in primitives",
+        Seq(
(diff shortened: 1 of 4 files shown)


netlogo-core/src/main/fileformat/AutoConversionList.scala
```diff
@@ -66,6 +66,16 @@ object AutoConversionList {
...
       changeAllCode("rename range to _range", Seq(_.replaceToken("range", "_range")), Seq("range"))
+    },
+    "NetLogo 6.1" -> {
+      changeAllCode("remove CF extension; replace with build-in primitives",
+        Seq(
+          _.removeExtension("cf"),
+          _.replaceToken("cf:ifelse", "ifelse"),
+          _.replaceToken("cf:ifelse-value", "ifelse-value"),
+        ),
+        Seq("cf")
```
LaCuneta: Would it be better to check for `Seq("cf:ifelse", "cf:ifelse-value")` instead?  I know it's unlikely someone used just `cf` without it being the extension, but if we can be more precise that seems preferable to me.

qiemem: We need to cover the case that the extension is imported but not used. This won't do anything to files that just use `cf` as a variable name or anything; it will scan them, but won't change anything.

PR #1747, opened 2019-05-22
Use array appends of for headless JVM args

qiemem:
Indexes were not being calculated properly, so arguments were being
overwritten if the user added, e.g., `--3D`.

Note that I've tested these changes in my distribution of 6.1, but haven't actually tested the mustache build targets, since that requires a full packaging for distribution build.

The way I tested was by throwing `echo "${JVM_OPTS[*]}` in the appropriate places and watching to see if arguments got overwritten or not.


qiemem:
> ran the produced headless shell script and checked with jps -v to see the JVM args, and they included all three options

I believe there should be four arguments when you run with `--3D`. When you don't run with any extra args, the append doesn't occur.

LaCuneta:
> I believe there should be four arguments when you run with `--3D`. When you don't run with any extra args, the append doesn't occur.

Right, I meant the original three options are included.

PR #1749, opened 2019-05-24
Implement NotAny Optimization

diparthshah:
This PR introduces ` _equal(_count, _constdouble: 0.0) => _not(_any(*))` optimization. This is also part of GSoC 2019 project.

### DIFF
--- a/netlogo-core/src/main/compile/middle/optimize/package.scala
+++ b/netlogo-core/src/main/compile/middle/optimize/package.scala
@@ -334,4 +334,20 @@ package optimize {
       }
     }
   }
+  // _equal(_count, _constdouble: 0.0) => _not(_any(*))
+  object NotAny extends RewritingReporterMunger {
+    val clazz = classOf[_equal]
+    def munge(root: Match) {
+      val count = root.matchOneArg(classOf[_count])
+      if(root.matchOtherArg(count, classOf[_constdouble]).reporter.asInstanceOf[_constdouble]
+           .primitiveValue == 0)
+      {
+        val oldRoot = root.node.asInstanceOf[ReporterApp]
+        root.strip()
+        root.replace(classOf[_not])
+        val any = root.addArg(classOf[_any], oldRoot)
+        any.graftArg(count.matchArg(0))
+      }
+    }
+  }
 }
--- a/netlogo-core/src/main/nvm/Optimizations.scala
+++ b/netlogo-core/src/main/nvm/Optimizations.scala
@@ -33,7 +33,8 @@ object Optimizations {
         "Nsum4",       // optimizes summing neighbor4 values
         "PatchAt",     // optimizes patch at offsets
         "RandomConst", // inlines const argument to random
-        "With"))       // optimizes "with" to patch-col / patch-row
+        "With",        // optimizes "with" to patch-col / patch-row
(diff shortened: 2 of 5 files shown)


qiemem:
Cool! Great idea! Could this be generalized further? Make a `_hasexactly(n, agentset)` which will bale early if the number of matches exceeds `n` or if the number of misses exceeds `count - n`? Then `_equal(_count, _constdouble: n)` could be optimized for all `n`. Could do similar optimizations for the other numeric comparison operators as well.

diparthshah:
Hey @qiemem that sounds like a good idea. My other parts of my GSoC project is too implement:
` _equal(_count, _constdouble: 0.0) => _not(_any(*))`
` _greaterthan(_count, _constdouble: 0.0) => _any(*)`
` _notequal(_count, _constdouble: 0.0) => _any(*)`
and also port them to Tortoise project too. I will surely like to implement your idea after that.

qiemem:
That's great! Consider though:

- You can write these in a way that generalizes for all `n` (since counting is monotonic), so you don't actually need to write special cases for `0`. Just trying to save you work here ;)
- I think you can actually write just one optimized primitive that will do everything, or at least, a primitive that you can instantiate with conditions. Haven't quite worked this out, but basically something that takes a predicate that shouldn't be violated during counting. May not be worth it; just a thought.

diparthshah:
Hey @qiemem , i have made small step according to idea you proposed. 
https://github.com/diparthshah/NetLogo/commit/aad1efbe845f231057cc5c39125129f0af372e5a
Can you check if this is something we are aiming at ?

qiemem:
Don't have time to do a super close look right now, but that definitely looks closer to what I had in mind. I'll add other comments on the code itself. Note that I think there is probably a generalization that would let you get `_greaterthan`, `_lessthan`, etc with the same optimized primitive, but maybe it's better to do them separate first, and then combine once the correct abstraction is apparent.

diparthshah:
Hey @qiemem i read your comments on the code and made some changes accordingly here: https://github.com/diparthshah/NetLogo/commit/de02a192ede90cb44aeced3f7588eeb5c8bd2abd
Let me know if i have understood that correctly. Thanks.

diparthshah:
Here's the dump comparison: https://www.diffchecker.com/mXbHjadq

qiemem:
Commented! The dump doesn't tell us much since it's just calling out to that method now. Have you actually benchmarked it yet?

Oh, I realized one way you could generalize. You could add an `inBounds` method or something that takes a lower bound and upper bound, with signature:

```
def inBounds(lower = 0, upper = Int.MaxValue): Boolean
```

For `=`, set them to the same value. For `>` and `>=`, set lower bound to the appropriate value. For `<` and `<=` set the upper bound. The upper bound gives you a clear early exit criterion and you could also exit early on the lower bound based on `ArrayAgentSet.arraySize` (though this is approximate of course). 

Also, it feels a little funny discussing commits that aren't actually in this PR here, so you might consider adding them to this PR or opening a new one. Also, be good to get input from @LaCuneta or someone else more active in NetLogo than me.

diparthshah:
Hey, @qiemem Thanks for your inputs on this idea. I'll definitely look into it.

Issue #1742, opened 2019-05-07
3D mode relies on global java property


JusteRaimbault:
The switch between 3d and 2d worlds is handled with the `org.nlogo.api.Version.is3d` method, which uses the jvm property `org.nlogo.is3d`.
This becomes an issue when trying to simultaneously run a 3d model and a 2d model in the same jvm (one will necessarily fail).
Would for example putting this global variable inside the HeadlessWorkspace (so 3d and 2d workspaces could be simultaneously handled) be problematic ?

SethTisue:
I always wanted to refactor this and never got around to it :-(

JusteRaimbault:
is `Version.is3d` used in many places besides HeadlessWorkspace when in headless mode ? otherwise setting it as a variable of the HeadlessWorkspace would be a quick fix ? I have no idea how deep the implied refactoring is

SethTisue:
(sorry, I don't remember)

LaCuneta:
There is a little bit of information related to this in [the `nlogox` file format notes](https://github.com/NetLogo/NetLogo/wiki/The-nlogox-format#arity-considerations).  My understanding is that @mrerrormessage had started some of the work to make this happen, but that there was still a lot left to do.  Sorry I don't know more than that!

mrerrormessage:
Sorry for the slow response. @LaCuneta 's link is the best documented source of information. As of the time I stopped work on the nlogox branch (a little over a year ago) I had refactored the code enough that it was possible to change out 2D and 3D modes at runtime (there was a menu item which allowed NetLogo to switch between 2D and 3D). I'm unsure how up-to-date this branch is or whether it has a targeted release time. The changes were quite substantial to be able to move from 2D to 3D.

To your original question:

> Would for example putting this global variable inside the HeadlessWorkspace (so 3d and 2d workspaces could be simultaneously handled) be problematic ?

I ended up doing something similar, which was introducing a concept called a `Dialect` which encapsulated version and language information. The fundamental internals problem was the circular dependency between the workspace and the compiler. Having `Dialect`, which was neither a workspace nor a compiler solved the problem.

If you are interested in running 2D & 3D models side-by-side, I believe the path of least resistance would be to start with the nlogox branch and update LevelSpace to work with that branch (which made substantial changes to the extensions API) and also updating LevelSpace to support 3D models (it only supported 2D as of the time I stopped working on the nlogox branch). 

[...]

PR #1750, opened 2019-05-29
Headless workspace : add a 3d property

JusteRaimbault:
Quick fix affecting only the HeadlessWorkspace used from the api, answering to [issue 1742](https://github.com/NetLogo/NetLogo/issues/1742)
( see [this openmole issue](https://github.com/openmole/openmole/issues/412) of why it was needed in a model embedding context )

### DIFF
--- a/netlogo-gui/src/main/headless/HeadlessWorkspace.scala
+++ b/netlogo-gui/src/main/headless/HeadlessWorkspace.scala
@@ -30,16 +30,19 @@ object HeadlessWorkspace {
    * Makes a new instance of NetLogo capable of running a model "headless", with no GUI.
    */
   def newInstance: HeadlessWorkspace =
-    newInstance(classOf[HeadlessWorkspace])
+    newInstance(classOf[HeadlessWorkspace],false)
+
+  def newInstance(is3d: Boolean): HeadlessWorkspace = newInstance(classOf[HeadlessWorkspace],is3d)
 
   /**
    * If you derive your own subclass of HeadlessWorkspace, use this method to instantiate it.
    */
-  def newInstance(subclass: Class[_ <: HeadlessWorkspace]): HeadlessWorkspace = {
+  def newInstance(subclass: Class[_ <: HeadlessWorkspace],is3d: Boolean = false): HeadlessWorkspace = {
     val pico = new Pico
-    pico.addComponent(if (Version.is3D) classOf[World3D] else classOf[World2D])
+    //pico.addComponent(if (Version.is3D) classOf[World3D] else classOf[World2D])
+    pico.addComponent(if (is3d) classOf[World3D] else classOf[World2D])
     pico.add("org.nlogo.compile.Compiler")
-    if (Version.is3D)
+    if (is3d)//(Version.is3D)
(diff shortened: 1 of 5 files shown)


netlogo-gui/src/main/headless/HeadlessWorkspace.scala
```diff
@@ -30,16 +30,19 @@ object HeadlessWorkspace {
    * Makes a new instance of NetLogo capable of running a model "headless", with no GUI.
    */
   def newInstance: HeadlessWorkspace =
-    newInstance(classOf[HeadlessWorkspace])
+    newInstance(classOf[HeadlessWorkspace],false)
+
+  def newInstance(is3d: Boolean): HeadlessWorkspace = newInstance(classOf[HeadlessWorkspace],is3d)
```
LaCuneta: Small style issue, we prefer space after commas in parameter lists.

netlogo-gui/src/main/headless/HeadlessWorkspace.scala
```diff
@@ -30,16 +30,19 @@ object HeadlessWorkspace {
...
 
   /**
    * If you derive your own subclass of HeadlessWorkspace, use this method to instantiate it.
    */
-  def newInstance(subclass: Class[_ <: HeadlessWorkspace]): HeadlessWorkspace = {
+  def newInstance(subclass: Class[_ <: HeadlessWorkspace],is3d: Boolean = false): HeadlessWorkspace = {
     val pico = new Pico
-    pico.addComponent(if (Version.is3D) classOf[World3D] else classOf[World2D])
+    //pico.addComponent(if (Version.is3D) classOf[World3D] else classOf[World2D])
```
LaCuneta: I saw a few places where commented code was left in.  That can be removed.

JusteRaimbault:
I investigate if the addition to world is superfluous - not sure it was necessary (furthermore it seems to cause problems with the build of headless and `FakeWorld`)

JusteRaimbault:
It seems that the way the `HeadlessWorkspace` is constructed through pico injection, it would not be possible to set an immutable property is3d without using the property is3d of the injected world. What would be possible is an internal var that we set before returning the new HeadlessWorkspace, but I'm not sure this is good practice ? (but would limit the modifications to this file indeed).

Also one question : I worked on the code in `netlogo-gui` ; what about the project in `netlogo-headless` (which does not seem to have an api ?) ? to what extent is the `HeadlessWorkspace` in that project different from the one in `netlogo-gui` ? (that we use when calling the api in headless mode)

SethTisue:
(pretty sure pico allows overriding choices that would otherwise be automatically made?)

SethTisue:
(as an aside, the whole picocontainer thing was something of an experiment (of mine) that I now consider failed, and not only because picocontainer itself is now a dead, unmaintained project)

LaCuneta:
Closing this for now.  I'm still open to a change with a smaller footprint if possible.

JusteRaimbault:
Sorry I had left this unfinished - I made a new PR with a footprint only on `HeadlessWorkspace`

PR #1803, opened 2019-12-18
Add links to histogram documentation. Fixes #1800

### LINKED ISSUE #1800: histogram documentation enhancement
Unlike most Dictionary entries, in the histogram entry references to other primitives do not link to the docs for those primitives. Also, there are no "See also" entries, which should list primitives useful for histogram creation.

brandesNW:
Histogram documentation now links to NetLogo commands it references

### DIFF
--- a/autogen/docs/dictionary.html.mustache
+++ b/autogen/docs/dictionary.html.mustache
@@ -3507,9 +3507,9 @@ set hidden? not hidden?
         <p>
           The histogram is drawn on the current plot using the current plot
           pen and pen color. Auto scaling does not affect a histogram's
-          horizontal range, so set-plot-x-range should be used to control the
+          horizontal range, so <a href="#set-plot--range">set-plot-x-range</a> should be used to control the
           range, and the pen interval can then be set (either directly with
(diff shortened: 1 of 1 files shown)


autogen/docs/dictionary.html.mustache
```diff
@@ -3507,9 +3507,9 @@ set hidden? not hidden?
         <p>
           The histogram is drawn on the current plot using the current plot
           pen and pen color. Auto scaling does not affect a histogram's
-          horizontal range, so set-plot-x-range should be used to control the
+          horizontal range, so <a href="#set-plot--range">set-plot-x-range</a> should be used to control the
```
SethTisue: `set-plot--range`, is that the right link?

brandesNW:
Hi Seth,
Thanks for taking a look, especially as I am new at this.
I think it is correct. It’s because two commands are described in one section.
Earlier in the file NetLogo/autogen/docs/dictionary.html.mustache where the plotting commands are grouped you can find the lines

      <a href="#set-plot--range">set-plot-x-range</a>
      <a href="#set-plot--range">set-plot-y-range</a>

That show the pattern.

I generated both the html and pdf document and tested all the links.


  *   Aaron

PR #1817, opened 2020-02-06
Doc current plot

### LINKED ISSUE #1787: Current plot does not reset when deleting a plot
Steps to reproduce:

- Open Wolf Sheep Predation
- Run `plot-pen-exists? "default"` and you'll get `false` back because the plot in the WSP model doesn't have that pen.
- Add a new plot using all the pre-filled values (no changes).  This new plot becomes the current plot.
- In the command center run `plot-pen-exists? "default"` and you'll get `true` back because the new plot has a `"default"` pen.
- Delete the new plot.
- Run `plot-pen-exists? "default"` and you'll get a runtime error `"There is no current plot. Please select a current plot using the set-current-plot command.  error while observer running PLOT-PEN-EXISTS?"`

It seems like NetLogo tries to keep the current plot updated when opening models and adding plots, but not when plots are deleted.  I think it should also update the current plot when a plot is deleted, if there is another to choose.  

This was reported by a user using the Systems Dynamics Modeler which will throw the error if you delete a plot after adding it since it uses the `plot-pen-exists?` primitive in its auto-generated code.

brandesNW:
Added documentation to resolve #1787 for commands requiring a current plot or plot pen.

### DIFF
--- a/autogen/docs/systemdynamics.md.mustache
+++ b/autogen/docs/systemdynamics.md.mustache
@@ -312,6 +312,7 @@ a plot to watch the changes in sheep population.
 >
 >         to setup
 >           ca
+>           set-current-plot "populations"
 >           system-dynamics-setup
 >         end
 >
--- a/autogen/docs/systemdynamics.md.mustache
+++ b/autogen/docs/systemdynamics.md.mustache
@@ -312,12 +312,12 @@ a plot to watch the changes in sheep population.
 >
 >         to setup
 >           ca
->           set-current-plot "populations"
 >           system-dynamics-setup
 >         end
 >
 >         to go
 >           system-dynamics-go
+>           set-current-plot "populations"
 >           system-dynamics-do-plot
 >         end
 > - Move to the Interface tab
--- a/autogen/docs/dictionary.html.mustache
+++ b/autogen/docs/dictionary.html.mustache
@@ -1267,6 +1267,9 @@ end
           plotted values in the current plot, regardless of the current plot
           ranges.
         </p>
+        <p>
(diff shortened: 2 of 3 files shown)


autogen/docs/systemdynamics.md.mustache
```diff
@@ -312,6 +312,7 @@ a plot to watch the changes in sheep population.
 >
 >         to setup
 >           ca
+>           set-current-plot "populations"
```
LaCuneta: Did you consider putting the `set-current-plot` command in the `go` procedure right before the `system-dynamics-do-plot` call?  To me that seems like the safer place for it, to make sure it's always correct no matter what other commands are run.

brandesNW:
Hi Jeremy,
My thought was that for the purposes of the tutorial the logic of setting the value every tick would not be clear.
However, what you are suggesting is a safer practice, in case the user adds plots of their own.

I could change the code to
to setup
  ca
  system-dynamics-setup
end

to go
  system-dynamics-go
  set-current-plot "populations"
  system-dynamics-do-plot
end


--

LaCuneta:
> My thought was that for the purposes of the tutorial the logic of setting the value every tick would not be clear.

Gotcha, I can see that logic.  To me, I'd prefer the safer option as that'd be more likely to resolve the SDM error issue, but if you want to leave it as is, I'm fine with that.

brandesNW:
Hi Jeremy,
I can change it and make a new commit, since I don’t have a strong preference.

--

brandesNW:
Hi Jeremy,
I have pushed the change

--

PR #1829, opened 2020-03-31
correct 3D hatched turtle zcor. Fixes #1822

### LINKED ISSUE #1822: hatched turtle in 3D does not have zcor of parent
When a turtle with a non-zero z-coordinate hatches a turtle and gives it no commands, the new turtle has a z-coordinate of 0.
If there are commands, the z-coordinate is correct.

**Notes:**
If the hatched turtle is asked its zcor it gives the zcor of the parent, not its actual zcor.
However distancexyz as well as visual inspection after rotating the view make it clear
that it is incorrectly placed.

The bug does not appear when optimization is turned off. 
This suggests a bug in the code that is used when hatch is optimized in the case of no command arguments:
./netlogo-gui/src/main/prim/_hatchfast.scala

**Model Code**
This behavior can be verified using the code below
```
breed [walkers walker]
breed [points point]

to bug-hatch
  ca
  reset-ticks
  ask patch 0 0 10
  [
    sprout-walkers 1
  ]

  ask walkers [
    hatch-points 1
  ]
  print ""
  print "This demonstrates a bug in hatch."
  print "patch 0 0 10 sprouts a walker"
  print "the walker hatches a point"
  print "the point is not at the same location as the walker"
  ask walkers [ type "distance to 0 0 10 " show distancexyz 0 0 10]
  ask points [ type "distance to 0 0 10 " show distancexyz 0 0 10]
  print ""
  print "This can also be seen by rotating the view."
  print ""
  print "The z coordinate of the point does not print properly"
  ask walkers [ type "zcor"  show zcor ]
  ask points [ type "zcor"  show  zcor ]
end


to nobug-hatch
  ca
  reset-ticks
  ask patch 0 0 10
  [
    sprout-walkers 1
  ]

  ask walkers [
    hatch-points 1 [ set color yellow ]
  ]
  print ""
  print "This demonstrates that the bug in hatch does not occur when the hatched turtle has commands."
  print "patch 0 0 10 sprouts a walker"
  print "the walker hatches a yellow point"
  print "the point is at the same location as the walker"
  ask walkers [ type "distance to 0 0 10 " show distancexyz 0 0 10]
  ask points [ type "distance to 0 0 10 " show distancexyz 0 0 10]
end
```

brandesNW:
In NetLogo3D when a turtle with a non-zero z-coordinate hatches a
  turtle and gives it no commands, the new turtle had a z-coordinate of
  0. This was fixed by adding a hatch(TreeAgentSet breed) method to the
  Turtle3D class. Fixes #1822

 Note for those who wonder why this change fixed the bug:
 The _hatch primitive called hatch(), which had to be implemented in 
Turtle3D.java, because it is declared abstract in Turtle.java.
 On the other hand when there is no command after hatch, the optimized
 version _hatchfast is used. It calls hatch(TreeAgentSet breed). Since that
 is not abstract and not implemented in Turtle3D.java, the version in Turtle.java,
 which only sets the x and y coordinates was called.

### DIFF
--- a/netlogo-core/src/main/agent/Turtle.java
+++ b/netlogo-core/src/main/agent/Turtle.java
@@ -117,7 +117,7 @@ public abstract strictfp class Turtle
 
   // The observant reader will notice that the abstract methods are
   // primarily those which depend on the world arity (2D/3D).
-  public abstract Turtle hatch();
+  public abstract Turtle hatch(TreeAgentSet breed);
   public abstract Patch getPatchAtOffsets(double dx, double dy) throws AgentException;
   public abstract Patch getPatchAtHeadingAndDistance(double delta, double distance) throws AgentException;
   public abstract void jump(double distance) throws AgentException;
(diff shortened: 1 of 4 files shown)


brandesNW:
The code was refactored so that in Turtle.java hatch() now has an implementation that works for both 2D and 3D (it was formerly abstract), and hatch(treeAgentSetBreed) which had an implementation in Turtle.java that was 2D specific, is now abstract and is implemented in Turtle2D.java and Turtle3D.java.

PR #1837, opened 2020-05-01
Fix or delete NetLogo/bin scala scripts.

brandesNW:
Fix problems scripts found in other files.
  remove deprecated -nocompdaemon, use lazyLines
  Fix/update paths and jar file versions


bin/findbadmath.scala
```diff
@@ -42,15 +44,15 @@ for {
 
 // now do the StrictMath check
 
-for{path <- Process("find netlogo-core/src netlogo-gui/src -name *.java").lineStream
+for{path <- Process("find netlogo-core/src netlogo-gui/src -name *.java").lazyLines
     if !path.containsSlice("/gl/render/")  // we don't care if OpenGL stuff is strictfp
+    if !path.containsSlice("Activation.java") // Activation uses Math.max, on "safe" data
```
LaCuneta, line 49: I'm just curious - why did this file exclusion need to be moved?

brandesNW:
Hi Jeremy,

OK I’ll remove / April 2020 - AAB - remove deprecated -nocompdaemon, use lazyLines

Re bin/findbadmath.scala<https://urldefense.com/v3/__https:/github.com/NetLogo/NetLogo/pull/1837*discussion_r419547354__;Iw!!Dq0X2DkFhyF93HkjWTBQKhk!HKiM1iJfczegYtvbQR-xLjIPmtFQVpeRTj85DHPpGX7FiGPUNu9XuSzE-VfIiEk36g8GSH6aiA$>:
Activation.java is in the set of paths, and should be skipped.
It was in the   if(io.Source.fromFile(path).getLines which is searching through the lines in a file.
The string “Activation.java” will not necessarily be in the file Activation.java.

for{path <- Process("find netlogo-core/src netlogo-gui/src -name *.java").lazyLines
    if !path.containsSlice("/gl/render/")  // we don't care if OpenGL stuff is strictfp
    if !path.containsSlice("Activation.java") // Activation uses Math.max, on "safe" data
    if path != "src/org/nlogo/headless/TestCommands.java"}
  // this isn't the absolutely correct check to be doing, but it seems like a good enough heuristic
  // for now - ST 5/8/03
  if(io.Source.fromFile(path).getLines
     .filter(!_.containsSlice("Mathematica"))
     .filter(!_.containsSlice("DummyMath"))
     .exists(_.matches(""".*[^t]Math.*""")))
    println("needs StrictMath: " + path)

BTW I have updated benchmark statistics which could be added to benches.scala.
Currently it has NetLogo 4.0 & 4.1 numbers created by Seth
I could drop 4.0 and add 6.1.1 if you think it’s worthwhile.

- Aaron
--

bin/findbadmath.scala
brandesNW: Activation.java is in the set of paths, and should be skipped.
It was in the   if(io.Source.fromFile(path).getLines which is searching through the lines in a file.
The string “Activation.java” will not necessarily be in the file Activation.java.
 
for{path <- Process("find netlogo-core/src netlogo-gui/src -name *.java").lazyLines
    if !path.containsSlice("/gl/render/")  // we don't care if OpenGL stuff is strictfp
    if !path.containsSlice("Activation.java") // Activation uses Math.max, on "safe" data
    if path != "src/org/nlogo/headless/TestCommands.java"}
  // this isn't the absolutely correct check to be doing, but it seems like a good enough heuristic
  // for now - ST 5/8/03
  if(io.Source.fromFile(path).getLines
     .filter(!_.containsSlice("Mathematica"))
     .filter(!_.containsSlice("DummyMath"))
     .exists(_.matches(""".*[^t]Math.*""")))
    println("needs StrictMath: " + path)

bin/benches.scala
```diff
@@ -8,6 +8,10 @@ import sys.process.Process
 import java.io.File
 import collection.mutable.{ HashMap, ListBuffer, HashSet }
 import Ordering.Double.TotalOrdering
+<<<<<<< HEAD
+
+=======
+>>>>>>> a7cd2e90344ca0b84e0b876712639fe83b055264
```
SethTisue, line 14: oops

SethTisue:
follow up issue on some scripts that still don't necessarily work: #1851

brandesNW:
Hi,
I did the script updates, and had the ivy cache on my system.
I can do the quick and dirty fix.
We hadn’t used the scripts in a while (if for no other reason than that most of them weren’t working)
It would probably be good to have a cleaner fix before the next release.
Aaron

--

UriCCL:
Thanx Aaron. Marvelous.

Please also be careful to make sure all such messages get to the people I addressed in my messages: devel-core, forrest, seth. 
Some of those people get the GitHub messages, but not all, and it’s the most important to me that they get it right away.

Thanx,

—Uri


Uri Wilensky
Lorraine H. Morton chaired Professor 
Professor of Learning Sciences, Computer Science and Complex Systems
Director, Center for Connected Learning  and Computer-Based Modeling (CCL)
Northwestern University

Issue #1868, opened 2020-11-16
sbt script fails on MacOS Big Sur (11.0.1)


JamesHovet:
When running the sbt bash script from the root NetLogo directory on MacOS Big Sur (11.0.1), I get this error:
```
The operation couldn’t be completed. Unable to locate a Java Runtime that supports (null).
Please visit http://www.java.com for information on installing Java.

./sbt: line 87: /bin/java: No such file or directory
```

I was able to fix it, at least on my machine, by changing line 22 from:
```
export JAVA_HOME=`/usr/libexec/java_home -F -v1.8*`
```
to 
```
export JAVA_HOME=`/usr/libexec/java_home -F -v1.8`
```

I will submit a PR shortly with this change.

JamesHovet:
See PR #1869

PR #1869, opened 2020-11-16
Modify sbt script to work on MacOS Big Sur (11.0.1) (issue #1868)

JamesHovet:
See issue #1868 for the issue. 

Changing this one line seems to fix the issue for me and allowed me to run sbt and compile again. 

Looking around the web, the `/usr/libexec/java_home` behavior certainly was changed in Big Sur, (see https://developer.apple.com/forums/thread/666681 and https://dev.to/pablohs1986/macos-users-java-problems-in-big-sur-help-7pa) although I have not been able to find exactly how.

### DIFF
--- a/sbt
+++ b/sbt
@@ -19,7 +19,7 @@ else
     fi
   else
     if [ `uname -s` = Darwin ] ; then
-      export JAVA_HOME=`/usr/libexec/java_home -F -v1.8*`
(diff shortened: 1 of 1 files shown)


LaCuneta:
Confirmed things still work with the updated command on Catalina.  Thanks for the update!

PR #1876, opened 2020-12-01
fix codetab z-level issues by making CodeTabContainer a JFrame

brandesNW:


netlogo-gui/src/main/app/AppTabManager.scala
```diff
@@ -92,7 +92,7 @@ class AppTabManager(val appTabsPanel:          Tabs,
   def setDirtyMonitorCodeWindow(): Unit = {
     codeTabsPanelOption match {
       case None                => dirtyMonitor.setCodeWindow(None)
-      case Some(codeTabsPanel) => dirtyMonitor.setCodeWindow(Some(codeTabsPanel.getCodeTabContainer.asInstanceOf[JDialog]))
+      case Some(codeTabsPanel) => dirtyMonitor.setCodeWindow(Some(codeTabsPanel.getCodeTabContainer.asInstanceOf[JFrame]))
```
LaCuneta, line 95: Does this need to be cast to a `JFrame`?

netlogo-gui/src/main/app/DirtyMonitor.scala
```diff
@@ -38,8 +38,8 @@ with SaveModel.Controller
   private var loading = true
   private var _modelDirty = false
   private var priorTempFile = Option.empty[Path]
-  private var _codeWindow : Option[javax.swing.JDialog] = None
-  def setCodeWindow(codeWindow: Option[javax.swing.JDialog]): Unit = {
+  private var _codeWindow : Option[javax.swing.JFrame] = None
+  def setCodeWindow(codeWindow: Option[javax.swing.JFrame]): Unit = {
```
LaCuneta, line 42: I think `JFrame` is already imported in this file so we can drop the `javax.swing` part of the package name.

LaCuneta (changes requested):
I tested on Linux and macOS Big Sur and things definitely worked better.  I also found that the snap-to-side functionality also works after this change, letting me quickly put the interface and code side-by-side with system hotkeys.

I'm glad the z-ordering functionality could be provided/fixed with a relatively small change.  I had just a couple small issues I found in the code.

PR #1878, opened 2020-12-02
zoom component only if isShowing and window is focussed

brandesNW:


LaCuneta (changes requested):
This mostly looks good to me, just a couple small issues noted in the code.  I tested out on Linux, the code tab (docked and undocked), the info tab, and the interface tab / command console and everything worked as expected.

PR #1880, opened 2020-12-07
Model dirty marker only in main NetLogo window

brandesNW:


netlogo-gui/src/main/app/App.scala
```diff
@@ -754,16 +754,21 @@ class App extends
...
-
-    val title = e.newTab match {
-        case tab: TemporaryCodeTab => externalFileTitle(tab.filename.merge)
-        case _                     => modelTitle
-      }
-    frame.setTitle(title)
     tabManager.codeTabsPanelOption match {
-      case None           =>
+      case None           => {
+        val AppWindowTitle = e.newTab match {
```
LaCuneta, line 759: Prefer lower-case letters to start local variable names.

netlogo-gui/src/main/app/App.scala
```diff
@@ -794,10 +803,12 @@ class App extends
   def handle(e: LoadBeginEvent): Unit = {
     val modelName = workspace.modelNameForDisplay
     errorDialogManager.setModelName(modelName)
-    if(AbstractWorkspace.isApp) frame.setTitle(modelTitle)
-    tabManager.codeTabsPanelOption match {
-      case None           =>
-      case Some(thePanel) => thePanel.codeTabContainer.setTitle(modelTitle)
+    if(AbstractWorkspace.isApp) {
```
LaCuneta, line 806: Prefer space after the `if` before the opening paren.  I know this was in the existing code (or I missed it in a prior review), but either way if we're touching this line we should fix it up.

netlogo-gui/src/main/app/App.scala
```diff
@@ -869,10 +880,10 @@ class App extends
     if (dirty) s"* $title" else title
   }
 
-  private def modelTitle = {
+  private def modelTitle(allowDirtyMarker:Boolean = true) = {
```
LaCuneta, line 883: Prefer space after the colon and before the type.

LaCuneta (commented):
Looks good to me.  I tested the behavior and it is at least more consistent, if not working exactly as I would expect. 

Just a few small code items to clean up.

brandesNW:
Hi Jeremy,
Will make the changes.
What is not consistent? The rules that the marker is currently designed to follow are:
Model dirty markers should appear only in the main NetLogo window. They should remain visible when the CodeWindow is open.
The only dirty markers in the CodeWindow should be for included files.
When there is no CodeWindow, the title should be for the model unless the tab for an included file is open.

Let me know if they are not.
Implementing a dirty marker for a model title in the CodeWindow as well as in the main NetLogo window beyond the scope of the beta release.

I sent out a longish email presenting the current use of dirty markers which I hope will generate a discussion about future designs.

Aaron

--

PR #1897, opened 2021-02-02
Inspector jdialog 1859

brandesNW:


brandesNW:
Note:  I removed unneeded code from ColorDialog.java - this was discovered while making the code fix

Here is some code to create a turtle, patch and link inspector/monitor window.
globals  inspector/monitor windows require the Tools menue

LaCuneta (changes requested):
This looks pretty good to me, I love to see code get deleted.  I tested out the functionality on a two-monitor system and it all worked as expected.  The title bars were a little chunky on standard Ubuntu 20.04, but the windows were definitely usable.

I think you captured your screenshots with drop shadows (the macOS default for some reason).  I think we'd prefer not to have those in the docs as the rest of them don't have them and they waste some space:  https://www.maketecheasier.com/screenshots-without-drop-shadows-mac/

Also, on Ubuntu I noticed I only have a close button but your screenshots showed the usual macOS trio of close, hide, and "biggen".  Are those just disabled but visible on macOS?

This does not require a change, but I think your last commit message was a little backwards - it listed the issues it closed first and then a description afterwards.  Usually you want the "header" line to be descriptive, and you can include the issues it closes as extra on lines afterwards if they don't fit on a single 50 character line.  GitHub will still pick them up for closing if they are the "body".  Just for future reference.

PR #1898, opened 2021-02-03
Linkorder 1888

brandesNW:


test/commands/ComparingAgents.txt
```diff
@@ -2,12 +2,17 @@ ComparingLinks
...
+  directed-link-breed [directed-edges2 directed-edge2]
   O> crt 3 [ fd 5 ]
   O> ask turtle 0 [ create-directed-edges-to other turtles ]
   O> ask turtle 1 [ create-directed-edges-to other turtles ]
   O> ask turtle 0 [ create-undirected-edges-with other turtles ]
+  O> ask turtle 0 [ create-directed-edges2-to other turtles ]
+  O> ask turtle 0 [ create-undirected-edges2-with other turtles ]
   directed-edge 1 0 > directed-edge 0 1 => true
   undirected-edge 0 1 > directed-edge 0 1 => true
+  directed-edge2 0 1 > undirected-edge2 0 1 => true
```
LaCuneta, line 15: Should we also test the same-turtle, differently-breeded directed and undirected links against each other?  `directed-edge` versus `directed-edge2` and `undirected-edge` versus `undirected-edge2`?

LaCuneta (commented):
This looks mostly good to me, glad the fix wasn't too involved.  I had one question about adding a few more checks to the test.

LaCuneta:
I noticed this also had a checksum failure, probably because the link order changed:  https://netlogobuild.northwestern.edu/job/NetLogo/job/linkorder-1888/3/console

If you confirm the model would be affected by the change you can then regenerate the checksums for it and commit those. [The NetLogo wiki has some info on that[(https://github.com/NetLogo/NetLogo/wiki/Unofficial-features#model-checksums) in a [couple places](https://github.com/NetLogo/NetLogo/wiki/Preparing-the-Models-Library-for-release#update-the-models-checksums).

brandesNW:
Hi Jeremy,
I think I remarked that the display changed when I reported the issue.
I will change the checksum.
The probably has to do with z-ordering which I think is awkward for links.
Ideally users who want multiple links between a pair of turtles need to use shapes that don’t overlap.
Failing that it would probably be better if ordered links for display by the order they were created.
The current ordering depends on the who ordering of the turtle pair in the link (which the user can’t control for undirected links – I think) and link-breed order declaration is used to break ties.
Aaron

--

test/commands/ComparingAgents.txt
brandesNW: tests added

Issue #1899, opened 2021-02-03
Changing the breed of a link without any variable to a breed with a variable, causes an error.


Wolgo:
### The bug:

When I try to change the breed of a link w without any variables defined in breed-own, to a breed2 with a variable defined in breed2-own. I get the attached error message. 

When the breed does have *any* variable defined, changing to the breed2 does not give an error.
The attached zip file contains a NetLogo project with comments that showcases the bug. There are some pieces commented that I investigated. Please be aware that the error caused by this bug might not be gone after pressing setup (which uses clear-all), and that you might have to reload the model.

This bug is especially nasty because removing the offending code and rerunning still gives the error-message. The model even completely stops reacting to my buttons in some instances.

### The attached files:
[error-message.txt](https://github.com/NetLogo/NetLogo/files/5920515/error-message.txt)
[bugged_link_breed_change.zip](https://github.com/NetLogo/NetLogo/files/5920697/bugged_link_breed_change.zip)

### My diagnosis

When we look at the error I see it happening on:
`at org.nlogo.agent.Link.realloc(Link.java:175)
`
https://github.com/NetLogo/NetLogo/blob/341263d8fcd659505573a17827f2eefdd3f96fc2/netlogo-core/src/main/agent/Link.java#L167-L178

For each variable, we check the old position if it existed in the previous breed, so that we retain data.
Line 175 should only be reached when there was an old position for this variable. 
The variable we are changing to did not exist in the previous breed, thus we should never have reached line 175.
This means we got a wrong result for `oldpos`.

This means the error exists in the function `oldBreedsOwnIndexOf` 
https://github.com/NetLogo/NetLogo/blob/341263d8fcd659505573a17827f2eefdd3f96fc2/netlogo-core/src/main/agent/Link.java#L187-L197

Here is where my knowledge of the NetLogo is too lacking to exactly know where to look further. Because if we define any variable at all for the first breed, I suspect that `found.isEmpty()` does not return true when there are no "breed-own" defined variables. This is probably because the link does have variables such color defined. (I don't know what the intended behaviour is)

If the `found.isEmpty()` functions correctly, then `found.get().owns().indexOf(name)` is most likely not returning -1 when the variable does not exist.

Similar code is used for turtle, which does not suffer the same problem! 
https://github.com/NetLogo/NetLogo/blob/341263d8fcd659505573a17827f2eefdd3f96fc2/netlogo-core/src/main/agent/Turtle.java#L205-L236

[...]

PR #1900, opened 2021-02-07
Fix Link breed changes to compare to the old breed instead of the new breed. 

Wolgo:
Change link `setbreed` and `realloc` function to use and pass along the old breed into OldBreed correctly instead of getting the new breed by using getBreed();

### DIFF
--- a/netlogo-core/src/main/agent/Link.java
+++ b/netlogo-core/src/main/agent/Link.java
@@ -128,9 +128,12 @@ public strictfp class Link
 
   ///
 
-  @Override
+ @Override
   public Agent realloc(Program oldProgram, Program newProgram) {
-    TreeAgentSet oldBreed = getBreed();
+    return realloc(oldProgram, newProgram, null);
+  }
+
+  Agent realloc(Program oldProgram, Program program, AgentSet oldBreed) {
     boolean compiling = oldProgram != null;
 
     // first check if we recompiled and our breed disappeared!
@@ -644,7 +647,7 @@ public strictfp class Link
       _world.linkManager().addLink(this);
     }
     shape(_world.linkBreedShapes().breedShape(breed));
-    realloc(null, _world.program());
+    realloc(null, _world.program(), oldBreed);
   }
 
   // returns the index of the breed of this link, 0 means a generic link;


Wolgo:
Should fix issue: #1899

Wolgo (commented):
I realize now that this line is problematic as it will throw an error when used. Unlike breedsOwnIndexOf used in turtle, linkBreedsOwnIndexOf throws an error on a null input. 

I don't think anything actually uses the realloc function now which makes it not fail any tests.
(The function is needed because of the interface requiring a realloc with 2 parameters.)

SethTisue:
> I don't think anything actually uses the realloc function now 

hmm, just checking: are you sure? historically, `realloc` was used so that you didn't lose all your turtles (and in general, all your world state) if you e.g. added or removed a `turtles-own` variable in the Code tab.

brandesNW:
Hi Seth and others,
That comment by person making the PR is incorrect, but the code seems to be correct.
It is true that compilation makes use of realloc, as does set breed.

There was a difference between how Turtle realloc  and Link realloc worked which resulted in a bug because the Link version determined the oldBreed incorrectly. To make them consistent he created a three argument version of Link realloc that took oldBreed as an argument. The original realloc nows calls the three arg version with oldBreed = null.


The PR maker was afraid this would fail in the following code in Link.java

      int oldpos = compiling ? oldBreedsOwnIndexOf(oldProgram, getBreed(), name)
          : _world.linkBreedsOwnIndexOf(oldBreed, name);

Because linkBreedsOwnIndexOf explicitly throws an exception if the breed is null.

It took some investigation for me to determine that there is not be a problem because that code is always run in a compiling context.
If you (Seth) know why there is a check for compiling and how that variable could be false, please let me know.


He may also have been concerned because linkBreedsOwnIndexOf explicitly throws an exception if the breed is null

  def linkBreedsOwnIndexOf(breed: AgentSet, name: String): Int = {
    if (breed == null) throw new IllegalArgumentException("invalid breed")
    breedsOwnCache.getOrDefault(breed.printName + "~" + name, NegativeOneInt).intValue
  }

while the Turtle version would also fail, but doesn’t throw an explicit exception:

  def breedsOwnIndexOf(breed: AgentSet, name: String): Int =
    breedsOwnCache.getOrDefault(breed.printName + "~" + name, NegativeOneInt).intValue

Please add any insight you can recall from days past.

Best,
Aaron

--

SethTisue:
> If you (Seth) know why there is a check for compiling and how that variable could be false, please let me know.

ohhhhhh..... boy. I'm afraid I don't remember anything about that :-/

SethTisue:
> If you (Seth) know why there is a check for compiling and how that variable could be false, please let me know.

ohhhhhh..... boy. I'm afraid I don't remember anything about that :-/

(maybe it's just that it will be false when the Code tab is compiled for the first time? but then why wouldn't it be called `recompiling`? 🤷 )

Issue #211, opened 2012-09-12
Negative size values and stamp


turtlemaniac:
Turtles with negative size values are not visualized and inverted shape is stamped in case "stamp" primitive is used.
To reproduce:
crt 1[set heading 0 hatch 1[set size -1 jump 1 hatch 1 [jump 1 stamp]]]

System:
NetLogo 5.0.2 (July 27, 2012)
Extension API version: 5.0
Java HotSpot(TM) Server VM 1.6.0_33 (Sun Microsystems Inc.; 1.6.0_33-b03)
operating system: Windows XP 5.1 (x86 processor)
Scala version 2.9.2
Java heap: used = 8 MB, free = 11 MB, max = 494 MB

JOGL: (3D View not initialized)
OpenGL Graphics: (3D View not initialized)

Wolgo:
First some updates on the state of this bug:
Currently, Netlogo3D does show agents and stamps with negative sizes inverted.
Netlogo2D does not show agents with negative sizes but it does show stamps.

This bug is firstly caused by:
https://github.com/NetLogo/NetLogo/blob/1dc4331cfcacc3fcdfa8beb97ca5e58e018c040f/netlogo-gui/src/main/render/TurtleDrawer.java#L22

The intention of this code seems to be to not draw turtles that are too small to be rendered. It doesn't expect a negative size, and of course with a negative value we will be always deemed "too small". 

One possible fix is to change this line to: `if (Math.abs(turtle.size() * patchSize) >= MIN_PATCH_SIZE_FOR_TURTLE_SHAPES) {`
This would be consistent with Netlogo3D. The turtle would now be rendered inverted. 

Stamp works differently because it doesn't call the drawTurtle function but the drawTurtleShape function, and thus avoids the above line. 

https://github.com/NetLogo/NetLogo/blob/1dc4331cfcacc3fcdfa8beb97ca5e58e018c040f/netlogo-gui/src/main/render/TrailDrawerJ.java#L107

Changing this line to: `turtleDrawer.drawTurtle` also makes the stamp not render, but this is not in line with Netlogo3D.

It would be needed to introduce this size check for the 3D version in:
https://github.com/NetLogo/NetLogo/blob/1dc4331cfcacc3fcdfa8beb97ca5e58e018c040f/netlogo-gui/src/main/gl/render/Renderer.java#L308
It might be needed to do this separately for the stamp in 3D. (I don't know how to run the 3D version from my own code currently to test it.)

[...]

PR #1901, opened 2021-02-07
Proposal for solution to issue #211.

Wolgo:
Make minimal changes to make 2D NetLogo fall in line with 3D NetLogo concerning negatively sized turtles and their stamps.

Addresses issue #211.
Instead of removing the stamp, this change makes turtles with negative size display inverted (as 3D NetLogo does this.)

If removing the stamp is still preferred:
Removing the Math.abs() from this change would instead make the stamp not display as suggested as a solution.

### DIFF
--- a/netlogo-gui/src/main/render/TurtleDrawer.java
+++ b/netlogo-gui/src/main/render/TurtleDrawer.java
@@ -19,21 +19,21 @@ public strictfp class TurtleDrawer {
   public void drawTurtle(GraphicsInterface g, TopologyRenderer topology,
                          org.nlogo.api.Turtle turtle, double patchSize) {
     if (!turtle.hidden()) {
-      if (turtle.size() * patchSize >= MIN_PATCH_SIZE_FOR_TURTLE_SHAPES) {
-        drawTurtleShape(g, topology, turtle, patchSize);
-      } else {
-        topology.drawWrappedRect(g, org.nlogo.api.Color.getColor(turtle.color()),
-            0.0f, turtle.xcor(), turtle.ycor(), turtle.size(), patchSize, true);
-      }
-      if (turtle.hasLabel()) {
-        drawTurtleLabel(g, topology, turtle, patchSize);
-      }
+      drawTurtleShape(g, topology, turtle, patchSize);
+    }
+    if (turtle.hasLabel()) {
+      drawTurtleLabel(g, topology, turtle, patchSize);
     }
   }
 
   void drawTurtleShape(GraphicsInterface g, TopologyRenderer topology, org.nlogo.api.Turtle turtle, double patchSize) {
-    Drawable d = getShapeFromCacheOrCreateDrawable(turtle, patchSize, shapes.getShape(turtle));
-    topology.wrapDrawable(d, g, turtle.xcor(), turtle.ycor(), turtle.size(), patchSize);
+    if (Math.abs(turtle.size() * patchSize) >= MIN_PATCH_SIZE_FOR_TURTLE_SHAPES) {
+	  Drawable d = getShapeFromCacheOrCreateDrawable(turtle, patchSize, shapes.getShape(turtle));
+      topology.wrapDrawable(d, g, turtle.xcor(), turtle.ycor(), turtle.size(), patchSize);
+	} else {
+      topology.drawWrappedRect(g, org.nlogo.api.Color.getColor(turtle.color()),
+		0.0f, turtle.xcor(), turtle.ycor(), turtle.size(), patchSize, true);
+	}
   }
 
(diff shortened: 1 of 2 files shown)


SethTisue:
Y'all should do as you like of course, but the NetLogo 2D behavior is actually what we designed, back in the day, whereas I'm at least 90% sure the NetLogo 3D behavior is just a weird accident. I would suggest making 3D match the 2D behavior, rather than vice versa.

Wolgo:
The reason why I argued for matching the 3D behaviour is that the code that makes negatively sized turtles not render does not seem to intend to do that. (It looks like it should just not render turtles that are too small to render/be visible) As the decision was made for 3D that they should render that seemed more intentional. 

Regardless, if we don't want to render negative sizes, I would want to do that somewhere before the renderer, so that both 2D and 3D will be consistent in that without having to edit both. Not permitting negative sizes would be most logical to me, as I don't think there are many people who used negative sizes to invert their rendering anyways. 

Gotta figure out what I broke first though.

Wolgo:
Seems that in the kicked rotators model the stamps are < MIN_PATCH_SIZE_FOR_TURTLE_SHAPES, this causes the stamps to have a different shape than before. (Always a rectangle instead of sometimes a cross shape.)
In the picture below the six on the left are on 6.2 and the six on the right on this branch. 

image

I will move the comparison for MIN_PATCH_SIZE_FOR_TUTLE_SHAPES back to drawTurtle, removing both stamp and turtle can then still be done by adding a turtle.size() < 0 check in drawTurtleShape.

brandesNW:
Hi,

I echo the comment from Nicholas Payette in issue #211<https://github.com/NetLogo/NetLogo/issues/211>

‘My first impulse would be to just not stamp anything if turtle size is negative.’

He goes on to ask ‘Should negative turtle sizes even be allowed in the first place? (Answering "no" could potentially raise problems for existing models...)’

I appreciate your efforts Erik, however I think that the folks engaged in NetLogo development need to address the issue – should we allow negative sizes, and if so is either the current 2D or 3D implementation what we want - before further action can be taken on your pull request.

Aaron

--

Ron-Teller:
Hey, 

I happened by chance upon this request, and would like to add my thoughts:
I think that the 'size' variable should only reflect the size of the turtle, and not be allowed a negative value. 
Enabling it to have a negative value means that the size now not only represents the size of the turtle, but also the direction it is rendered. 

This breaks the single responsibility of the size variable, and leads to more difficult code, since naturally in order to increase the size of something, you increase the size variable (positive value). But if the size is negative, you now need to decrease the value, which is very confusing.

More so, if the size is negative, which direction does the image reverse? Horizontally or vertically? This information is not conveyed by the name of the variable 'size', nor should it be.
The size variable should just be size, and not allow a value below 0. 
If there is a need to enable image transformation such as reversing an image, then I think it would be better to create a new variable just for that, such as "transform-x" and "transform-y" (not sure these are the proper terms used for image rendering). 

An image that is rendered as it was created originally would have values "transform-x = 1" and "transform-y = 1". 
A horizontally reversed image would have the values "transform-x = -1" and "transform-y = 1". 
A stretched image (twice in length) could have the values "transform-x = 2" and "transform-y = 1". 

In order to keep former netlogo models from breaking that use negative sizes, if size was set to a negative value, then:
- set transform-x to -1 
- set size to the abs value.

Ron

Wolgo:
Currently, negatively sized turtles are just invisible (in 2D), so to stop old models from breaking we would just have to allow sizes below 0 and render them as 0. (Of course, that would change 3D to now not have an invert option). 

I personally don't think reversing images is the most useful functionality, and it would require quite a bit of change (don't forget the z-axis in 3D either), while it could be achieved by a user make inverted shapes and setting the turtle to have the inverted shape. 

I agree that size should probably not even be allowed to be below 0, especially as it's currently not different from setting size 0. (Besides the inverting in 3D and allowing inverted stamps.)

Erik

Wolgo:
As an additional argument for rendering negatively sized turtles for now (i think the design choice of whether we want this at all is a different issues) Is that the web version of NetLogo also does render negatively sized turtles (inverted). 

I think it's good to make all versions fall in line in what they do first, and if a change the behaviour is needed or wanted that we make that second.

(For those who want to try it in the web version: 
[The web version set to a model where you can easily see the turtles if you use the command below in the command center](http://www.netlogoweb.org/launch#http://www.netlogoweb.org/assets/modelslib/Sample%20Models/Art/Fireworks.nlogo) 
Using command: crt 1 [set size -10 stamp jump 10]
)

brandesNW:
Because there is no agreement that negative sizes should cause inversion in 2D, the this pull request closed.

Issue #243, opened 2012-11-30
BehaviorSpace does nothing if output file is open in Excel


ToonTalk:
I click run, I select an existing file that is open in Excel, I answer yes to the dialog if I want to override the file and then nothing happens -- no warnings, no runs.

Windows 7; NetLogo 5.0.3

Wolgo:
I tried this on windows 10, with the newest NetLogo 6.2. 
At least for me it now reports an error: 

![image](https://user-images.githubusercontent.com/15855950/107157016-0544ec00-6982-11eb-9cc0-97eb207f8630.png)

ERSUCC:
I made a potential fix for this issue, see pull request #2258. I have tested it and confirmed that it seems to have resolved the issue on Windows, and the issue still does not occur on Mac (M1). I don't have access to a dedicated Linux computer, so I can't speak to any Linux OS at the moment.

ERSUCC:
This was resolved by PR #2258.

PR #1902, opened 2021-02-07
Add try catch to show an error message when an IOException occurs while using BehaviourSpace.

Wolgo:
Essentially this changes the error message from:

image

to:

image

This is related to issue #243
I have no idea what the impact would be on Mac or Linux. But for windows, it seems to work. If Mac and Linux give any IOException it should come in a nice little window now instead of an error that asks you to report it.

### DIFF
--- a/netlogo-gui/src/main/lab/gui/Supervisor.scala
+++ b/netlogo-gui/src/main/lab/gui/Supervisor.scala
@@ -12,6 +12,7 @@ import org.nlogo.nvm.{EngineException, Workspace}
 import org.nlogo.workspace.{CurrentModelOpener, WorkspaceFactory}
 import org.nlogo.nvm.LabInterface.ProgressListener
 import org.nlogo.api.LogoException
+import java.io.{IOException}
 
 object Supervisor {
   case class RunOptions(threadCount: Int, table: Boolean, spreadsheet: Boolean, updateView: Boolean, updatePlotsAndMonitors: Boolean)
@@ -74,21 +75,31 @@ class Supervisor(dialog: java.awt.Dialog,
       val fileName = org.nlogo.swing.FileDialog.showFiles(
         workspace.getFrame, "Exporting as spreadsheet", java.awt.FileDialog.SAVE,
         workspace.guessExportName(worker.protocol.name + "-spreadsheet.csv"))
-      addExporter(new SpreadsheetExporter(
+      try{
+	    addExporter(new SpreadsheetExporter(
         workspace.getModelFileName,
         workspace.world.getDimensions,
         worker.protocol,
-        new java.io.PrintWriter(new java.io.FileWriter(fileName))))
+		new java.io.PrintWriter(new java.io.FileWriter(fileName))))
+	  }
+	  catch{
+		case e: IOException => failure(e); return
+      }
     }
     if(options.table) {
       val fileName = org.nlogo.swing.FileDialog.showFiles(
         workspace.getFrame, "Exporting as table", java.awt.FileDialog.SAVE,
         workspace.guessExportName(worker.protocol.name + "-table.csv"))
-      addExporter(new TableExporter(
+      try{
+	  addExporter(new TableExporter(
         workspace.getModelFileName,
         workspace.world.getDimensions,
         worker.protocol,
         new java.io.PrintWriter(new java.io.FileWriter(fileName))))
+	  }
+	  catch{
+		case e: IOException => failure(e); return
+      }
     }
     progressDialog.setUpdateView(options.updateView)
     progressDialog.setPlotsAndMonitorsSwitch(options.updatePlotsAndMonitors)


qiemem:
Are you sure it keeps outputting after the exception is thrown? An error is much preferred to silently dropping data. Indeed, that error means that netlogo can't continue to write to the file while it's open in another program. It seems like the real fix is to lock the file on windows while it runs (though I don't know anything about windows).

Wolgo:
I'm not entirely sure what you mean. Both before and after my change the behaviour space aborts the experiment (and thus never starts outputting). The difference is that instead of showing the first picture, which tells you that you should report this problem on GitHub, we show you an error that only tells you that the file can't be used because it is already open, which the user can fix by closing the file.

This mostly addresses when the file is open in Excel before we start the experiment. (And thus we can't get a lock, because it already is locked by Excel.) A quick test on my end shows that we already lock the file.

qiemem:
Ah! My apologies. I misread your commit message. Thanks for clarifying the lock file situation as well.

netlogo-gui/src/main/lab/gui/Supervisor.scala
```diff
@@ -12,6 +12,7 @@ import org.nlogo.nvm.{EngineException, Workspace}
 import org.nlogo.workspace.{CurrentModelOpener, WorkspaceFactory}
 import org.nlogo.nvm.LabInterface.ProgressListener
 import org.nlogo.api.LogoException
+import java.io.{IOException}
```
LaCuneta, line 15: We've switched to adding a space after the `{` and before the `}`, so `import java.io.{ IOException }`.  Also, prefer to move this above the internal `org.nlogo` imports, but below the package statement.

netlogo-gui/src/main/lab/gui/Supervisor.scala
```diff
@@ -74,21 +75,31 @@ class Supervisor(dialog: java.awt.Dialog,
       val fileName = org.nlogo.swing.FileDialog.showFiles(
         workspace.getFrame, "Exporting as spreadsheet", java.awt.FileDialog.SAVE,
         workspace.guessExportName(worker.protocol.name + "-spreadsheet.csv"))
-      addExporter(new SpreadsheetExporter(
+      try{
+	    addExporter(new SpreadsheetExporter(
```
LaCuneta, line 79: We use two spaces, never tabs for indentation.  Here and a couple other spots used tabs.

netlogo-gui/src/main/lab/gui/Supervisor.scala
```diff
@@ -74,21 +75,31 @@ class Supervisor(dialog: java.awt.Dialog,
...
         workspace.getModelFileName,
         workspace.world.getDimensions,
         worker.protocol,
```
LaCuneta, line 82: Make sure these statements are indented one level in from the `addExporter` start.

netlogo-gui/src/main/lab/gui/Supervisor.scala
```diff
@@ -74,21 +75,31 @@ class Supervisor(dialog: java.awt.Dialog,
       val fileName = org.nlogo.swing.FileDialog.showFiles(
         workspace.getFrame, "Exporting as spreadsheet", java.awt.FileDialog.SAVE,
         workspace.guessExportName(worker.protocol.name + "-spreadsheet.csv"))
-      addExporter(new SpreadsheetExporter(
+      try{
```
LaCuneta, line 78: Prefer a space before the `{`

netlogo-gui/src/main/lab/gui/Supervisor.scala
```diff
@@ -74,21 +75,31 @@ class Supervisor(dialog: java.awt.Dialog,
...
+	    addExporter(new SpreadsheetExporter(
         workspace.getModelFileName,
         workspace.world.getDimensions,
         worker.protocol,
-        new java.io.PrintWriter(new java.io.FileWriter(fileName))))
+		new java.io.PrintWriter(new java.io.FileWriter(fileName))))
+	  }
+	  catch{
```
LaCuneta, line 85: Prefer a space before the `{`

netlogo-gui/src/main/lab/gui/Supervisor.scala
```diff
@@ -74,21 +75,31 @@ class Supervisor(dialog: java.awt.Dialog,
...
+		case e: IOException => failure(e); return
+      }
```
LaCuneta, line 87: Make sure this closing brace is indented to the same level as the `try/catch` statements.

netlogo-gui/src/main/lab/gui/Supervisor.scala
```diff
@@ -74,21 +75,31 @@ class Supervisor(dialog: java.awt.Dialog,
...
+      try{
+	    addExporter(new SpreadsheetExporter(
         workspace.getModelFileName,
         workspace.world.getDimensions,
         worker.protocol,
-        new java.io.PrintWriter(new java.io.FileWriter(fileName))))
+		new java.io.PrintWriter(new java.io.FileWriter(fileName))))
+	  }
+	  catch{
+		case e: IOException => failure(e); return
```
LaCuneta, line 86: I think I see why the `; return` is here as this is copied from above.  If you're going to split things onto separate lines I'd prefer to drop the `;` and have the `failure(e)` and `return` on their own lines.  You could also do things are they are above in this file and keep it all on a single line, `catch { case e: IOException => failure(e); return }`

netlogo-gui/src/main/lab/gui/Supervisor.scala
```diff
@@ -74,21 +75,31 @@ class Supervisor(dialog: java.awt.Dialog,
...
+      try{
+	  addExporter(new TableExporter(
         workspace.getModelFileName,
         workspace.world.getDimensions,
         worker.protocol,
         new java.io.PrintWriter(new java.io.FileWriter(fileName))))
+	  }
+	  catch{
+		case e: IOException => failure(e); return
+      }
```
LaCuneta, line 102: Similar comments here as above, make sure indentation uses spaces, make sure things are properly aligned, and either split the catch block up or move it to one line.

LaCuneta (changes requested):
Thank you for your contribution!  I found mostly small code style issues in the PR, I left comments for those.  One additional change I'd request, since you're adding the `java.io.IOException` import, it'd be good to remove the fully-qualified reference on line 174, so `case ex: java.io.IOException =>` can just be `case ex: IOException =>`.  Once those items are resolved I think we can merge.

Let me know if you have questions on anything.

netlogo-gui/src/main/lab/gui/Supervisor.scala
LaCuneta: I should add, I know the other imports don't exactly match this style as they are older.  I'll leave it to you if you want to correct them, but we do add the spaces in new code.

LaCuneta:
I went ahead and made the changes and did a little more cleanup:  https://github.com/NetLogo/NetLogo/commit/8232012794418cf48f07a001a659a55696d7a710

PR #1908, opened 2021-03-23
Infrastructure: Update JOGL to latest 2.4 release candidate

EwoutH:
Updates JOGL to latest 2.4 release candidate.


LaCuneta:
Thanks for putting this pull request together.  One step that is missing (that an external contributor can't do) is bundling the JOGL native libraries into an archive to post on AWS for NetLogo to use via `NativeLibs.scala`.  That's why this build failed.  I went ahead an got the bundle put together for this RC and am re-running the build in Jenkins now.  If everything passes and my local tests work, too, I'll get this merged, and I'll let you know if I have problems otherwise.

I also put together [some documentation](https://github.com/NetLogo/NetLogo/wiki/JOGL) on the process, since I think that wasn't anywhere in the wiki before.

EwoutH:
Thanks for the effort and extensive explanation!

PR #1912, opened 2021-04-08
Shape3d

### LINKED ISSUE #1896: Custom 2D shapes in NetLogo 3D not appearing until restart
Several of my students encountered this bug this year -- if they created a new shape (with the turtle shapes editor) in NetLogo 3D, their turtles would still appear as the "default" turtle shape even after setting the shape, until after they closed NetLogo and restarted it.

brandesNW:
Closes #1896
The attached archive has models useful for testing the fix in NetLogo and NetLogo3D
[TestFix1896..zip](https://github.com/NetLogo/NetLogo/files/6279339/TestFix1896.zip)

### DIFF
--- a/netlogo-gui/src/main/app/App.scala
+++ b/netlogo-gui/src/main/app/App.scala
@@ -250,8 +250,6 @@ class App extends
   // all these guys get set in the locally block
   private var _workspace: GUIWorkspace = null
   def workspace = _workspace
-  private var _shapeChangeListener: ShapeChangeListener = null
-  def shapeChangeListener = _shapeChangeListener
   lazy val owner = new SimpleJobOwner("App", workspace.world.mainRNG, AgentKind.Observer)
   private var _tabs: Tabs = null
   private var _codeTabsPanel: CodeTabsPanel = null
@@ -364,7 +362,7 @@ class App extends
       }
     }
 
-    _shapeChangeListener = new ShapeChangeListener(_workspace, world)
+    ShapeChangeListener.listen(_workspace, world)
 
     pico.addComponent(new EditorColorizer(workspace))
 
--- a/netlogo-gui/src/main/app/App.scala
+++ b/netlogo-gui/src/main/app/App.scala
@@ -250,6 +250,8 @@ class App extends
   // all these guys get set in the locally block
   private var _workspace: GUIWorkspace = null
   def workspace = _workspace
+  private var _shapeChangeListener: ShapeChangeListener = null
+  def shapeChangeListener = _shapeChangeListener
   lazy val owner = new SimpleJobOwner("App", workspace.world.mainRNG, AgentKind.Observer)
   private var _tabs: Tabs = null
   private var _codeTabsPanel: CodeTabsPanel = null
@@ -361,6 +363,8 @@ class App extends
         model.withOptionalSection("org.nlogo.modelsection.modelsettings", Some(ModelSettings(snapOn)), Some(ModelSettings(false)))
       }
     }
+
+    _shapeChangeListener = new ShapeChangeListener(_workspace, world)
 
     pico.addComponent(new EditorColorizer(workspace))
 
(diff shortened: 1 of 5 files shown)


LaCuneta (changes requested):
Hey Aaron, thanks for fixing these issues.  The sample models were very helpful for testing.  It's a shame they aren't automated tests we can re-use in the future, but that's probably too much effort without enough reward for such a seldom-used and seldom-changing feature.  A few items:

In `App.scala` I see you added a `private var` and a `def` for the `new ShapeChangeListener()` instance, but I noticed that that instance isn't referenced anywhere else in code.  In peeking at the `ShapeChangeListener` class, I see it just sets up the listeners and nothing ever calls into it after that and it doesn't do anything else.  Could it be refactored into an `object ShapeChangeListener` with a `listen(workspace, world)` method that does the same work the class constructor used to do?  That way you can get rid of the instance storage in `App.scala` and make it clear it's just setting up event listeners and is not useful on its own.

In `ShapeManager.scala` I see you did switch the `checkQueue()` method to take in the shapes as needed, which is good.  I think it'd be worthwhile to go ahead and remove the `turtleShapeList` and `linkShapeList` instance variables from the `ShapeManager` class entirely, too.  That will require also updating the `addModelShapes()` and `addLinkShapes()` methods to also take in their respective shape lists when they are used in the constructor.  Overall I think not having those instance variables will avoid any confusion in the future that the `ShapeManager` has the latest shapes at all, as it does not anymore.

No modification needed here, but I see for the last commit you put the the `Fixes #1896` as the header text, and the more descriptive "Return default link shape if named shape not found" as the body text.  We prefer to have the more descriptive bit as the header so you can see what the commit does clearly in the logs.  You can add the "Fixes..." stuff in the body.  Also, you added `Closes #1896` to the pull request body text which will also cause the auto-close to happen.  You shouldn't really need to do both, one or the other is fine.

Let me know if you have any questions.

brandesNW:
Hi Jeremy,
Thanks for the comments. I will make changes tomorrow. I need to think through how the view2.5 extension will deal with shape changes to see if that has any implications for ShapeChangeListener. Currently the shapes are set in an init method that is called only once. So there will need to be a way to update the shapes, as well as a way of being notified that an update is needed.
Aaron

--


From: Jeremy B ***@***.***>
Reply-To: NetLogo/NetLogo ***@***.***>
Date: Thursday, April 8, 2021 at 3:14 PM
Cc: Aaron Andre Brandes ***@***.***>, Author ***@***.***>
Subject: Re: [NetLogo/NetLogo] Shape3d (#1912)


@LaCuneta requested changes on this pull request.

Hey Aaron, thanks for fixing these issues. The sample models were very helpful for testing. It's a shame they aren't automated tests we can re-use in the future, but that's probably too much effort without enough reward for such a seldom-used and seldom-changing feature. A few items:

In App.scala I see you added a private var and a def for the new ShapeChangeListener() instance, but I noticed that that instance isn't referenced anywhere else in code. In peeking at the ShapeChangeListener class, I see it just sets up the listeners and nothing ever calls into it after that and it doesn't do anything else. Could it be refactored into an object ShapeChangeListener with a listen(workspace, world) method that does the same work the class constructor used to do? That way you can get rid of the instance storage in App.scala and make it clear it's just setting up event listeners and is not useful on its own.

In ShapeManager.scala I see you did switch the checkQueue() method to take in the shapes as needed, which is good. I think it'd be worthwhile to go ahead and remove the turtleShapeList and linkShapeList instance variables from the ShapeManager class entirely, too. That will require also updating the addModelShapes() and addLinkShapes() methods to also take in their respective shape lists when they are used in the constructor. Overall I think not having those instance variables will avoid any confusion in the future that the ShapeManager has the latest shapes at all, as it does not anymore.

No modification needed here, but I see for the last commit you put the the Fixes #1896 as the header text, and the more descriptive "Return default link shape if named shape not found" as the body text. We prefer to have the more descriptive bit as the header so you can see what the commit does clearly in the logs. You can add the "Fixes..." stuff in the body. Also, you added Closes #1896 to the pull request body text which will also cause the auto-close to happen. You shouldn't really need to do both, one or the other is fine.

Let me know if you have any questions.

PR #1934, opened 2021-07-16
extract-rgb

IsaacNoel:
Updated the extract-rib primitive to accept more types of input

### DIFF
--- a/netlogo-gui/src/main/prim/etc/_extractrgb.java
+++ b/netlogo-gui/src/main/prim/etc/_extractrgb.java
@@ -26,7 +26,7 @@ public final strictfp class _extractrgb
         org.nlogo.api.Color.validRGBList(rgb,true);
       }
       catch(AgentException a){
-        throw new org.nlogo.nvm.RuntimePrimitiveException(context, this, "Color must be a number or a valid RGB List.");
+        throw new org.nlogo.nvm.RuntimePrimitiveException(context, this, "Color must be a number or a valid RGB(A) List.");
       }
       return rgb;
     }
--- a/netlogo-gui/src/main/prim/etc/_extractrgb.java
+++ b/netlogo-gui/src/main/prim/etc/_extractrgb.java
@@ -4,6 +4,7 @@ package org.nlogo.prim.etc;
 
 import org.nlogo.api.LogoException;
 import org.nlogo.api.AgentException;
+import org.nlogo.api.LogoListBuilder;
 import org.nlogo.core.LogoList;
 import org.nlogo.core.Syntax;
 import org.nlogo.core.SyntaxJ;
@@ -28,7 +29,11 @@ public final strictfp class _extractrgb
       catch(AgentException a){
         throw new org.nlogo.nvm.RuntimePrimitiveException(context, this, "Color must be a number or a valid RGB(A) List.");
       }
-      return rgb;
+      LogoListBuilder rgbNoAlpha = new LogoListBuilder();
+      for(int i = 0; i < 3; i++){
+        rgbNoAlpha.add(rgb.get(i));
+      }
+      return rgbNoAlpha.toLogoList();
     }
   }
 
--- a/netlogo-gui/src/main/prim/etc/_extractrgb.java
+++ b/netlogo-gui/src/main/prim/etc/_extractrgb.java
@@ -29,7 +29,7 @@ public final strictfp class _extractrgb
     }
     catch(AgentException a) {
       throw new org.nlogo.nvm.RuntimePrimitiveException(
-        context, this, I18N.errorsJ().getN("org.nlogo.prim.$common.invalidColor"));
+        context, this, I18N.errorsJ().getN("org.nlogo.prim.etc._extractrgb.invalidColor"));
     }
     LogoListBuilder rgbNoAlpha = new LogoListBuilder();
     for(int i = 0; i < 3; i++) {
--- a/netlogo-gui/src/main/prim/etc/_extractrgb.java
+++ b/netlogo-gui/src/main/prim/etc/_extractrgb.java
@@ -1,8 +1,8 @@
 // (C) Uri Wilensky. https://github.com/NetLogo/NetLogo
-
 package org.nlogo.prim.etc;
 
 import org.nlogo.api.LogoException;
+import org.nlogo.api.AgentException;
 import org.nlogo.core.LogoList;
 import org.nlogo.core.Syntax;
 import org.nlogo.core.SyntaxJ;
@@ -21,6 +21,12 @@ public final strictfp class _extractrgb
     catch(LogoException l){
       org.nlogo.api.Exceptions.ignore( l ) ;
       LogoList rgb = argEvalList(context, 0);
+      try{
+        org.nlogo.api.Color.validRGBList(rgb,true);
+      }
+      catch(AgentException a){
+        throw new org.nlogo.nvm.RuntimePrimitiveException(context, this, "Color must be a number or a valid RGB List.");
+      }
       return rgb;
     }
   }
--- a/netlogo-gui/src/main/prim/etc/_extractrgb.java
+++ b/netlogo-gui/src/main/prim/etc/_extractrgb.java
@@ -5,17 +5,24 @@ package org.nlogo.prim.etc;
 import org.nlogo.api.LogoException;
 import org.nlogo.core.LogoList;
 import org.nlogo.core.Syntax;
+import org.nlogo.core.SyntaxJ;
 import org.nlogo.nvm.Context;
 import org.nlogo.nvm.Reporter;
 
 public final strictfp class _extractrgb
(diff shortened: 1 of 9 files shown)


netlogo-gui/src/main/prim/etc/_extractrgb.java
```diff
@@ -1,21 +1,34 @@
 // (C) Uri Wilensky. https://github.com/NetLogo/NetLogo
-
```
brandesNW, line 2: Keep the blank line.

netlogo-headless/src/main/prim/etc/_extractrgb.java
```diff
@@ -1,17 +1,34 @@
 // (C) Uri Wilensky. https://github.com/NetLogo/NetLogo
-
```
brandesNW, line 2: Keep the blank line.

netlogo-headless/src/main/prim/etc/_extractrgb.java
```diff
@@ -1,17 +1,34 @@
...
+      return report_1(context, d);
+    }
+    catch(LogoException l){
+      org.nlogo.api.Exceptions.ignore( l ) ;
+      LogoList rgb = argEvalList(context, 0);
+      try{
+        org.nlogo.api.Color.validRGBList(rgb,true);
+      }
+      catch(AgentException a){
+        throw new org.nlogo.nvm.RuntimePrimitiveException(context, this, "Color must be a number or a valid RGB List.");
```
brandesNW, line 28: valid RGB(A) List

netlogo-gui/src/main/prim/etc/_extractrgb.java
```diff
@@ -1,21 +1,34 @@
...
+    catch(LogoException l){
+      org.nlogo.api.Exceptions.ignore( l ) ;
+      LogoList rgb = argEvalList(context, 0);
+      try{
+        org.nlogo.api.Color.validRGBList(rgb,true);
+      }
+      catch(AgentException a){
+        throw new org.nlogo.nvm.RuntimePrimitiveException(context, this, "Color must be a number or a valid RGB List.");
+      }
+      return rgb;
```
brandesNW, line 30: This should return only the first 3 elements of rgb. Alpha should not be included. The test also needs to be updated.

netlogo-headless/src/main/prim/etc/_extractrgb.java
```diff
@@ -1,17 +1,34 @@
...
+      }
+      return rgb;
```
brandesNW, line 30: This should return only the first 3 elements of rgb. Alpha should not be included. The test also needs to be updated.

netlogo-gui/src/main/prim/etc/_extractrgb.java
```diff
@@ -1,21 +1,34 @@
...
+      return report_1(context, d);
+    }
+    catch(LogoException l){
+      org.nlogo.api.Exceptions.ignore( l ) ;
+      LogoList rgb = argEvalList(context, 0);
+      try{
+        org.nlogo.api.Color.validRGBList(rgb,true);
+      }
+      catch(AgentException a){
+        throw new org.nlogo.nvm.RuntimePrimitiveException(context, this, "Color must be a number or a valid RGB List.");
```
brandesNW, line 28: valid RGB(A) List

brandesNW (changes requested):
Isaac, great job finding and making changes in all the relevant files.
There are some minor change requests. The one substantive one is that 
> extract-rgb  [215 50 41 100]
shouled be  [215 50 41], not  [215 50 41 100]

brandesNW (approved):

netlogo-gui/src/main/prim/etc/_extractrgb.java
```diff
@@ -3,19 +3,38 @@
 package org.nlogo.prim.etc;
 
 import org.nlogo.api.LogoException;
+import org.nlogo.api.AgentException;
+import org.nlogo.api.LogoListBuilder;
 import org.nlogo.core.LogoList;
 import org.nlogo.core.Syntax;
+import org.nlogo.core.SyntaxJ;
```
LaCuneta, line 10: I don't see this `SyntaxJ` import used anywhere.  Can it be removed?  Actually, I think `Syntax` above it could be, too, if you wanted.

netlogo-gui/src/main/prim/etc/_extractrgb.java
```diff
@@ -3,19 +3,38 @@
...
 
 public final strictfp class _extractrgb
     extends Reporter {
 
-
   @Override
   public Object report(final org.nlogo.nvm.Context context) throws LogoException {
-
-    return report_1(context, argEvalDoubleValue(context, 0));
+    try{
```
LaCuneta, line 19: Prefer space after keyword and before opening curly bracket.  There are other occurrences of this, but I'll just leave this comment.

netlogo-gui/src/main/prim/etc/_extractrgb.java
```diff
@@ -3,19 +3,38 @@
...
+      double d = argEvalDoubleValue(context, 0);
+      return report_1(context, d);
+    }
+    catch(LogoException l){
```
LaCuneta, line 23: Prefer spaces after opening bracket before opening paren and after closing paren and before opening curly bracket.  There are other occurrences of this, but I'll just leave this comment.

netlogo-gui/src/main/prim/etc/_extractrgb.java
```diff
@@ -3,19 +3,38 @@
...
+      org.nlogo.api.Exceptions.ignore( l ) ;
```
LaCuneta, line 24: Prefer no spaces around method arguments or before closing semicolon.

netlogo-gui/src/main/prim/etc/_extractrgb.java
```diff
@@ -3,19 +3,38 @@
...
+      LogoList rgb = argEvalList(context, 0);
+      try{
+        org.nlogo.api.Color.validRGBList(rgb,true);
+      }
+      catch(AgentException a){
+        throw new org.nlogo.nvm.RuntimePrimitiveException(context, this, "Color must be a number or a valid RGB(A) List.");
```
LaCuneta, line 30: Move the error message into the internationalization files, so they can be internationalized at some future point.  There are [two](https://github.com/NetLogo/NetLogo/blob/hexy/shared/i18n/Errors_en.txt) [spots](https://github.com/NetLogo/NetLogo/blob/hexy/netlogo-gui/project/autogen/i18n/Errors_en.txt) for this, oddly.  You can see they organize their "access strings" by prim package and name.  [Here is a sample usage](https://github.com/NetLogo/NetLogo/blob/hexy/netlogo-gui/src/main/prim/etc/_sublist.java#L23-L24).

Change the error message to be a little more helpful, something like "...number or a valid RGB(A) list with 3 or 4 number items with values from 0 to 255." to be explicit about what an "RGB(A) list" actually is.

LaCuneta (changes requested):
Thanks for your work on this!  Nice job tracking down all the required update locations.  Having the positive/negative language tests is also much appreciated.

My usual proviso when doing a PR review that works with old code in NetLogo or one of its extensions:  I'll be applying our [current style guide](https://github.com/NetLogo/NetLogo/wiki/Code-formatting) and coding standards.  You may see older code in the same or nearby file(s) that doesn't fit these standards, but our policy is all new and updated code going in should be at our currently expected quality.  It'll mostly be small, easily corrected stuff.

I see you did a `try`/`catch` around getting a `double` value and fell back to trying to get a `LogoList`.  My preference would be to just explicitly do the type check instead of relying on exceptions in what is not really an exceptional circumstance (user sent a list value). Something like this should work:

```java
    Object val = args[0].report(context);
    if (val instanceof Double) {
      Double d = (Double)val;
      // do double stuff...
    } else if (val instanceof LogoList) {
      LogoList rgb = (LogoList)val;
      // do list stuff...
    } else {
      // throw...
    }
```

You might wind up throwing exceptions in a couple of places, but if it's internationalized (see comments), that shouldn't be too big a deal.

I only left review comments on the GUI version, but updates would be needed in the headless version, too.  Let me know if you have any questions.

LaCuneta (changes requested):
Thanks for the I18N changes and other style work!  

I should've been clearer about one style point - we prefer spaces after statement keywords and before opening parens, so `if (..) {}`, `catch (..) {}`, and `for (..) {}`.  Once those changes are made, let me know and I think we can merge this.

SethTisue:
> the extract-rib primitive

So the LORD God caused the man to fall into a deep sleep, and while he slept, He took one of the man’s ribs and closed up the area with flesh. And from the rib that the LORD God had taken from the man, He made a woman and brought her to him.

LaCuneta (approved):

PR #1941, opened 2021-08-17
Fixes #1939

### LINKED ISSUE #1939: Behavior of `scale-color` with equal range bounds
This behavior is undocumented, but [`scale-color`](https://ccl.northwestern.edu/netlogo/docs/dictionary.html#scale-color) accepts equal bounds for its range specification. For example:

```
scale-color grey x 1 1
```
where both bounds are set to 1.

In such a case, passing an input that's less than the bounds will report 0 (black) and passing an input that's higher than the bounds will report 9.999 (white):

```
observer> show scale-color grey 0.9 1 1
observer: 0
observer> show scale-color grey 2 1 1
observer: 9.9999
```

If you pass in a number that's equal to the bounds, however, things go wrong. For example:

```
show scale-color grey 1 1 1
```

will give:

```
ERROR: Runtime error: math operation produced a non-number 
```

(That's in NetLogo 6.2; I haven't tried other versions.)

I would either:

- Disallow equal bounds completely, for consistency's sake, or
- Report the middle of the range (i.e., `grey` in the above example) when the input is equal to both bounds.

The latter option seems more sensible to me and less likely to break existing models.

***

Just to give a bit of context, I'm working on a model where we have something like:

```
let values [ variable ] of patches
ask patches [ set pcolor scale-color grey variable min values max values ]
```

If all values are equal, we get the aforementioned runtime error.

IsaacNoel:
Changing scale-color to match palette scaling prims

### DIFF
--- a/autogen/docs/dictionary.html.mustache
+++ b/autogen/docs/dictionary.html.mustache
@@ -7737,12 +7737,12 @@ show (runresult [ [a b] -&gt; a + b ] 10 5)
           <i>number</i>.
         </p>
         <p>
-          If <i>range1</i> is less than <i>range2</i>, then the larger the
+          If <i>range1</i> is less or equal to than <i>range2</i>, then the larger the
           number, the lighter the shade of <i>color</i>. But if <i>range2</i>
           is less than <i>range1</i>, the color scaling is inverted.
         </p>
         <p>
-          If <i>number</i> is less than <i>range1</i>, then the darkest shade
+          If <i>number</i> is less than or equal to <i>range1</i>, then the darkest shade
           of <i>color</i> is chosen.
         </p>
         <p>
--- a/autogen/docs/dictionary.html.mustache
+++ b/autogen/docs/dictionary.html.mustache
@@ -1007,9 +1007,7 @@ show approximate-hsb 180 57.143 76.863
         </p>
         <p>
           The color reported may be only an approximation, since the NetLogo
-          color space does not include all possible colors. (See the Programming Guide
-          information on approximate-hsb and approximate-rgb for an example that demonstrates how
-          these primitives are approximations).
+          color space does not include all possible colors.
         </p>
         <pre>
 show approximate-rgb 0 0 0
--- a/autogen/docs/dictionary.html.mustache
+++ b/autogen/docs/dictionary.html.mustache
@@ -7737,17 +7737,16 @@ show (runresult [ [a b] -&gt; a + b ] 10 5)
           <i>number</i>.
         </p>
         <p>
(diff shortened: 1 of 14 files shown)


autogen/docs/dictionary.html.mustache
```diff
@@ -7737,12 +7737,12 @@ show (runresult [ [a b] -&gt; a + b ] 10 5)
           <i>number</i>.
         </p>
         <p>
-          If <i>range1</i> is less than <i>range2</i>, then the larger the
+          If <i>range1</i> is less or equal to than <i>range2</i>, then the larger the
```
SethTisue: missing "than"

autogen/docs/dictionary.html.mustache
```diff
@@ -7737,17 +7737,16 @@ show (runresult [ [a b] -&gt; a + b ] 10 5)
           <i>number</i>.
         </p>
         <p>
-          If <i>range1</i> is less than <i>range2</i>, then the larger the
-          number, the lighter the shade of <i>color</i>. But if <i>range2</i>
+          When <i>range1</i> is less or equal to than <i>range2</i>, then the larger the
```
brandesNW, line 7740: should be
When <i>range1</i> is less than or equal to

brandesNW (approved):

autogen/docs/dictionary.html.mustache
```diff
@@ -1007,9 +1007,9 @@ show approximate-hsb 180 57.143 76.863
         </p>
         <p>
           The color reported may be only an approximation, since the NetLogo
-          color space does not include all possible colors. (See <a href="#approximate-hsb">approximate-hsb</a> for a description of what
-          parts of the HSB color space NetLogo colors cover; this is
-          difficult to characterize in RGB terms.)
+          color space does not include all possible colors. (See the Programming Guide
+          information on approximate-hsb and approximate-rgb for an example that demonstrates how
+          these primitives are approximations).
```
LaCuneta, line 1012: I don't see this example in the programming guide, just a quick note about the NetLogo color space not include all possible colors.  But if it does exist, or we're going to add it, or we're just talking about the existing text, then we should link directly to it from here.  [Example](https://github.com/NetLogo/NetLogo/blob/hexy/autogen/docs/dictionary.html.mustache#L847)

LaCuneta (changes requested):
Just one small issue in the docs changes I noticed.

autogen/docs/dictionary.html.mustache
brandesNW: I think the parenthetical remark should wait for Isaac's upcoming  PR for the Programming Guide, or he could add the link in the Prog Guide PR.

PR #1942, opened 2021-08-18
Fixes Oddities with command center Tools menu actions #1906

### LINKED ISSUE #1906: Oddities with command center Tools menu actions
I found these problems while trying to check for a different issue a user reported.  I've only tested on Linux, but I suspect the behavior will be similar on other systems.  I'm keeping them together because they seem small and related, but they could be split up if I'm wrong about that.

Oddity 1:

- Pop out the Code tab using the checkbox
- Click back to the Interface tab/window
- Check the Tools menu, see the hide/jump command center items are in the menu for the interface tab (as we would expect)
- Close popped out code tab by clicking the close button.  Do not click anywhere else in the code tab window to give it focus first.
- Check Tools again and the hide/jump command center items are no long in the menu and shortcuts do not work.

Oddity 2:

- Pop out the Code tab using the checkbox.
- Make sure the Code tab has focus - edit some code there, compile it.
- Hit Ctrl-1 to jump to the Interface tab
- Enter `100` or whatever in the command center (which should have focus) and run it just to run something.
- Check the Tools menu and you'll see duplicate entries for the hide/jump command center items.
- Sometimes I had to repeat the process a couple times to get it to work, but I did always get it to work.

brandesNW:
Make sure Tools-> Hide Command Center and Edit-> Undo menu items
appear appropriately.
Don't revoke or offer Actions in CodeTabsPanel.scala.
Eliminate unnecessary tab switching with InterfaceTab.
Set current tab when the new tab is already selected.

### DIFF
--- a/netlogo-gui/src/main/app/AbstractTabsPanel.scala
+++ b/netlogo-gui/src/main/app/AbstractTabsPanel.scala
@@ -42,7 +42,15 @@ abstract class AbstractTabsPanel(val workspace:           GUIWorkspace,
   def getAppJFrame() = { jframe }
   var fileManager: FileManager = null
   var dirtyMonitor: DirtyMonitor = null
-  var currentTab: Component = interfaceTab
+  var currentTab: Component = null
+
+  def setCurrentTab(tab: Component): Unit = {
+    currentTab = tab
+  }
+
+  def getCurrentTab(): Component = {
+    currentTab
+  }
 
   def initManagerMonitor(manager: FileManager, monitor: DirtyMonitor): Unit =  {
     fileManager = manager
@@ -58,4 +66,5 @@ abstract class AbstractTabsPanel(val workspace:           GUIWorkspace,
   }
 
   override def requestFocus() = { currentTab.requestFocus() }
+
 }
--- a/netlogo-gui/src/main/app/AppTabManager.scala
+++ b/netlogo-gui/src/main/app/AppTabManager.scala
@@ -64,7 +64,7 @@ class AppTabManager(val appTabsPanel:          Tabs,
 
   // might want access to these owner methods to be only in the app package
   // Need to carefully decide which methods are private. AAB 10/2020
-  def getCodeTabsOwner = {
+  def getCodeTabsOwner() : AbstractTabsPanel = {
     codeTabsPanelOption match {
       case None           => appTabsPanel
       case Some(theValue) => theValue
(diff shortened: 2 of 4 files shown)


brandesNW:
I tested with all the combinations of opening and closing the separate code tab on my MacOs system.
**Open a separate Code tab**

-  Click Code tab checkbox on
- Control-click with the mouse on the tab bar of the Code tab
- Command + Shift + W on Mac, Control + Shift + W on other operating systems

**Return a separated Code tab to the main NetLogo window** 
-  Click the window’s close button
-  Click Code tab checkbox on
- Control-click with the mouse on the tab bar of the Code tab
- Command + W on Mac, Control + W on other operating systems

Note: before the fix moving between the Info tab and a separated Code tab sometimes produced multiple Edit->Undo and Redo menus.

**Move between the tabs**

- Click the tab
- Use shortcuts, as shown in the Tabs Menu
- Choose options from the Tabs Menu

LaCuneta (changes requested):
Thanks for looking into this.

- Please re-word your commit message to start with a short description of the change and not the `"Fixes #1906"` text.  You can include the `"Fixes #1906"` in the first line if it's less than 72 characters, otherwise put it on a separate line in the body.
- Make sure you have a blank line between the first line "header" of your commit message and any extra "body" text.  GitHub formats the continuous lines okay, but git has trouble.
- Make sure pull request titles also indicate what the issue/change instead of just referencing an issue number.

Functionality issues:

Scenario 1: Open NetLogo, pop out the code tab (using the checkbox if it matters) and make sure the interface tab is selected and visible.  Click in the code area, type `Ctrl+/`.  The command center is hidden, but a `/` is also type in the code window.  If a command hotkey is triggered, I would expect no `/` to be added to the code area.

Also with this scenario `Ctrl+Shift+C` does not bring me to the command center from the code area.  That isn't the biggest deal, but it is a little weird for one of those hotkeys to work and the other to not.

Scenario 2: Open NetLogo, pop out the code tab (using the checkbox).  Then click on the Info tab in the main window.  Note that the command center menu items are still in the tools menu but neither of them does anything.  I'd expect them to be hidden or disabled if they are of no use on the current tab.  Note that they are hidden as expected if the code tab is docked (as they were in NetLogo versions prior to the pop-out code tab).  

Then with the Info tab still selected, close the code tab window, then click on the interface tab, then check the Tools menu and there are duplicates of the command center hotkeys.

Let me know if you have questions on anything.

brandesNW:
R.e. Scenario 1: Open NetLogo, pop out the code tab (using the checkbox if it matters) and make sure the interface tab is selected and visible try CMD+/  and Ctrl+Shift+C
I tested this on my MacOS system
When in the code tab, with the interface tab visible, no command hotkey is triggered for CMD+/ or Ctrl+Shift+C, nothing is typed in the code text area.
When in the interface tab, with the separate code tab visible both hotkeys work as expected.

I  think this should be the expected behavior.

brandesNW:
These are not reproducible on my MacOS system

Scenario 2: Open NetLogo, pop out the code tab (using the checkbox). Then click on the Info tab in the main window.
On my MacOS system neither the Hide or Jump commands appears in the menu, as expected.

Then with the Info tab still selected, close the code tab window, then click on the interface tab, then check the Tools menu.  A single copy of the Hide and Jump commands is in the Tools menu

LaCuneta (approved):
Confirmed all test scenarios worked as expected (except the Ctrl-/ in the code tab, but that's less critical).

PR #1943, opened 2021-08-19
Programming guide edits

IsaacNoel:


brandesNW (approved):

autogen/docs/programming.md.mustache
```diff
@@ -368,14 +367,40 @@ Since many colors are missing from the NetLogo color space, [[approximate-hsb]]
...
 they try to come as close as possible.
 
-Example: you can change any turtle from it's existing NetLogo color to a half
-transparent version of that color using:
+Example:
 
-    set color lput 125 extract-rgb color
+    show extract-rgb approximate-rgb 0 0 255 ;; shows [48 88 161]
+
+This is an approximation, but it is still near NetLogo blue 104.7.
```
LaCuneta, line 374: This example was a little confusing for me.  Maybe split it into two steps?  `let my-color approximate-rgb 0 0 255 ; my-color is 104.7`, then the next line `show extract-rgb my-color`, and in the explanation below add a line about how (48, 88, 161) is pretty far from the starting (0, 0, 255) RGB color (pure blue)?

LaCuneta (changes requested):
In general we'd want to avoid giving an indication that using turtle `who` numbers is a good idea.  Can we switch the examples from `ask turtle turtle-number ...` to `ask one-of turtles ...` ?

Also, we should link to the Palette extension docs from the section about it.  [Here is an example of the Vid extension being linked to].

autogen/docs/programming.md.mustache
```diff
@@ -368,14 +367,41 @@ Since many colors are missing from the NetLogo color space, [[approximate-hsb]]
...
 
+**Palette Extension**
+
+The palette extension offers primitives that give the user more control over colors. It allows for direct
+manipulation of RGBA and HSB components without the requiring list manipulations.
+
+Example: changing the transparency of any turtle to sixty percent transparent with the palette extension:
+
+    extensions [palette]
+    ask turtle turtle-number [palette:set-transparency 60]
```
LaCuneta, line 397: One more `who` number example to change here.

autogen/docs/programming.md.mustache
```diff
@@ -368,14 +367,41 @@ Since many colors are missing from the NetLogo color space, [[approximate-hsb]]
...
+
+The extension offers similar commands for Red, Green, Blue, Alpha, Hue, Saturation, and Brightness.
+
+See the [documentation](https://ccl.northwestern.edu/netlogo/docs/palette.html) of the palette extension for more information.
```
LaCuneta, line 401: Thanks for adding the link, but it should be relativized for use by the offline docs as well.  I see I forgot to paste the example into my original comment, sorry about that!  https://github.com/NetLogo/NetLogo/blob/hexy/autogen/docs/programming.md.mustache#L1973

LaCuneta (changes requested):

LaCuneta (approved):

PR #1946, opened 2021-08-31
remove command center related accelerators from separate code tab

brandesNW:
When in a separate code tab the accelerators for "Hide Command Center" (cmd-/ or ctrl-/) and "Jump to Command Center" (cmd-shift-c or ctrl-shift-c) no longer exist.


LaCuneta (approved):
It's a little odd to me to remove the hotkeys after copying rather than avoiding the copy in the first place, but the fix is simple enough.  Tested the scenarios here and from the prior PR on Windows, things worked fine.

PR #1975, opened 2021-11-03
CI: Use multiple dependent jobs to run tests in parallel

EwoutH:
This commit splits the test into multiple jobs that can be executed in parallel. It hugely speeds up build time and improves the overview which tests are failing.

A cache is used to transfer the build to the tests jobs. The workflow configuration now looks like this:

Screenshot_653

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -26,7 +26,7 @@ jobs:
           java-version: liberica@1.8=tgz+${{ env.LIBERICA_URL }}
 
       - name: Clone Git submodules
-        run: git submodule update --init --depth 1
+        run: git submodule update --init
 
       - name: Build NetLogo
         run: |
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -9,31 +9,111 @@ env:
   LIBERICA_URL: https://download.bell-sw.com/java/8u292+10/bellsoft-jdk8u292+10-linux-amd64-full.tar.gz
 
 jobs:
-  build-and-test:
+  build:
     runs-on: ubuntu-latest
     if: ${{ 'pull_request' != github.event_name || (github.event.pull_request.head.repo.git_url != github.event.pull_request.base.repo.git_url) }}
     steps:
       - name: Cancel Previous Runs
-        uses: styfle/cancel-workflow-action@0.9.0
+        uses: styfle/cancel-workflow-action@0.9.1
         with:
           access_token: ${{ github.token }}
 
       - uses: actions/checkout@v2
 
-      - uses: olafurpg/setup-scala@v10
+      - uses: olafurpg/setup-scala@v13
+        id: setup-scala
+        with:
+          java-version: liberica@1.8=tgz+${{ env.LIBERICA_URL }}
+
+      - name: Clone Git submodules
+        run: git submodule update --init --depth 1
+
+      - name: Build NetLogo
+        run: |
+          ./sbt update all headless/compile parserJS/compile
+          ./sbt depend headless/depend
+
+      - uses: actions/cache@v2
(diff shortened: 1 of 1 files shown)


SethTisue:
This is probably orthogonal to what you're doing, but as an aside, you don't need setup-scala, you can just use actions/setup-java:

https://twitter.com/olafurpg/status/1449448553068339205

EwoutH:
Thanks for the head-up!

For some reason the `org.nlogo.headless.misc.TestChecksums` test is consistently failing. Is that related to (the same problem fixed in) https://github.com/NetLogo/NetLogo/commit/3a69f11d8a485186cb2659f2c63c2c1979d4bd3c?

LaCuneta:
> For some reason the org.nlogo.headless.misc.TestChecksums test is consistently failing. Is that related to (the same problem fixed in) 3a69f11?

I don't believe it's related to that.  I double-checked and the checksum tests were running and passing on [the most recent commit to hexy](https://github.com/NetLogo/NetLogo/runs/4094774540), and I ran them locally and things were good as well.

It seems to me like the checksums are being run against the models library that was updated to add version 6.2.1 in the nlogo files, but the NetLogo repo itself is still pointing to the 6.2.0 models.  Or some other reason the model versions would be different than what the checksums were originally run against.  I can't see a way this PR would change how the models library gets setup.  I'll do a little more poking around.

LaCuneta:
I was mistaken on the versions being the issue, it looks like it's something to do with the actual git revision that the model last changed with.  I'm still not sure how this PR would affect that, but that is the issue - the headless checker is seeing every model as changed with the commit `e1f00779ac`, but [that commit just changed the River Meanders model](https://github.com/NetLogo/models/commit/e1f00779ac4bc23b0ae89daf69b8d5daf06c1238).  So maybe something about the caching process is messing with the models submodule git history?  I would think that would also cause the 2D test to fail the same way, though.

EwoutH:
Thanks for investigation. I tried a [bunch of different ways](https://github.com/EwoutH/NetLogo/actions?query=branch%3Aci-parallel-git) to clone and cache the extensions and submodules, but none worked so far unfortunately.

One compromise could be to leave the headless tests in the main build job, and run the other tests in separate jobs. Then for the headless job caching isn't required, but I would take a lot longer, which isn't ideal, since the other jobs are waiting on it.

LaCuneta:
That's very strange!  If it does work with headless tests back with the build, we can get this merged and fixed up later.  It's definitely a nice improvement and worth getting working properly.

I'm still stumped on how the 2D tests are passing as they contain a nearly identical test that uses [the same code to check git revisions that the headless one does](https://github.com/NetLogo/NetLogo/blob/hexy/netlogo-core/src/main/headless/ChecksumsAndPreviews.scala#L235-L246).

EwoutH:
[Fixed it](https://github.com/EwoutH/NetLogo/actions/runs/1418637489)!

To reduce the cache size, I added the `--depth 1` parameter to the submodule cloning (`git submodule update --init`). It reduces the cache size from ~258 MB to around ~190 MB. Since I didn't want to burden GitHub with unnecessary storage and I/O load (and only 5 GB of cache is kept by GitHub), it seemed very useful. But apparently doing a shallow clone resulted in (some) changed checksums for the extensions.

Anyway, I reverted to cloning the submodules with full depth and now it passes. At some point it would be worth investigating if we can do shallow submodule clones with checksums we can validate.

LaCuneta (approved):

LaCuneta:
Nice work!  I can live with the extra space usage, I think that's fine.  I'll also be adding a similar setup to our Tortoise project, where it should help us out even more, so many thanks for thinking of this and contributing it.

PR #1993, opened 2021-12-07
Codetab hotkeys

brandesNW:


LaCuneta (approved):
Looks okay to me.  Tested on Windows with the code tab popped in and out and back in and everything worked as expected.

PR #2011, opened 2022-01-28
Add methods to BrowserLauncher.scala to help replace deprecated openURL

brandesNW:
Update BrowserLauncher in preparation for replacing uses of the deprecated method openURL.
Added method openURIString, and versions of openURI, and openPath with different signatures.

Note that the calls that need to be updated are in the behaviorsearch directory, so this needs to be integrated before they can be updated.

Once the calls are updated, the following method can be removed
```
  @deprecated("Use openURI or openPath instead", "6.0.2")
  def openURL(comp: Component, urlString: String, local: Boolean): Unit = {
```


LaCuneta (approved):
LGTM.

PR #2024, opened 2022-04-03
Sdm better gui

ocaml-n00b:
Updates mostly **/sdm/gui** to add:
1. The use of right mouse button for adding connections between components.
2. An input drop-down in editor for viewing current inputs to selected figure.

It also sets the scene for adding input name to expression from drop-down
and a test button that checks model / gui correspondence (all inputs used in expressions, etc).

### DIFF
--- a/netlogo-gui/src/main/sdm/gui/ConverterFigure.java
+++ b/netlogo-gui/src/main/sdm/gui/ConverterFigure.java
@@ -9,9 +9,11 @@ import org.jhotdraw.standard.HandleEnumerator;
 import org.jhotdraw.standard.NullHandle;
 import org.jhotdraw.standard.RelativeLocator;
 import org.nlogo.api.Property;
-
 import java.util.ArrayList;
-import java.util.List;
+
+import org.nlogo.api.Options;
+import scala.collection.immutable.List;
+import scala.collection.JavaConverters;
 
 public strictfp class ConverterFigure extends DiamondFigure
     implements
@@ -25,6 +27,7 @@ public strictfp class ConverterFigure extends DiamondFigure
         (FigureAttributeConstant.FILL_COLOR,
             org.nlogo.window.InterfaceColors.SLIDER_BACKGROUND);
     converter = new org.nlogo.sdm.Converter();
+    converter.setSelected("Select");
   }
 
   public org.nlogo.sdm.ModelElement getModelElement() {
@@ -82,7 +85,7 @@ public strictfp class ConverterFigure extends DiamondFigure
   // Return no resize handles
   @Override
   public HandleEnumeration handles() {
-    List<NullHandle> handles = new ArrayList<NullHandle>();
+    java.util.List<NullHandle> handles = new ArrayList<NullHandle>();
     handles.add(new NullHandle(this, RelativeLocator.southEast()));
     handles.add(new NullHandle(this, RelativeLocator.southWest()));
     handles.add(new NullHandle(this, RelativeLocator.northEast()));
@@ -108,7 +111,7 @@ public strictfp class ConverterFigure extends DiamondFigure
     return scala.Option.apply(null);
   }
 
-  public List<Property> propertySet() {
+  public java.util.List<Property> propertySet() {
     return Properties.converter();
   }
 
@@ -144,4 +147,34 @@ public strictfp class ConverterFigure extends DiamondFigure
     return converter.getExpression();
   }
 
+  public void inputs(scala.collection.immutable.List<String> ls){
+    converter.setInputs(ls);
+  }
+
+  public Options<String> inputsC(){
+    Options<String> inputs = new Options<String>();
+    scala.collection.Iterator<String> ins = converter.getInputs().iterator();
+    inputs.addOption("Select","null");
+
+    while(ins.hasNext()){
+      inputs.addOption(ins.next(),"A");
+    }
+
+    String name = converter.getSelected() ;
+    if (name == ""){
+      name = inputs.names().head();
+      inputs.selectByName(name);
+      converter.setSelected(name);
+    }
+    else {
+      inputs.selectByName(name);
+    }
+    return inputs;
+  }
+  public void inputsC(Options<String> inList){
+    String name = inList.chosenName();
+    converter.setSelected(name);
+    //TODO: onChange
+   // converter.setExpression( converter.getExpression()+name+" " );
+  }
 }
--- a/netlogo-gui/src/main/sdm/gui/InspectionTool.java
+++ b/netlogo-gui/src/main/sdm/gui/InspectionTool.java
@@ -8,7 +8,8 @@ strictfp class InspectionTool
 
   private final AggregateModelEditor editor;
   private Model model;
-  public RateConnectionTool rct;
+  private RateConnectionTool rct;
+  private AggregateConnectionTool link;
 
   InspectionTool(AggregateModelEditor editor, Model model) {
     super(editor);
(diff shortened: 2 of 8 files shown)


LaCuneta (changes requested):
Thank you very much for this contribution!  The SDM doesn't get a lot of attention, so improvements are most welcome.

The code mostly looks good to me.  I left some style notes in the code itself.  I like to point out that a lot of the NetLogo code is old and was written without much aattention to style, so you may find code in the files you're working on that doesn't adhere to the items I listed.  We try to make sure all new code submitted improves the situation.

I like the idea of the inputs listing, and I see that you intend to make an auto-insert feature for it.  I wonder if a dropdown is the right UI element to use, though?  When viewing the inputs you probably want to see all of them, as opposed to one at a time.  Maybe a `JList` could work?  Adding the button to "insert" the selected one would be most welcome.  If you didn't want to bother with the complexity of the list or the insert, a read only text box might be enough to allow manual copy and pasting of the desired inputs, with the code creating a single string representation of all the inputs.

When I right-click on a stock and start dragging, it shows me it's making a flow, which is fine.  But if I drop that flow on a variable, it turns into a link.  I think that's okay, but maybe the in-progress dragging flow should change to a link when I hover over a variable, so I know what dropping will do?

Some documentation updates would be much appreciated.  [Here is where the docs live in the repo in Markdown format](https://github.com/NetLogo/NetLogo/blob/hexy/autogen/docs/systemdynamics.md.mustache).

ocaml-n00b:
Thank you for the feedback, much appreciated.

That sounds reasonable, I'll clean those along with a few others I found and do a 
better job at staying on top of the code style in the future.

As for the Inputs, yeah, the original intent was to have a list but I had issues
with the LogoListString Property that seemed like the best fit so I moved to the
dropdown to keep it moving as I get more acquainted with the project (and java/scala).
I will try and go back to using a list when I get that figured out.
For inserting I am currently testing with on dropdown change 
(or double clicking on item when list is added but button would be nice as well)

Showing the correct link/flow makes sense; I'll see if I can get that figured for a future push.

I didn't even think of the docs, I'll take a look for where it may make sense to update/change.

ocaml-n00b:
The code style issues have been fixed and Docs updated.
Not sure if I should add the **insert** feature on this pull request.
What are your thoughts? 
If I can leave for afterwards then it's ready for review.

ocaml-n00b:
Ready for review :)

netlogo-gui/src/main/sdm/gui/RateConnection.java
```diff
@@ -356,4 +359,31 @@ public boolean bivalentWrapper() {
...
   }
 
+  public void inputsC(Options<String> inList) {
+    rate.setSelected(inList.chosenName()); // not needed
+  }
+
+  public Options<String> inputsC() {
+    Options<String> inputs = new Options<String>();
+    inputs.addOption("Select", "null");
+    //String name = rate.getSelected(); // not needed for our case
```
LaCuneta, line 369: We can remove the commented code if not needed.

netlogo-gui/src/main/sdm/gui/RateConnection.java
```diff
@@ -356,4 +359,31 @@ public boolean bivalentWrapper() {
...
+    inputs.selectByName("Select");
+
+    scala.collection.Iterator<String> ins = rate.getInputs().iterator();
+    while (ins.hasNext()) {
+      inputs.addOption(ins.next(), "A");
```
LaCuneta, line 374: Since we're not using the value parameter, I think we should either make it more meaningful, like `"unused"` instead of `"A"`, or alternatively add a comment explaining what it's for.

netlogo-gui/src/main/sdm/gui/InspectionTool.java
```diff
@@ -1,33 +1,114 @@
...
     super(editor);
     this.editor = editor;
+    this.model = model;
   }
 
   protected void inspectFigure(org.jhotdraw.framework.Figure f) {
     editor.inspectFigure(f);
   }
 
+  public void fillInputs(Figure figure, org.jhotdraw.framework.DrawingView view) {
```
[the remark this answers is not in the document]
LaCuneta: I think this import can still be added.

netlogo-gui/src/main/sdm/gui/ConverterFigure.java
```diff
@@ -144,4 +147,34 @@ public String expressionWrapper() {
...
+    else {
+      inputs.selectByName(name);
+    }
+    return inputs;
+  }
+  public void inputsC(Options<String> inList) {
+    String name = inList.chosenName();
+    converter.setSelected(name);
+    //TODO: onChange
+   // converter.setExpression( converter.getExpression()+name+" " );
```
LaCuneta, line 178: Can this comment be removed now?

LaCuneta (changes requested):
This is very close to merging, thank you so much for working on it.  

My apologies for not replying sooner, what I was going to say was that to merge I'd want the new feature (inputs list) to be usable on its own.  With your latest update you can use the dropdown to add the input to the expression, so that definitely qualifies. 

The only items remaining were small issues in the code, I left comments on them.  Once they're fixed we should be good to go.

LaCuneta (approved):
Looks good to me!  I'll merge once the Actions tests pass.

LaCuneta:
Hey thanks for trying to clear the dependency error.  I should've noticed that in my initial review, sorry I missed it.  I'll go ahead and clear it up and merge things; it just requires moving some of the code around to avoid adding dependencies between some of the projects.

ocaml-n00b:
Thanks for the update. 
Yeah, it makes sense that it would complain about the circular dependency created by this.
Originally I tried to find somewhere in the SDM code to add the changes but could not find where,
it seems that part is automatically generated.
Let me know if there is anything I can help with.

LaCuneta:
Sorry this took me so long to get fixup up and merged.  Thanks once again for your work on it!  And I love your username, by the way!

EwoutH:
Currently JHotDraw 6 beta (`6.0b1`) is [used](https://github.com/NetLogo/NetLogo/blob/afab18b34b2ac50f61e83dc1415db74d8f260d22/build.sbt#L188) in NetLogo, which was released in 2004. Would is be worth it to update the System Dynamics (SDM) interface to [JHotDraw 7](http://www.randelshofer.ch/oop/jhotdraw/index.html) ([GitHub](https://github.com/wrandelshofer/jhotdraw)) or the (experimental?) [JHotDraw 8](https://github.com/wrandelshofer/jhotdraw8)?

> ### JHotDraw 7
> JHotDraw 7 is a two-dimensional graphics framework for structured drawing editors. It can be used to create many different editors from a simple doodle program up to full fledged diagram editors and vector animation tools.
> 
> JHotDraw 7 is a major rewrite of earlier versions of [JHotDraw](http://www.jhotdraw.org/) which in turn is a derivation of [HotDraw](http://st-www.cs.uiuc.edu/users/brant/HotDraw/HotDraw.html).
> 
> #### Features
> 
> - All contracts in the JHotDraw 7 framework are implemented against Java Interfaces. The framework provides many points to hook in your own code.
> - The framework is small enough to create Java applets with it.
> - Abstract and Default implementations are provided to get you a quick start.
> - The framework makes use of common Java idioms, such as the JavaBeans setter and getter methods. This is supposed to reduce your learning curve.
> - The drawing code takes advantage of Java2D. It uses double precision coordinates for all drawings.
> - A drawing can be shown and edited in multiple windows with different scale factors.
> - Undo/Redo is implemented based on the Swing undo package.
> - Cut, Copy and Paste clipboard operations are supported.
> - XML is used as the default external storage format and for clipboard operations.

ocaml-n00b:
Sure, I'll give it a try.

PR #2064, opened 2022-11-16
CI: Update Liberica JDK to 17.0.5+8

EwoutH:
Update the Liberica JDK from 17.0.3+7 to 17.0.5+8.

Also use the regular JDK, which is smaller, instead of the full one.

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -6,7 +6,7 @@ on:
   workflow_dispatch:
 
 env:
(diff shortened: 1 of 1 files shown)


brandesNW:
We usually don't update the JDK until we're doing release preparation.
Since this is a *Critical Patch Update release* I will integrate it if you switch back to the full version. We currently need the full version for the release process, so would need it if we integrate packaging into our CI.

EwoutH:
Done!

Please squash when merging.

PR #2068, opened 2022-11-28
Feature: Set default max heap size to percent instead of hard limit

LaCuneta:
~~Java 17 can handle allocating memory as necessary, so we no longer have to specify a maximum to get around their (previously low) default limits.  For those who do want to limit the memory NetLogo uses, I updated the docs to reflect how to re-add the setting.~~

After more research, J17 doesn't dynamically allocate the memory, but by default it allocates a percentage based on total system memory.  That default is 25%.  That is fine for most modern systems and for those with 4gb or more of system memory  NetLogo will have the same limit it had before (1gb) or greater.  But for older systems with less memory we don't want them hitting the smaller maximums.  As such we'll switch the default NetLogo uses to 50% of system memory.  For most users they'll never notice it.  For users on older machines they'll have a similar starting limit.  For BehaviorSpace users with memory-heavy models they'll be more likely to have things "just work" without resorting to tweaking config file settings.  So overall this should be an improvement.  

I did testing of both memory-heavy models on a system with plenty of memory available as well as running basic models on a memory-limited VM and everything worked well with the new setting.  Users running with less than 2gb of memory will have a lower maximum than before.  I ran a suite of models from the library on such a system, including running basic BehaviorSpace experiments and LevelSpace models, and everything stayed under 512mb of memory.  So for those users this limit should suffice as they are unlikely to be trying to run big models anyway.

### DIFF
--- a/autogen/docs/faq.md.mustache
+++ b/autogen/docs/faq.md.mustache
@@ -552,35 +552,34 @@ the import items on the File menu or the `import-*` primitives.
 
 ### How big can my model be? How many turtles, patches, procedures, buttons, and so on can my model contain?
 
-We have tested NetLogo with models that use hundreds of megabytes of RAM and they work
-fine. We have reports from NetLogo users that they have run models that required up to 50
-gigabytes of RAM to use millions of patches and tens of thousands of agents, and they
-worked as desired (with the proper Java settings). Theoretically the only limit is the RAM
-available on your system, but you might hit some limits that are inherent in the
-underlying Java runtime or the operating system (either designed-in limits, or bugs).
+We have tested NetLogo with models that use hundreds of megabytes of RAM and they work fine. We have reports from
+NetLogo users that they have run models that required up to 50 gigabytes of RAM to use millions of patches and tens of
+thousands of agents, and they worked as desired with the proper Java settings. Theoretically the only limit is the RAM
+available on your system, but you might hit some limits that are inherent in the underlying Java runtime or the
+operating system (either designed-in limits, or bugs).
 
-The NetLogo engine has no fixed limits on size. By default, though, NetLogo
-ships with a one-gigabyte ceiling on how much total RAM it can use. If your
-model exceeds that limit, you'll get an OutOfMemoryError dialog.
+The NetLogo engine has no fixed limits on model size.  By default NetLogo sets Java  to use up to half of your available
+system memory.  If your model exceeds that memory limit, you'll get an `OutOfMemoryError` dialog.
 
-If you are using BehaviorSpace, note that doing runs in parallel will multiply
-your RAM usage accordingly.
+If you are using BehaviorSpace, note that doing runs in parallel will multiply your RAM usage accordingly.  Similarly,
+using LevelSpace with many child models increases RAM usage greatly.
 
-<a id="where-is-cfg-file"></a> Each platform contains ".cfg" files containing JVM
-settings. There is one cfg file for each sub-application (NetLogo, NetLogo 3D, HubNet
-Client, etc.).  Windows also has a cfg file for the NetLogo_Console app; on macOS and
-Linux the cfg file for NetLogo_Console is just the NetLogo.cfg file.  Although the file
-location varies by platform, the process for changing it is the same. Locate the section
-of the file that looks like the following:
+If you want to increase the memory limit Java uses, you can do so using the NetLogo app configuration files.
+
+<a id="where-is-cfg-file"></a>Each platform contains ".cfg" files containing JVM settings. There is one cfg file for
+each sub-application (NetLogo, NetLogo 3D, HubNet Client, etc.).  Windows also has a cfg file for the NetLogo_Console
+app; on macOS and Linux the cfg file for NetLogo_Console is just the NetLogo.cfg file.  Although the file location
+varies by platform, the process for changing it is the same. Locate the section of the file that looks like the
+following:
 
     [JavaOptions]
     # there may be one or more lines, leave them unchanged
-    -Xmx1024m
+    java-options=-XX:MaxRAMPercentage=50
     # there may be one or more lines, leave them unchanged
 
-Modify the value immediately following `-Xmx`, changing it to the amount of
-space you need, save the file, and restart NetLogo. Platform specific notes
-follow:
+You can change the percentage number from 50 to whatever you want.  Save the file and restart NetLogo for the setting to take effect.  If you need to set an exact amount of memory (as opposed to a percentage), you can also comment out the `java-options=-XX:MaxRAMPercentage=50` line with a `#` at the start and add a new line with `java-options=-Xmx####m`.  Replace the `####m` with whatever amount of memory you like; using `4096m` would limit NetLogo to 4096 megabytes of memory, or 4 gigabytes.
+
+Platform specific notes follow:
 
 - **Windows:** The file will typically be in `C:\Program Files\NetLogo {{version}}\app`,
   unless you are running 32 bit NetLogo on a 64 bit Windows, then it will be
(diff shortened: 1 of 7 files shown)


SethTisue:
Wow, cool. This really works?

LaCuneta:
@SethTisue Yep!  Tested on some fairly memory-heavy models using many threads in BehaviorSpace.  We'll obviously test it out more before releasing this to the wild, as well.

LaCuneta:
@SethTisue Your question made me actually read docs (which I should've done in the first place) and I was wrong.  The change in the J17+ is to increase the max heap size memory limit based on the memory of the system.  So it isn't fully dynamic to the limit of the system, it's just much greater than the 1gb limit NetLogo was using and can be even higher than we'd think to pre-set based on system memory.  So it still makes sense to remove this hard-coded setting, but I'll clarify the language that setting a max limit may still be necessary.

autogen/docs/faq.md.mustache
```diff
@@ -552,35 +552,34 @@ the import items on the File menu or the `import-*` primitives.
...
     # there may be one or more lines, leave them unchanged
-    -Xmx1024m
+    java-options=-XX:MaxRAMPercentage=50
     # there may be one or more lines, leave them unchanged
 
-Modify the value immediately following `-Xmx`, changing it to the amount of
-space you need, save the file, and restart NetLogo. Platform specific notes
-follow:
+You can change the percentage number from 50 to whatever you want.  Save the file and restart NetLogo for the setting to take effect.  If you need to set an exact amount of memory (as opposed to a percentage), you can also comment out the `java-options=-XX:MaxRAMPercentage=50` line with a `#` at the start and add a new line with `java-options=-Xmx####m`.  Replace the `####m` with whatever amount of memory you like; using `4096m` would limit NetLogo to 4096 megabytes of memory, or 4 gigabytes.
+
```
brandesNW, line 581: Optional: 4GB can also be allocated using  
`-Xmx4g`

brandesNW (changes requested):

brandesNW:
@SethTisue - here is a simple demo that allocating maximum heap by percentage of total memory works

My Mac has 16 GB
$ `sysctl -a | grep memsize`

> hw.memsize: 17179869184

By default java allocates 25% = 4 GB for the maximum heap 
$ `java -Xlog:gc+heap=debug -version   2>&1 | grep heap`

> [0.008s][debug][gc,heap] Minimum heap 8388608  Initial heap 268435456  Maximum heap 4294967296

Setting the max ram percentage to 50 increases the maximum heap to 8 GB
$  [java -XX:MaxRAMPercentage=50 -Xlog:gc+heap=debug -version   2>&1 | grep heap](url)

> [0.025s][debug][gc,heap] Minimum heap 8388608  Initial heap 268435456  Maximum heap 8589934592

BTW you can also specify the maximum heap in GB, not just MB
$` java  -Xmx4g -Xlog:gc+heap=debug -version   2>&1 | grep heap`

> [0.006s][debug][gc,heap] Minimum heap 8388608  Initial heap 268435456  Maximum heap 2147483648

autogen/docs/faq.md.mustache
LaCuneta: I don't want to make it overly confusing, so I'd like to stick to just one unit.  I would just list the `g` option, but it looks like the JVM doesn't allow fractional arguments like `1.5g`.  So I think just listing just the megabytes option should be okay.

brandesNW:
That’s fine, it was just a thought

  *   Aaron

--


From: Jeremy B ***@***.***>
Date: Tuesday, November 29, 2022 at 5:33 PM
Cc: Aaron Andre Brandes ***@***.***>, Review requested ***@***.***>
Subject: Re: [NetLogo/NetLogo] Feature: Set default max heap size to percent instead of hard limit (PR #2068)

@LaCuneta commented on this pull request.

________________________________

In autogen/docs/faq.md.mustache<https://urldefense.com/v3/__https:/github.com/NetLogo/NetLogo/pull/2068*discussion_r1035350262__;Iw!!Dq0X2DkFhyF93HkjWTBQKhk!Ro_Y_Cv8Iuq8xRI20qLypDPPFJNkI8yKi8YlP6UkhCIlSTK06SXey4sKPtSuXOcXBtqQxoafMnUgGa9bJcRS-jm73gtCKhyYOQ$>:

>      # there may be one or more lines, leave them unchanged



-Modify the value immediately following `-Xmx`, changing it to the amount of

-space you need, save the file, and restart NetLogo. Platform specific notes

-follow:

+You can change the percentage number from 50 to whatever you want.  Save the file and restart NetLogo for the setting to take effect.  If you need to set an exact amount of memory (as opposed to a percentage), you can also comment out the `java-options=-XX:MaxRAMPercentage=50` line with a `#` at the start and add a new line with `java-options=-Xmx####m`.  Replace the `####m` with whatever amount of memory you like; using `4096m` would limit NetLogo to 4096 megabytes of memory, or 4 gigabytes.

+

I don't want to make it overly confusing, so I'd like to stick to just one unit. I would just list the g option, but it looks like the JVM doesn't allow fractional arguments like 1.5g. So I think just listing just the megabytes option should be okay.

EwoutH:
This is really nice, the default allocation was 1 GB on all devices with NetLogo 6.2 and older right? That will be a huge increase by default for all 8, 16 and 32 GB devices out there!

EwoutH:
@LaCuneta Could we do a new patch release with this PR? This could be extremely useful to have in by default.

brandesNW:
Hi Ewout,
We appreciate your continued interest in improving NetLogo.
I believe that all a user would need to do to get the benefit of the new argument is to replace the line in the config file that says
java-options=Xmx1024
With
java-options=-XX:MaxRAMPercentage=50

I would also recommend adding
java-options=-XX:+UseStringDeduplication

which will save heap memory by having the char array of duplicate strings point to the same location.
I might make sense to send a message out to NetLogo users to that effect.

My belief is that all the other changes in the PR were to “bake” the change into our packaging process.

Anyone who downloads the source and builds and packages NetLogo would create a configuration file with java-options=-XX:MaxRAMPercentage=50

If we did a release it would include all of the changes to date. A release requires a build and package process on MacOS, Windows 32 and 64 bit and Linux 32 and 64 bit plus in group testing and a beta release. I think this is too much overhead at this time.

Best,
Aaron

--


From: Ewout ter Hoeven ***@***.***>
Date: Tuesday, December 6, 2022 at 6:22 AM
Cc: Aaron Andre Brandes ***@***.***>, Review requested ***@***.***>
Subject: Re: [NetLogo/NetLogo] Feature: Set default max heap size to percent instead of hard limit (PR #2068)

@LaCuneta<https://urldefense.com/v3/__https:/github.com/LaCuneta__;!!Dq0X2DkFhyF93HkjWTBQKhk!Q6zGT3ueyl0hNE0ghU1c6pCCi9xOWfuuaY_WAnxCYyZalffDKvZC-cjnSitMIjCQyVA5A37H9-pmrhxMFSJ_7x-jwuv9EkqNbw$> Could we do a new patch release with this PR? This could be extremely useful to have in by default.

EwoutH:
Thanks for your reply. I have some thoughts on the release process, but I will save those for another thread.

> java-options=-XX:+UseStringDeduplication

Do you think it would be useful to enable this by default to compiled builds? I can imagine many agents and patches have comparable variables.

brandesNW:
Hi Ewout,
Yes, I think we will add string deduplication for the next release, but no formal decision has been made.

  *   Aaron

--


From: Ewout ter Hoeven ***@***.***>
Date: Tuesday, December 6, 2022 at 11:28 AM
Cc: Aaron Andre Brandes ***@***.***>, Review requested ***@***.***>
Subject: Re: [NetLogo/NetLogo] Feature: Set default max heap size to percent instead of hard limit (PR #2068)

Thanks for your reply. I have some thoughts on the release process, but I will save those for another thread.

java-options=-XX:+UseStringDeduplication

Do you think it would be useful to enable this by default to compiled builds? I can imagine many agents and patches have comparable variables.

PR #2076, opened 2023-01-23
Increase default memory limit percentage from 50% to 70%

EwoutH:
Increase the default memory limit that NetLogo uses by default from 50 to 70 percent. Follow up on #2068.

On most systems this gives an appreciable increase in available memory:

| GB | 50% | 70%  |
|----|-----|------|
| 4  | 2   | 2,8  |
| 8  | 4   | 5,6  |
| 16 | 8   | 11,2 |
| 32 | 16  | 22,4 |
| 64 | 32  | 44,8 |

There are few cases this will introduce problems, while it would allow more models to run by default.

### DIFF
--- a/autogen/docs/faq.md.mustache
+++ b/autogen/docs/faq.md.mustache
@@ -574,12 +574,12 @@ following:
 
     [JavaOptions]
     # there may be one or more lines, leave them unchanged
-    java-options=-XX:MaxRAMPercentage=50
+    java-options=-XX:MaxRAMPercentage=70
     # there may be one or more lines, leave them unchanged
 
-You can change the percentage number from 50 to whatever you want.  Save the file and restart NetLogo for the setting to
+You can change the percentage number from 70 to whatever you want.  Save the file and restart NetLogo for the setting to
 take effect.  If you need to set an exact amount of memory (as opposed to a percentage), you can also comment out the
-`java-options=-XX:MaxRAMPercentage=50` line with a `#` at the start and add a new line with `java-options=-Xmx####m`.
+`java-options=-XX:MaxRAMPercentage=70` line with a `#` at the start and add a new line with `java-options=-Xmx####m`.
 Replace the `####m` with whatever amount of memory you like; using `4096m` would limit NetLogo to 4096 megabytes of
 memory, or 4 gigabytes.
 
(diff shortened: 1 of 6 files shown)


brandesNW:
My initial reaction is that we should keep the 50% limit, and leave it to the users of very large models to change their configuration files. This is how we handle it now. [In the FAQ](https://ccl.northwestern.edu/netlogo/6.3.0/docs/faq.html#how-big-can-my-model-be-how-many-turtles-patches-procedures-buttons-and-so-on-can-my-model-contain) we set a default of 1 GB (`-Xmx1024m`) and indicate how to modify it. 
We want to avoid out of memory errors for users who have computers with modest amounts of RAM. The default value of  `-XX:MaxRAMPercentage` is .25, so we have increased it significantly.

brandesNW:
The two guidelines I found are for users of containers, and they hold opposing views.

[The more conservative of these says:](https://www.ccampo.me/java/docker/containers/kubernetes/2019/10/31/java-in-a-container.html)

> Don’t touch these parameters unless you want to squeeze some extra memory headroom for your Java process. In most cases, the 25% default is a safe approach to managing memory. It might not be the most memory efficient, but RAM is cheap and it’s always better to err on the side of caution, versus having your JVM process OutOfMemory-killed for some unknown reason.
> 
> If you do want to tweak these, be conservative. 50% is most likely a safe value where you can expect to avoid issues (in most cases), however this is still is largely dependent on how big (memory-wise) your container is. I don’t recommend going to 75% unless your container has at least 512 MB of RAM (preferably 1 GB), and you have a good understanding of the memory usage of the application in question.
> 
> If your container is running multiple processes in addition to Java, be extra cautious when tweaking these values. The container memory is shared across all of these processes, and understanding total container memory usage in these cases is more complicated.
> 
> Anything over 90% is probably asking for trouble.

[Another ](https://www.alibabacloud.com/help/en/sae/latest/best-practices-for-jvm-heap-size-configuration) recommends 70%

> `-XX:MaxRAMPercentage`	The option to specify the maximum percentage of container memory that can be used by the JVM. Due to system component overheads, we recommend that you set this option to a value that is less than or equal to 75.0. The recommended value is 70.0

SethTisue:
> We want to avoid out of memory errors for users who have computers with modest amounts of RAM

I find this convincing.

PR #2078, opened 2023-01-30
Fix build failure on Linux

kgmt0:
On distros that do not use `update-alternatives`, build currently fails with the following error message:
```
[error] java.io.IOException: Cannot run program "update-alternatives": error=2, No such file or directory
```

The error is caused by JavaPackager trying to use `update-alternatives` to find alternative JDKs, which would fail if it isn't available. This PR fixes this by catching the `IOException` that occurs in this situation and just use the path-specified JDK.

### DIFF
--- a/project/JavaPackager.scala
+++ b/project/JavaPackager.scala
@@ -145,7 +145,9 @@ object JavaPackager {
       try {
         Process("update-alternatives --list javac".split(" ")).!!
       } catch {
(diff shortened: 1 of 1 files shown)


LaCuneta (approved):
Thanks for this!  I generally find the use of `update-alternatives` and other ways of trying to be "smart" about locating the Java installs to be annoying, but it's one of those things that hasn't been worth the time to update.

PR #2108, opened 2023-04-21
In BehaviorSpace from the GUI if a plot fails to compile print error …

brandesNW:
…message to console and keep processing.

Fixes: 425, 1087


LaCuneta (changes requested):
The `HeadlessWorkspace.newInstance` method (and so the `HeadlessModelOpener`) is how we [advise people to run headless models using the controlling API](https://github.com/NetLogo/NetLogo/wiki/Controlling-API#example-headless).  It's not only used for BehaviorSpace.  As such I don't think we want to squelch all plot compile errors all the time there, as that would prevent embedding users from detecting plot compile errors without jumping through some extra hoops.  

The next best solution I can think of is wiring up an optional argument for ignoring plot errors, and threading that through the call stack back to BehaviorSpace.  It's a little ugly, but we already do something similar for `shouldAutoInstallLibs`.  If it got any more complex, adding a model open settings case class to pass everything together might be worth it, but for two args I don't think it is.

PR #2122, opened 2023-06-30
save the number of threads in BehaviorSpace dialog Fixes #2121

### LINKED ISSUE #2121: BehaviorSpace does not remember the number of threads from prior dialog
To recreate:

- Create a model with a BehaviorSpace experiment  or Open a model with a BehaviorSpace experiment such as Red Queen
- Open the BehaviorSpace dialog - Tools->BehaviorSpace
- Select an experiment
- Select Run. A dialog will pop up
- Change the value for "Simultaneous runs in parallel"
- Change some other dialog values such as   Disable spreadsheet output, Update view 
- Click OK
- After experiments ends (you can use abort) Select Run again
- The number of threads will again be the default, but _Disable_ and _Update_ settings will be remembered

I think a user would prefer to have the number of threads remembered.
Note: the file name is not remembered. Using the same file name could lead to accidental data overwrites.

brandesNW:


LaCuneta (approved):
:+1: Looks good to me

LaCuneta:
I'll merge once the tests pass.

Issue #2126, opened 2023-07-14
If an included file is edited and closed it is not compiled.


brandesNW:
The issue can be reproduced using the attached NetLogo model and nls file.

1) Click Test button - see a word in command center.
2) From the code tab, open set-parameters.nls.
3) Edit it - set param-a to a different value.
[saved-nls-file-not-compiled.zip](https://github.com/NetLogo/NetLogo/files/12053692/saved-nls-file-not-compiled.zip)

4) Click "close" and save set-parameters.nls.
5) Go to the Interface and click "Test" again.
     The original value is printed out.
     Corrrect behavior would be for new saved value to be printed.

PR #2127, opened 2023-07-14
Improve bookkeeping in TemporaryCodeTab

### LINKED ISSUE #2021: Save dialog appears after clicking the "close" button on a saved included file
To reproduce

1. Open a (could be new) NetLogo model
2. Go to the `Code` tab.
3. Open a tab for an included file. (Create a new one, or open an existing one.)
4. Make changes in the file buffer.
5. Save the changes (Cmd-S on MacOS, Ctr-S on other OSs)
6. Click the `Close` tab

A dialog asking you if you want to save the changes will appear.
This should not happen, because the changes are already saved.
If you open an existing included file and do not change it, no dialog will appear asking if you want to save it.

brandesNW:
If an included file with unsaved changes is closed and saved, compile it. If an included file with saved changes is closed don't ask if user wants to save it.

fixes  #2021, #2126

### DIFF
--- a/netlogo-gui/src/main/app/codetab/TemporaryCodeTab.scala
+++ b/netlogo-gui/src/main/app/codetab/TemporaryCodeTab.scala
@@ -89,7 +89,9 @@ class TemporaryCodeTab(workspace: AbstractWorkspace with ModelTracker,
     super.dirty_=(d)
     if (d) {
       saveNeeded = true
+      // This call lets the DirtyMonitor know if there is a separated code window
       tabs.setDirtyMonitorCodeWindow
+
       new WindowEvents.DirtyEvent(Some(filename.merge)).raise(this)
     }
   }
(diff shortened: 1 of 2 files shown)


netlogo-gui/src/main/app/codetab/TemporaryCodeTab.scala
```diff
@@ -101,17 +103,42 @@ class TemporaryCodeTab(workspace: AbstractWorkspace with ModelTracker,
...
+
     new WindowEvents.ExternalFileSavedEvent(filename.merge).raise(this)
   }
 
   def close() {
+    var compileNeeded = false
     ignoring(classOf[UserCancelException]) {
-      if(dirty && Dialogs.userWantsToSaveFirst(filenameForDisplay, this))
-        save(false)
+      if(saveNeeded) {
```
LaCuneta, line 117: Prefer space after keywords before open paren.

LaCuneta (changes requested):
Thanks for this; it'll be nice to get some of these included file bugs resolved.  I left a comment for a minor code style issue.

I think you've explained it to me before, but I'm unclear on the difference between `saveNeeded` and `dirty`.  Why do they need to be tracked separately?

This change also appears to include code to recompile the model when an unsaved (but compiled) change is discarded - the `compileNeeded` flag in the `close()` method.  I don't see an issue for that, but I tested the scenario anyway and it does not appear to work.  After closing the unsaved change the UI still reflects the previously compiled version.  If the intent is to fix that scenario, I think there might still be something wrong there.  If fixing that scenario isn't part of this work I think those changes should be removed from this PR.

brandesNW:
the `compileNeeded` flag has been removed

PR #2130, opened 2023-07-25
BehaviorSpace Custom Measure Conditions

ERSUCC:
### Motivation

Currently, BehaviorSpace experiments can only run metrics either every tick or at the beginning and end of each run. Allowing the user to specify conditions for when to run metrics will yield more relevant and concise data.

### GUI Changes

There is now an extra reporter field after the run metrics every step checkbox, which becomes available when the checkbox is unchecked. The user can type a single-line reporter into the field, which will be run every tick to determine whether or not to run metrics. The functionality from the previous version of BehaviorSpace is still possible, as indicated by the text below the reporter field.

### Other Changes

With the addition of the metrics reporter field, there is a new optional element in the XML that stores BehaviorSpace experiments in NetLogo files, called “runMetricsCondition”. If present, the new metrics reporter field will be populated with the contents of the element, but if it is not present, nothing will happen. This means that old BehaviorSpace experiments will still load correctly, and their functionality will be retained.

The spreadsheet exporter had to be modified to be compatible with having measurement conditions. 

When there are no inputs for the reporter, measurements are now run at the beginning and end of each run rather than just at the end.

Due to changes in the GUI and the XML, the BehaviorSpace documentation will also need to be changed.

### Test Models

[MetricsTest.nlogo.zip](https://github.com/NetLogo/NetLogo/files/12176037/MetricsTest.nlogo.zip)
[test_experiment.nlogo.zip](https://github.com/NetLogo/NetLogo/files/12176053/test_experiment.nlogo.zip)


ryanp8:
Test experiments: 
[test_experiment.nlogo.zip](https://github.com/NetLogo/NetLogo/files/12165692/test_experiment.nlogo.zip)

brandesNW (commented):
in this line and the next it should be `reporter`, not `command`

brandesNW (commented):
in this line and the next it should be `reporter`, not `command`

brandesNW (commented):
a

brandesNW (commented):
"runMetricsWithReporters" outdated?

brandesNW (commented):
a

brandesNW (commented):
"runMetricsWithReporters" outdated?

netlogo-gui/src/main/properties/ReporterLineEditor.scala
```diff
@@ -6,7 +6,7 @@ package properties
 import org.nlogo.awt.Fonts.platformMonospacedFont
 import javax.swing.{JScrollPane, ScrollPaneConstants}
 import editor.{EditorField, Colorizer}
-import java.awt.{Dimension, Font}
+import java.awt.{Dimension, Font, GridBagConstraints}
```
brandesNW, line 9: put before javax import
you could move the org.nlogo import below the editor one

netlogo-gui/src/main/properties/ReporterLineEditor.scala
```diff
@@ -18,6 +18,11 @@ abstract class ReporterLineEditor(accessor: PropertyAccessor[String],
     editor,
     ScrollPaneConstants.VERTICAL_SCROLLBAR_NEVER,
     ScrollPaneConstants.HORIZONTAL_SCROLLBAR_NEVER)
-  override def get = super.get.map(_.trim).filter(_.nonEmpty)
-  override def getMinimumSize = new Dimension(500, 500)
+  override def get = super.get.map(_.trim)
```
brandesNW, line 21: would keeping the .nonEmpty filter cause an error? If so a comment might be helpful

brandesNW (changes requested):
Just needs some minor changes.

brandesNW (commented):
Just needs some minor changes.

brandesNW (commented):
Just needs some minor changes.

models
brandesNW, line 1: These models should not be part of the PR

netlogo-gui/src/main/properties/EditPanel.scala
```diff
@@ -2,7 +2,7 @@
 
 package org.nlogo.properties
 
-import java.awt.{Component, Insets, GridBagConstraints, Dimension, GridBagLayout, BorderLayout}
+import java.awt.{Component, Insets, GridBagConstraints, Dimension, GridBagLayout, BorderLayout }
```
brandesNW, line 5: remove extra space before }

netlogo-gui/src/main/properties/EditPanel.scala
```diff
@@ -130,18 +131,23 @@ class EditPanel(val target: Editable, val compiler: CompilerServices, colorizer:
 
   def valid() = {
     def valid(editor: PropertyEditor[_]) = {
+      // Exclude runMetricsCondition from checking if the box is empty because an empty input is valid
+      if (editor.accessor.accessString != "runMetricsCondition") {
```
brandesNW, line 135: I think it would be clearer to have 

if (string) == "runMetricsCondition" {
   true
} else {
previous stuff
}

netlogo-gui/src/main/properties/EditPanel.scala
```diff
@@ -65,8 +65,9 @@ class EditPanel(val target: Editable, val compiler: CompilerServices, colorizer:
       layout.setConstraints(panel, c)
       editorPanel.add(panel)
       propertyEditors += editor
-      editor.refresh()
       editor.setEnabled(property.enabled)
+      editor.refresh()
+
```
brandesNW, line 70: why was order changed?

brandesNW:
All of my requested changes have been made - Aaron

netlogo-gui/src/test/headless/TestBehaviorSpace.scala
```diff
@@ -175,19 +175,22 @@ with OneInstancePerTest with BeforeAndAfterEach {
...
-  if(!Version.is3D)
+  if (!Version.is3D)
     test("ModelWithIncludedExperiments", SlowTest.Tag) {
-      runExperimentFromModel("models/test/lab/FireWithExperiments.nlogo", "test1", "models/test/lab/FireWithExperiments1")
-      runExperimentFromModel("models/test/lab/FireWithExperiments.nlogo", "test2", "models/test/lab/FireWithExperiments2")
+      runExperimentFromModel("test/lab/FireWithExperiments.nlogo", "test1", "test/lab/FireWithExperiments1")
+      runExperimentFromModel("test/lab/FireWithExperiments.nlogo", "test2", "test/lab/FireWithExperiments2")
+      runExperimentFromModel("test/lab/FireWithExperiments.nlogo", "test3", "test/lab/FireWithExperiments3")
+      runExperimentFromModel("test/lab/FireWithExperiments.nlogo", "test4", "test/lab/FireWithExperiments4")
+      runExperimentFromModel("test/lab/FireWithExperiments.nlogo", "test5", "test/lab/FireWithExperiments5")
```
LaCuneta, line 191: These changes should also be made to the `netlogo-headless/src/test/headless/misc/TestBehaviorSpace.scala` file as well.  TLDR on that: There was a fork of the NetLogo engine that is being slowly merged back together into `netlogo-core` but many pieces still have two copies, so they need to be kept in sync.

LaCuneta (changes requested):
Nice, thanks for your work on this!  It's great you got up to speed so quickly, and this will be a much-loved feature among our BehaviorSpace users.  I left a few comments on small items in the code.

I opened Red Queen and added a reporter for the metric recording setting that would error on tick 10 `ifelse-value ticks mod 10 = 9 [ 10 / (count patches with [pcolor = blue]) ] [ true ]`.  When run, two error windows popped up, one reporting the divide by zero and the other reporting a failure to convert the result to a boolean in `shouldTakeMeasurements()` method.  I think the core issue here is that the `shouldTakeMeasurements()` method is using the same "error checking" as `takeMeasurements()`, which isn't quite right.  The way `exitConditionTrue` does the check using the `match {}` looks better.  The problem with that one is it just re-throws any errors without adding context, which means the user won't know where the error came from.  I'd recommend using the `exitConditionTrue` technique, but wrapping the exception with one that describes where in the run the problem is occuring so the user has an easier time tracking it down.

When I use the spreadsheet output type, there is no way to see what step/tick a measurement occured at.  Previously this wasn't much of an issue as it was all or nothing, but now that information might be valuable.  Of course the table output format does include that information for users who will need it, so I don't consider this a critical item that needs to be handled to merge this PR.

I thought we'd discussed in the devel meeting making the checkbox/reporter for metric measurement collapsable, but I don't see that implemented here.  I don't consider that necessary work to merge this PR, but I wanted to check and see if it's on the list.

netlogo-gui/resources/i18n/GUI_Strings_en.properties
```diff
@@ -342,6 +342,7 @@ edit.behaviorSpace.list.variablelist = The number of variable combinations for "
 edit.behaviorSpace.list.field = A list for "{0}" does not contain any elements
 edit.behaviorSpace.repetition.totalrun = Total number of runs exceeds limit
 edit.behaviorSpace.compiler.parser = The provided variables are not well-formed
+edit.behaviorSpace.metrics.invalidN = N must be 1 or greater
```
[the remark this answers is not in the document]
brandesNW: This dialog did not have any internationalization, so it was added throughout. The RunDialog already had internationalization

netlogo-gui/resources/i18n/GUI_Strings_en.properties
```diff
@@ -587,6 +590,14 @@ tools.behaviorSpace.finalCommands.info = run at the end of each run
 tools.behaviorSpace.timeLimit = Time limit
 tools.behaviorSpace.timeLimit.info = stop after this many steps (0 = no limit)
 
+tools.behaviorSpace.runoptions.spreadsheet = Spreadsheet output
```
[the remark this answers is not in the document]
LaCuneta: Oh, I see, these are using the `Prefix` method to set things up.  That's fine then, I should've thought to check it.

netlogo-gui/resources/i18n/GUI_Strings_en.properties
```diff
@@ -573,11 +574,13 @@ the experiments'' "var" values will be:<br>\
 sequential order: 1, 1, 2, 2, 3, 3<br>\
 alternating order: 1, 2, 3, 1, 2, 3
 
-tools.behaviorSpace.metrics = Measure runs using these reporters:
+tools.behaviorSpace.metrics = Measure runs using these reporters as metrics:
 # this property has an HTML break in it.
 tools.behaviorSpace.metrics.info = one reporter per line; you may not split a reporter<br>across multiple lines
-tools.behaviorSpace.runMetricsEveryStep = Measure runs at every step
-tools.behaviorSpace.runMetricsEveryStep.info = if unchecked, runs are measured only when they are over
+tools.behaviorSpace.runMetricsEveryStep = Run metrics every step
```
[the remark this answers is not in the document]
brandesNW: not actually for this item, but about: "I thought we'd discussed in the devel meeting making the checkbox/reporter for metric measurement collapsable, but I don't see that implemented here." We did discuss that at the meeting, the key point being to conserve GUI real estate. Isaac made a change so that the reporter would be on one line, which I think does the job. At any rate I think we should wait until after other GUI changes are made for other enhancements before deciding on what is best.

netlogo-gui/resources/i18n/GUI_Strings_en.properties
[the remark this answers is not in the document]
LaCuneta: I think you might've meant to comment on the other block of properties, which now make sense as I see the `Prefix()` method usage.  But I still don't see this `invalidN` key used anywhere.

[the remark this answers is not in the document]
ERSUCC: Yes, that was from a previous iteration. I will remove it. Thanks!

netlogo-gui/src/main/properties/MetricsBooleanEditor.scala
```diff
@@ -0,0 +1,21 @@
+package org.nlogo.properties
+
+import scala.collection.mutable.ArrayBuffer
+
+class MetricsBooleanEditor(accessor: PropertyAccessor[Boolean], properties: ArrayBuffer[PropertyEditor[_]]) extends BooleanEditor(accessor)
+{
+    override def changed(): Unit =
+    {
+        properties.find(_.accessor.accessString == "runMetricsCombine") match
```
LaCuneta, line 9: Sorry, I missed this the first go-through.  This `runMetricsCombine` appears to be a leftover from prior work and isn't referenced anywhere else.  I think this code can be removed.

But looking at this class again, I think it'd be preferable that it not even know which property it's meant to represent.  Could it can take in a single `ProperyEditor[_]` at creation so it doesn't have to know about the `accessString` at all?  I don't consider that necessary to merge, but I would like it looked at at some point.

LaCuneta (approved):

PR #2132, opened 2023-07-31
BehaviorSpace Import/Export Experiments + Naming Requirements

ERSUCC:
### Motivation

Currently, NetLogo has no functionality for importing and exporting individual BehaviorSpace experiments, but there is functionality for running individual experiments headlessly. To use this feature, you have to manually extract and format the desired experiments from the text version of your saved NetLogo model file. Having the ability to import and export experiments would make this easier, and it would make it easier to transfer individual experiments between models or devices.

### GUI Changes

In the initial BehaviorSpace window, there are now import and export buttons. You can import any number of experiments at once and at any time while the window is open. The export button is grayed out until one or more experiments have been added, at which point you can select any number of experiments to export them at the same time. If only one experiment is selected, it will be output into one file, named according to the name of the experiment. If multiple experiments are selected, they will also be output into one file, named according to the name of the model.

### Other Changes

Currently, you can name all of your experiments with the exact same name. If you were to export them all using the new export functionality, it would overwrite the same file over and over and you would end up with only one exported experiment. To prevent this, I added a new requirement that all experiments within a model have unique names. To ensure that importing goes smoothly with this new requirement, I also added code to automatically rename imported experiments if there are naming conflicts with the existing experiments in that model. For example, if you have an experiment named "experiment" and you try to import an experiment also named "experiment", the imported experiment will be renamed to "experiment-1".

You can also currently name experiments with an empty string, which would also cause issues with exporting experiments, so I added a requirement that experiments have non-empty names.

The documentation was also changed to reflect the new changes in the GUI and the new requirements for naming.

### DIFF
--- a/autogen/docs/behaviorspace.md.mustache
+++ b/autogen/docs/behaviorspace.md.mustache
@@ -68,7 +68,7 @@ and run experiment setups. Experiments are listed by name and how many model
 runs the experiment will consist of.
 
 Experiment setups are considered part of a NetLogo model and are saved as part
-of the model.
+of the model, but can be saved as individual files (see "Importing and exporting").
 
 To create a new experiment setup, press the "New" button.
 
@@ -80,7 +80,7 @@ or left with their default values, depending on your needs.
 
 **Experiment name:** Experiments in the same model must have different names. If
 you open a model that contains experiments with duplicate names, the conflicting
-names will be altered.
+names will be altered to ensure that all experiment names remain unique.
 
 **Vary variables as follows:** This is where you specify which settings you want
 varied, and what values you want them to take. Variables can include sliders,
@@ -212,6 +212,23 @@ the run in some other way.
 **Time limit:** This lets you set a fixed maximum length for each run. If you
 don't want to set any maximum, but want the length of the runs to be controlled
 by the stop condition instead, enter 0.
+
+### Importing and exporting
+
+Although experiments are tied to a model and are usually saved along with a
+model, they can also be imported and exported individually. This allows you to
+easily transfer experiments between models, and also prepares experiments to be
+run headlessly.
+
+The **Import** button allows you to import experiments from an xml file. The
+selected files may contain any number of experiments, but any experiments that
+are formatted incorrectly will not be loaded. If you load an experiment that has
+the same name as an existing experiment, the name of the loaded experiment will
+be slightly altered to ensure that experiment names remain unique.
+
+The **Export** button allows you to export experiments to an xml file. Any number
+of experiments may be selected for export at once, but they will all be combined
+into a single output file.
 
 ### Special primitives for BehaviorSpace experiments
 
--- a/autogen/docs/behaviorspace.md.mustache
+++ b/autogen/docs/behaviorspace.md.mustache
@@ -63,9 +63,9 @@ item on NetLogo's Tools menu.
 
 ### Managing experiment setups
 
-The dialog that opens lets you create, edit, duplicate, delete, and run
-experiment setups. Experiments are listed by name and how by model runs the
-experiment will consist of.
+The dialog that opens lets you create, edit, duplicate, delete, import, export,
+and run experiment setups. Experiments are listed by name and how many model
+runs the experiment will consist of.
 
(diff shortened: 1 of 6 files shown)


ERSUCC:
Just found a really important bug, don't run the tests yet!

brandesNW:
The directory choosing should be improved, and the user should be able to choose the the name of the output file. When I ran the code the dialog opened in my home directory, but I couldn't create the file there.  I think the dialog should be similar to that for tables and spreadsheets where the user is given a default name, but can browse for a directory, and create a new one if they want.

brandesNW (changes requested):
Good job. Please address the above issues.

brandesNW (changes requested):
Good job. Please address the above issues.

brandesNW (approved):
no more changes requested

netlogo-gui/src/main/lab/gui/ProtocolEditable.scala
```diff
@@ -121,6 +122,14 @@ class ProtocolEditable(protocol: LabProtocol,
   }
 
   override def invalidSettings: Seq[(String,String)] = {
+    if (name.trim.isEmpty) {
+      return Seq("Variable" -> I18N.gui.get("edit.behaviorSpace.name.empty"))
+    }
+    for (exp <- experimentNames) {
```
LaCuneta, line 128: I'd probably just do `if (experimentNames.contains(name.trim)) ...` here for clarity.

netlogo-gui/src/main/lab/gui/ManagerDialog.scala
```diff
@@ -145,6 +155,70 @@ private class ManagerDialog(manager:       LabManager,
...
+
+      dialog.setDirectory(System.getProperty("user.home"))
+      dialog.setFilenameFilter(new XMLFilter)
+      dialog.setMultipleMode(true)
+      dialog.setVisible(true)
+
+      for (file <- dialog.getFiles)
+      {
+        try {
+          for (protocol <- new LabLoader(manager.workspace.compiler.utilities, true, manager.protocols.map(_.name).to[collection.mutable.Set])(scala.io.Source.fromFile(file).mkString))
```
LaCuneta, line 178: Nit: Break up line to stay under 120 columns.

netlogo-gui/src/main/lab/gui/ManagerDialog.scala
```diff
@@ -145,6 +155,70 @@ private class ManagerDialog(manager:       LabManager,
...
+          {
+            manager.addProtocol(protocol)
+          }
+        } catch {
+          case e: org.xml.sax.SAXParseException => {
+            if (!java.awt.GraphicsEnvironment.isHeadless) {
+              javax.swing.JOptionPane.showMessageDialog(manager.workspace.getFrame, s"""Invalid format in "${file.getName}".""", "Invalid", javax.swing.JOptionPane.ERROR_MESSAGE)
```
LaCuneta, line 185: Nit: Break up line to stay under 120 columns.

autogen/docs/behaviorspace.md.mustache
```diff
@@ -212,6 +213,23 @@ the run in some other way.
 don't want to set any maximum, but want the length of the runs to be controlled
 by the stop condition instead, enter 0.
 
+### Importing and exporting
+
+Although experiments are tied to a model and are usually saved along with a
+model, they can also be imported and exported individually. This allows you to
+easily transfer experiments between models, and also prepares experiments to be
+run headlessly.
```
LaCuneta, line 221: Maybe point out the experiment XML can be hand-edited in a text editor or edited in another script/language for running headlessly as another benefit here.

LaCuneta (changes requested):
Nice work!  I tested out a bunch of variations and everything seems good.  I left a few very small items in the comments.

The only thing that seemed off to me was that when exporting a single experiment the model title wasn't included.  If you already discussed and decided on that for a reason, that's fine.  To me I'd prefer having my model title prepended (as when doing the multi-export) as it helps me keep track of where the experiment came from.

LaCuneta (approved):

PR #2133, opened 2023-08-02
BehaviorSpace New Parameter Varying Syntax

ERSUCC:
### Motivation

The existing syntax available for defining how parameters should be varied in BehaviorSpace only allows the user to perform all possible combinations of the given values. Being able to provide a series of specific parameter combinations would make it easier to perform just a few combinations without having to manually adjust the variables and run the experiment for each combination.

### GUI Changes

The informational text below the parameter box now indicates the possibility of using the new syntax. There are no major changes to the GUI.

### Syntax Changes

The new syntax, called tuple syntax, allows the user to specify a series of parameter combinations in the following form:

`[ [ "variable1" value ] [ "variable2" value ] ... ]`

Each new line in the parameter box indicates a new run with a new combination of parameters. Each set of tuples must contain the same variables in the same order. The tuple syntax may not be mixed with the existing syntax within one experiment, although multiple experiments in the same model may use different syntaxes.

### Other Changes

Major changes will be made to the documentation to reflect the changes in available syntax. New tests will also be added to ensure that the new syntax has been implemented correctly.


SethTisue:
looks like a great improvement

PR #2134, opened 2023-08-07
BehaviorSpace include ticks in spreadsheet when custom metrics conditions are provided.

ryanp8:
**Motivation**
If a user uses the new custom metrics condition reporter, it could be unclear in the spreadsheet output when the metrics are reported. Including the ticks would clarify when the metrics are run.

If a custom metrics condition is not used, ticks are not included to preserve the prior spreadsheet behavior. It would also not be necessary to include the ticks because the user will already know that the metrics are run at every step or only at the end of each run.

The documentation has also been modified to reflect these changes.


LaCuneta (approved):
Accidentally left my initial review on this on #2135, but after changes this looks good to me.  Thanks again!

Issue #105, opened 2012-03-14
BehaviorSpace: initial state of foreground vs. background runs is different


SethTisue:
Each background run gets a new, freshly initialized, independent world state, including fresh RNG state.

But in foreground runs, the ones you see in the GUI, if you don't set the random seed, each run inherits the world state, including the RNG state, from the end of the previous foreground run.

Normally, users don't notice the difference because, because it's normal for an experiment's setup commands to:
- call `clear-all` so all runs get fresh world state
- (if the user wants reproducible runs) explicitly seed the RNG

This erases the difference between foreground and background runs.

In a future version of BehaviorSpace we should probably just start every run with `clear-all` and `random-seed new-seed`, something like that.

reported by [Michal Kvasnicka](http://groups.yahoo.com/group/netlogo-users/message/10608), Peter Meißner, and others

nicolaspayette:
The state of interface widgets can also be affected by this issue.

For example:

- Create a slider `x`.
- _Don't_ change its value in `setup`.
- Put `set x x + 1` in `go`.
- Create a BehaviorSpace experiment and _remove_ the range variation suggested for `x`.

If you run the experiment, the final value of `x` will be different between foreground runs and background runs.

This is unlikely to occur, as users will usually specify a value range for `x` in their experiments, but further illustrates the discrepancy between types of runs.

I can't help but ask: are foreground runs really necessary?

nicolaspayette:
Another related issue: if you have code with side effects called from a widget (e.g., a monitor, or slider bounds), background runs vs. foreground runs will have different results.

(Just to be clear: having side effects in widget code is evil and you should never do that anyway.)

SethTisue:
foreground runs aren't really necessary, and IMO aren't worth the extra complication in the code. historically they came first, and background runs were added later; I now regret keeping foreground runs around at all.

at the time, I made the decision I did because:

1) backward compatibility with the old semantics for existing models with existing experiment definitions (said compatibility achieved by disabling background runs)
2) fear that there would be a fight over removing a feature. in practice this almost always produces controversy

regardless, it would be neat if you could look in on the progress of _any_ run, not just the ones BehaviorSpace happens to arbitrarily select to be foreground, but it's probably not that high priority in the full space of BehaviorSpace would-be-neats. if it were added, the modelruns code would be a good basis for it, I think.

nicolaspayette:
> backward compatibility with the old semantics for existing models

Is that still potentially an issue?

[...]

PR #2135, opened 2023-08-07
BehaviorSpace restore widget globals after experiment is completed

ryanp8:
**Motivation**
Currently, widget values are updated to the last value that was used in an experiment. When there are multiple runs, the run that updates the final widget values can't be predicted. Allowing the experiment to modify the widgets and the unpredictable final values could be confusing for the user. The widget values should be restored to their initial state when the experiment is finished.

**Changes**
When the experiment is started, the current widget values are stored in memory. When the experiment is completed, they are reset to the initial values. Non-widget globals are not restored. 

The documentation has also been updated to explain that the user may need to rerun their setup procedure at the end of the experiment to synchronize the view with the widgets because the final view and plots may not correspond to the restored widget globals.

The following model contains an experiment to test the new behavior:
[test_widgets.nlogo.zip](https://github.com/NetLogo/NetLogo/files/12283609/test_widgets.nlogo.zip)


SethTisue:
closely related: #105

autogen/docs/behaviorspace.md.mustache
```diff
@@ -83,7 +83,11 @@ names will help you keep them straight.
 
 **Vary variables as follows:** This is where you specify which settings you want
 varied, and what values you want them to take. Variables can include sliders,
-switches, choosers, and any global variables in your model.
+switches, choosers, and any global variables in your model. You may notice that
+the view and plots do not correspond to the values in the widgets once the experiment is completed. 
+This is because the globals that are controlled by widgets are reset to their initial state at the end of the 
+experiment. To synchronize your view and plots with the widgets, run your procedure
+that initializes the view.
```
brandesNW, line 90: please remove the trailing whitespaces
Screenshot 2023-08-09 at 12 00 49 PM

brandesNW (commented):
minor change - trailing whitespace in mustache file requested

brandesNW (commented):
minor change - trailing whitespace in mustache file requested

brandesNW (commented):
minor change - trailing whitespace in mustache file requested

brandesNW:
this PR modifies  autogen/docs/behaviorspace.md.mustache which is also modified by:
BehaviorSpace Import/Export Experiments + Naming Requirements
#2132
BehaviorSpace include ticks in spreadsheet when custom metrics conditions are provided.
#2134

netlogo-core/src/main/lab/SpreadsheetExporter.scala
```diff
@@ -114,8 +123,13 @@ class SpreadsheetExporter(modelFileName: String,
     out.println()
     out.print(csv.header(if (protocol.runMetricsEveryStep || !protocol.runMetricsCondition.isEmpty) "[all run data]"
                               else "[initial & final values]"))
-    for(_ <- runs; metric <- protocol.metrics)
-      out.print(',' + csv.header(metric))
+    for(_ <- runs) {
+      if (shouldIncludeSteps)
+        out.print(',' + csv.header("tick"))
```
LaCuneta, line 128: In the table output when it lists the ticks it uses `[step]` as the header; I think we should match that here since the **Go commands** for the experiment might not advance the tick counter, or might advance it less than 1.

LaCuneta (changes requested):
Looks good, thank you for your work!  Just one minor change left in comment.

LaCuneta (approved):
Looks good to me, thank you.

PR #2137, opened 2023-08-12
BehaviorSpace Subexperiment Syntax

ERSUCC:
### Motivation

It is currently impossible to create specific parameter combinations without trying every possible combination of the given values. This new subexperiment syntax allows the user to specify subsets of the parameter combinations, without disrupting the standard syntax for varying parameters.

### Syntax Changes

The new syntax allows users to define a list of subexperiments with specific parameter combinations, as well as a list of constants that are added to each subexperiment where they are not overridden. Subexperiments must be defined after all constants, so if no subexperiments are defined, the experiment will function as it would have before the new syntax was added, treating the constants as normal parameter combinations. Subexperiments are indicated by doubly nested square brackets, with each inner set of square brackets representing the parameter combinations of one variable for that subexperiment.

### Other Changes

The documentation has been updated to reflect the changes to the available syntax. The DTD for the XML format of experiments has also been changed, but it will remain fully backwards compatible, since all added elements are optional and don't change the order of existing elements.

### Test Model

The following model contains all experiments that were added to the test suite for the new syntax.

[New Syntax Tests.nlogo.zip](https://github.com/NetLogo/NetLogo/files/12363628/New.Syntax.Tests.nlogo.zip)


brandesNW:
[bspace-colors.nlogo.zip](https://github.com/NetLogo/NetLogo/files/12337225/bspace-colors.nlogo.zip)
I think that an experiment with the new-style mixed syntax
```
["num-green" 10 25]
[["num-blue" 33 66]["num-yellow" [10 1 13]]]
```

should produce the same number of runs as one with the old-style syntax
```
["num-green" 10 25]
["num-blue" 33 66]
["num-yellow" [10 1 13]]
```

I also thought (but less strongly) that they would output the same columns. However that is not the case. You can use the 4 experiments in the attached model to see that.

brandesNW:
It seems that when you have multiple reps and you use an over-ride of a constant the second rep uses the over-ridden value incorrectly. See the 2-rep expts in the attached model.
[bspace-colors-repetitions-error.nlogo.zip](https://github.com/NetLogo/NetLogo/files/12338502/bspace-colors-repetitions-error.nlogo.zip)

brandesNW:
I think it would be better to name testSubExperimentRepetitions testSubExperimentRepetitionsNonSequential to emphasize what it does

brandesNW:
All tests except the "time order problem" passed.
Ready for review by Jeremy

LaCuneta (approved):
I tested everything out and it worked well, both positive and negative cases.  Nice work. 

I've got two notes that don't need to be addressed to merge this:

1. The extra explanatory text for the functionality on the experiment editing screen means things are even more cramped there.  Just want to note again the existing issue where lower-resolution screens cannot display the full experiment window, so we may need to rethink how those tips are displayed.
2. The design time error messages were pretty clear, but when the syntax is wrong it can be hard to track down where/why.  It'd be worth giving some though on if we can point out where issues occurred when things like the square brackets are not used correctly.  I know enhancing the design time error checks is also something that's on the list.

PR #2141, opened 2023-08-17
BehaviorSpace Non-Blocking Edit Dialog

ERSUCC:
### Motivation

The edit dialog for BehaviorSpace is modal and blocks the main NetLogo window, which is not the desired behavior.

### GUI Changes

The edit dialog for BehaviorSpace is no longer modal and blocking. There are no other changes.

### DIFF
--- a/netlogo-gui/src/main/app/interfacetab/InterfaceToolBar.scala
+++ b/netlogo-gui/src/main/app/interfacetab/InterfaceToolBar.scala
@@ -91,8 +91,7 @@ class InterfaceToolBar(wPanel: WidgetPanel,
       }
       suppress(true)
       editButton.setSelected(true)
-      // "true" argument makes a modal dialog - IOB 8/16/23
-      wPanel.editWidgetFinished(target, dialogFactory.canceled(frame, target, true))
+      wPanel.editWidgetFinished(target, dialogFactory.canceled(frame, target))
       editButton.setSelected(false)
       suppress(false)
     }
--- a/netlogo-gui/src/main/app/interfacetab/InterfaceToolBar.scala
+++ b/netlogo-gui/src/main/app/interfacetab/InterfaceToolBar.scala
@@ -91,7 +91,8 @@ class InterfaceToolBar(wPanel: WidgetPanel,
       }
       suppress(true)
       editButton.setSelected(true)
-      wPanel.editWidgetFinished(target, dialogFactory.canceled(frame, target))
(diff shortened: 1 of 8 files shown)


brandesNW:
I approve. This is a redo of the previous PR bc many files previously showed up as different from hexy.

LaCuneta (changes requested):
The code looks good and I was able to interact with the code and interface of my model while setting up the experiment.  I'm sure that'll be much appreciated.

One issue: When I open a model, edit a BS experiment, then open another model through the UI, the BS windows close which is as expected.  But when I re-open BehaviorSpace through the **Tools** menu, the experiment I was editing on the previously opened model pops back up, along with the experiment list for the current model.  Same thing when just doing a new model.

One other thing: When I click to close the BS experiment list window, my experiment editor closes as well.  I dunno if I consider this a bug, but it was surprising to me at the time, although it makes sense thinking of the window hierarchy.  The reason I bring it up is that the experiment list window is over the main window by default, meaning closing it is a natural reaction when you go back to look at it from the experiment editor window.  If there is an easy way to change this behavior, great, but I don't consider it essential.

ERSUCC:
Hi Jeremy,

In regard to your first issue, I don't believe that's something solvable. Now that the edit dialog is non-modal, I don't see a way to ensure that the user closes that window before opening a new model or closing the main BehaviorSpace window, and the dialog can't be closed from elsewhere since it is stored internally in a function in another file. If you still think this behavior is really important, I could look into making a prompt that forces the user to deal with the edit dialog before closing the model.

Same goes for your second issue, there is no way now that it's non-modal to ensure that it gets saved or closed properly by the user before BehaviorSpace is closed. It wouldn't be possible to close the main BehaviorSpace window without closing the edit dialog, but it would be possible like the previous issue to prompt the user to save before closing.

Let me know your thoughts!

LaCuneta:
For issue 1, I do consider it critical as the behavior is odd, especially so because the editor window has information from the prior model in it, which is incorrect and confusing.  I would think there should be a way to programmatically close that window when the appropriate "new model is loading" event is raised.  If not and a prompt is necessary, so be it.

brandesNW:
if the main bspace window is the parent of the expt window I think it should be possible to get it to close when the parent does

ERSUCC:
It is possible but will require some rearranging and new functionality because currently there are no references saved to the edit dialog, it is opened and closed locally within one function in another file. I found a way to fix it, working on it now.

PR #2145, opened 2023-08-25
BehaviorSpace List Expansion Output

ERSUCC:
### Motivation

Reporters that return a list currently have all elements of each list combined into one cell in both the spreadsheet and table output formats, which makes parsing more difficult. The list expansion output serves as a supplementary output, displaying expanded versions of all list reporters with a single value in each cell.

### GUI Changes

The run options dialog now has an additional file select input after the existing inputs for spreadsheet and table output, which is used to select the output file location for the list expansion output. It must be selected in tandem with one or both of the existing output formats, because the expanded lists are extracted from the existing output rather than the data being collected separately during the experiment, which was done to conserve memory.

### New Format

The list expansion format contains all expanded lists, ordered first by run, then by reporter name, then by tick. Each line contains the reporter name, the run number, the parameter values specified for that run, the tick number, and all the list elements.

### Other Changes

The documentation was changed to include details about the new output format, and a sample image was added to match the sample images used for the existing output formats in the documentation.

### Note: reasonably large changes will need to be made once the pause/resume pull request has been integrated.


autogen/docs/behaviorspace.md.mustache
```diff
@@ -296,17 +296,20 @@ the "Run" button. A dialog titled "Run options" will appear.
...
-1) the NetLogo app version used, 2) the name of the NetLogo model file used, 3) the name of the
-BehaviorSpace experiment used, 4) the date and time at the start, and 5) the dimensions of the
-world used at the start.
+The run options dialog lets you choose to create data output files in two formats, **Table
+output** and **Spreadsheet output**. If one or both of these formats is selected, you can
+also select the supplementary **Lists output**. Each file path can be entered in its
+corresponding text box, or using the **Browse...* button to select a file path using the system
+file dialog. If you don't wish to use a particular format, you can clear the file text or click
+the **Disable** button and it'll be cleared for you. Data is collected for each run or step,
+according to the setting of **Measure runs at every step** option. For **Table output** and
```
brandesNW, line 305: Adjust for new wording and reporter option for "Run metrics", which replaces **Measure runs at every step**

brandesNW (changes requested):

brandesNW:
the --lists option didn't produce any output for me. In sbt I ran

brandesNW:
Ready for Jeremy's review

LaCuneta (approved):
Okay, looks good to me, thank you for your work on this.  We'll think about any needed UI changes to clarify the output options later.

PR #2148, opened 2023-08-31
BehaviorSpace Include Steps in Spreadsheet Output

ERSUCC:
### Motivation

It is useful and easier for data analysis to include the step numbers in the spreadsheet output format. Since the custom metrics condition was added, it became unclear in some cases what the step number for each run was, if the step numbers for the resulting data were not consecutive. Including the step numbers along with the current data solves this issue.

### Format Changes

The spreadsheet output now includes an additional "step" column at the beginning of each run. In the header, it is treated like a reporter, with the final statistics values displayed like the user-specified reporters.

## Other Changes

In addition, the output formats were changed to record tick 0 if specified by the custom metrics condition. To reflect recent changes in the spreadsheet format, the "[steps]" header was changed to "[total steps]" and the "[initial & final values]" header was changed to "[final value]". The BehaviorSpace tests were also changed to match the new format.


brandesNW (changes requested):

LaCuneta (approved):
LGTM.

PR #2150, opened 2023-09-07
BehaviorSpace exporter version number

ryanp8:
**Motivation**
Because there have been many changes to the format of the spreadsheet exporter which is used to generate lists, stats, and potentially plots in the future, it will be useful to keep track of the version of the spreadsheet. This allows older versions that cannot generate the new results through the post processors to be easily identified. While version numbers for lists, stats, and table are not necessary, they are added for consistency and future-proofing.

**Changes**
When a csv file is generated by a table, spreadsheet, lists, or stats exporter, the first row now has an additional column that has the type of exporter followed by `version 2.0`. For example, a table would say `Table version 2.0`.

Changes to existing tests are not needed because the tests omit the file header.

Table example: [Red Queen Beta2-2-table.csv.zip](https://github.com/NetLogo/NetLogo/files/12553601/Red.Queen.Beta2-2-table.csv.zip)


brandesNW:
There are tests that fail because of the header change and need to be updated

For example:

> [info] - successful run *** FAILED ***
> [info]   Expected "...sults (NetLogo 6.3)"[]
> [info]   "models/test/lab/Fi...", but got "...sults (NetLogo 6.3)"[,"Table version 2.0"]
> [info]   "models/test/lab/Fi..." (TestMain.scala:98)
> [info]   org.scalatest.exceptions.TestFailedException:

you can locate more by searching the results pages for "failed"

https://github.com/NetLogo/NetLogo/actions/runs/6115052606/job/16598123177?pr=2150
https://github.com/NetLogo/NetLogo/actions/runs/6115052606/job/16598123489?pr=2150

brandesNW:
The image files also need to be updated in the documentation.

brandesNW (approved):
Thanks for this update.
This depends on PR 2146 - BehaviorSpace Statistics Exporter

brandesNW:
All output files now have a second column in the first row "Table Version 2.0" etc.
This is in the doc, and all have NetLogo 4.0 and a recent date.

netlogo-core/src/main/api/LabExporterType.scala
```diff
@@ -0,0 +1,10 @@
+// (C) Uri Wilensky. https://github.com/NetLogo/NetLogo
+
+package org.nlogo.api
+
+object LabExporterType {
+ val TABLE = "Table";
+ val SPREADSHEET = "Spreadsheet";
+ val STATS = "Stats";
+ val LISTS = "Lists";
+}
```
LaCuneta, line 10: Nit: Prefer blank lines at end of files.

netlogo-core/src/main/api/LabExporterVersion.scala
```diff
@@ -0,0 +1,7 @@
+// (C) Uri Wilensky. https://github.com/NetLogo/NetLogo
+
+package org.nlogo.api
+
+object LabExporterVersion {
+  val version = "2.0"
+}
```
LaCuneta, line 7: Nit: Prefer blank lines at end of files.

LaCuneta (changes requested):
Thanks for this!  Just a couple very small items in the comments.

PR #2151, opened 2023-09-08
BehaviorSpace Experiment Naming Bug Fix

ERSUCC:
This pull request fixes two bugs with experiment renaming:

1. Experiments are not renamed when a model is opened even if the model contains experiments with duplicate names.
2. When a model is open and a model with the same experiment names is opened in the same session, the experiments are renamed in the newly opened model.


netlogo-core/src/main/fileformat/LabLoader.scala
```diff
@@ -17,10 +17,10 @@ object LabLoader {
 
 import LabLoader._
 
-class LabLoader(literalParser: LiteralParser, editNames: Boolean = false, existingNames: Set[String] = Set[String]()) {
+class LabLoader(literalParser: LiteralParser) {
   if (literalParser == null)
     throw new Exception("Invalid lab loader!")
-  def apply(xml: String): Seq[LabProtocol] = {
+  def apply(xml: String, editNames: Boolean, existingNames: Set[String]): Seq[LabProtocol] = {
```
brandesNW, line 23: It looks like the arguments in `class LabLoader `are missing in this version

netlogo-core/src/main/fileformat/NLogoLabFormat.scala
```diff
@@ -16,7 +16,7 @@ trait LabFormat[A <: ModelFormat[Array[String], A]]
 
   def componentName = "org.nlogo.modelsection.behaviorspace"
 
-  val loader = new LabLoader(literalParser, editNames)
+  val loader = new LabLoader(literalParser)
```
brandesNW, line 19: Similar to the previous comment - is `editNames` being made available in a different way?

brandesNW (changes requested):
thanks for this fix, please clarify about the use of `editNames` ...

netlogo-core/src/main/fileformat/LabLoader.scala
brandesNW: Isaac clarified

> i see the confusion, basically i initially did it as a constructor parameter, but that didn’t work because each labloader object is constructed once and then reused multiple times. since each time it’s used the “existingNames” parameter might be different, i moved the parameters to the apply method instead. i kept the “editNames” parameter for posterity to make sure it worked as it did before the move, but I realize i might be able to remove that parameter so i will check on that now.
> 
> actually the editNames parameter is necessary, since sometimes the existingNames parameter will start empty and fill up as experiments are loaded, so there is no other way to indicate whether names should be changed or not.

PR #2152, opened 2023-09-08
BehaviorSpace Remove Extra Space in Lists Output

ERSUCC:


brandesNW:
Thanks for the bug fix.
No problems.
Tested using model testStats.nlogo (attached)
Compare results of experiment list-test using table output (bug) and spreadsheet only output (no bug)

Issue #1877, opened 2020-12-02
BehaviorSpace experiment window can get cut off


LaCuneta:
From a user report:  The BehaviorSpace experiment window is not scrollable, so if you're using a smaller resolution screen or using a larger resolution screen with a high scaling factor, you can wind up unable to use the controls at the bottom of the dialog like **Setup commands**, **Go commands**, **Time limit**, and the **OK** and **Cancel** buttons.

A few potential solutions are: 1) make the window scrollable, 2) make some of the upper sections collapsible to be able to see the lower elements, or 3) split things out into tabs.

ERSUCC:
The BehaviorSpace experiment edit dialog is now resizable as of commit [3e724b5](https://github.com/NetLogo/NetLogo/commit/3e724b5870203df538d2d44182185aee6539c036), so this issue no longer applies.

PR #2155, opened 2023-09-11
BehaviorSpace Post Experiment Commands

ERSUCC:
### Motivation

There is currently a reporter box titled "Final commands", which allows the user to specify commands that they would like to run at the end of each run, but there is no way to run commands only after the completion of the entire experiment.

### GUI Changes

There is now a reporter box for "Post experiment commands" at the end of the edit dialog, and the existing "Final commands" box has been renamed to "Post run commands".

### Other Changes

The XML format was changed to reflect the renaming of "Final commands" to "Post run commands" and to add the new "Post experiment commands" element. This was also updated in the existing BehaviorSpace tests. Code was also added to convert existing models upon loading if they contain the "final" tag as opposed to the new "postRun". The documentation was also changed to reflect the above changes.


brandesNW (changes requested):
Thanks for this valuable PR.
The code looks good. 
There are a few bugs.
For the post-run commands a runtime error results if you use something like `let foo 1 / (behaviorspace-run-number - 3)`. It doesn't affect the results, or even keep a command that follows it from excuting, which is what happens in 6.3 with Final Commands.
However no runtime error results using the same code with pre or post experiment commands. Can you fix this?
Note: syntax errors are handled properly in all cases.

Second issue - if I paste some text into the post run box it sometimes adjusts so the gray scroll bar hides most of the text.
Also the pre and post areas sometimes expand to the size of the setup and go commands, but usually are one line (like the post run commands usually are) Probably something with the GrabBag layout.

brandesNW:
Works correctly if one of the three Command options (Pre expt, Post run, Post expt) has a runtime error, but if
 two of them, e.g. Pre expt and Post expt) have  runtime errors the result is an get error about bad syntax, which is incorrect.

brandesNW:
The logic executes correctly.
The other problem I observed pre-dates this PR. If the text in the post run box is too long the size adjusts so that the gray scroll bar hides most of the text.

brandesNW (approved):

brandesNW (approved):

autogen/docs/behaviorspace.md.mustache
```diff
@@ -337,6 +337,9 @@ will be always be taken at the end of each model run, even if this text box is
   would record every tenth tick, as well as the last tick. Multiple reporters
    can be combined using *and* and *or*.
 
+**Pre experiment commands:** These commands will be run once, before the
+experiment begins.
```
LaCuneta, line 341: I'm not 100% sure how to phrase it, but I think we need to point out that these pre/post experiment commands cannot affect the state of any of the runs (well, they can, but only the "primary" one).  To try a bit:

> These commands execute outside of the actual runs, so they cannot affect the global variables or agents in the model (use the **Setup commands** and **Post run commands** for that).  If you use primitives like `file-open` or the CSV extension in the pre-experiment commands, their internal state will not carry over into the runs and errors will likely occur if used there.

LaCuneta (changes requested):
Thanks for this!  Code looks good, just one comment about docs I think we should address.

It doesn't need addressing in this PR, but just to point it out, this will make the issue with the experiment editing screen being almost unusable on lower-resolution monitors worse:  #1877

LaCuneta (approved):

PR #2157, opened 2023-09-11
Don't save or run when "x" button is clicked

ryanp8:
**Motivation**
In edit dialogs, the input data is currently saved when the user presses the "x" button, but the data should be reverted instead of saved. This behavior also makes BehaviorSpace experiments run when the user presses the "x" in the run options dialog.

**Changes**
This change will affect dialogs that inherit from `netlogo-gui/src/main/properties/EditDialog.scala`. Note that this will affect widget and plot edit dialogs in addition to run options dialog explained above.
With the changes, if the user clicks "x" without first hitting "apply", changes will not be saved. If "apply" was clicked, then the applied changes will be saved.

### DIFF
--- a/netlogo-gui/src/main/properties/EditDialog.scala
+++ b/netlogo-gui/src/main/properties/EditDialog.scala
@@ -95,10 +95,15 @@ trait EditDialog extends javax.swing.JDialog {
   addWindowListener(
     new java.awt.event.WindowAdapter() {
       override def windowClosing(e: java.awt.event.WindowEvent) {
-        if(editPanel.valid) {
-          editPanel.apply()
-          if(target.editFinished) bye()
-        }}})
+        if (sendEditFinishedOnCancel) {
+          if(editPanel.valid) {
+            editPanel.apply()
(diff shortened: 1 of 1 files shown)


brandesNW:
Thanks for this fix. Looks good to me.
I tested and made the following observations

- Run Options Dialog
     Experiment does not run with Cancel or x'ing
- Experiments Dialog
   Experiment does not change, duplicate experiment not created with Cancel or x'ing
- Slider Widget
    Edit and change value or name of global
    -  If don't click Apply, Cancel and x'ing both go to original value
    -  If click Apply
        Cancel goes to original value
       x'ing both results in original value
- Chooser Widget
    Same behavior as Slider if Edit and change value or name of global
- Switch Widget
   If change global name, Cancel and x'ing both go to original value
- Input Widget
    If change global name, Type or Multi-line checkbox Cancel and x'ing both go to original value

brandesNW (approved):
Looks good

LaCuneta (approved):

PR #2159, opened 2023-09-12
BehaviorSpace change default threads

ryanp8:
**Motivation**
The documentation currently recommends users to "start with a --threads value at half the number of logical cores your system has, and bump up or down from there to see where your “sweet spot” is for least time to complete all runs." Setting the default thread count to half the cores is more consistent with the documentation than the current behavior where it is set to the number of cores.

**Changes**
The default number of threads in gui and headless runs is now set to ceil(cores / 2)

### DIFF
--- a/autogen/docs/behaviorspace.md.mustache
+++ b/autogen/docs/behaviorspace.md.mustache
@@ -15,7 +15,7 @@ A number of new features were introduce in NetLogo 6.4:
  combinations to be run separately, rather than being expanded combinatorically.
 - [**Run metrics when**](#run-metrics-when): Measurements can now be conditionally recorded,
  controlled by a reporter.
-- [**Importing and exporting**](#importing-and-exporting): Experiments can now be exported to 
+- [**Importing and exporting**](#importing-and-exporting): Experiments can now be exported to
 an XML file that can be used when running headlessly. Experiments can also be imported into a
  model.
 - [**Statistics output**](#statistics-output): The mean and standard deviation
@@ -24,7 +24,7 @@ of data from repetitions can be saved in an output file.
  element per cell.
 - [**Paused experiments**](#paused-experiments): Experiments can now be paused and resumed.
 
-Additional minor changes can be found by searching this page for *(Since 6.4)* 
+Additional minor changes can be found by searching this page for *(Since 6.4)*
 
 ## What is BehaviorSpace?
 
@@ -38,7 +38,7 @@ behaviors and determine which combinations of settings cause the behaviors of
 interest.
 
 If your computer has multiple processor cores, then by default, model runs will
-happen in parallel, one per core.
+happen in parallel, one per every two cores.
 
 ### Why BehaviorSpace?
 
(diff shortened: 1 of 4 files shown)


autogen/docs/behaviorspace.md.mustache
```diff
@@ -38,7 +38,7 @@ behaviors and determine which combinations of settings cause the behaviors of
 interest.
 
 If your computer has multiple processor cores, then by default, model runs will
-happen in parallel, one per core.
+happen in parallel, one per every two cores.
```
ERSUCC, line 41: Is it worth clarifying the rounding for odd numbers of cores?

SethTisue:
> start with a --threads value at half the number of logical cores your system has

What is the origin of this advice? Why would I want to stop at half? I don't understand the reasoning.

brandesNW:
> > start with a --threads value at half the number of logical cores your system has
> 
> What is the origin of this advice? Why would I want to stop at half? I don't understand the reasoning.

I'm not sure of the exact origin, but when I was investigating a question as to whether NetLogo 6.3 used more memory than 6.2.2 (with a user model that read in lots of data from files) I found empirically that total execution time was greater using all the cores versus fewer. 

Do you have a strong objection to the change? Other recommendations?

brandesNW (changes requested):
Just a minor doc change, suggested by Isaac.

brandesNW:
R.e. Seth's comment - the number of cores used is now saved as a property, so the user will get the default value only once. If they set it to the max number of cores it will remain that way until they change it.

SethTisue:
> Do you have a strong objection to the change? Other recommendations?

Uh, there is a lot of nuance and possible uncertainty here, so here comes a long answer.

I am not truly expert on this. But I would advise being cautious about over-generalizing from one test or a handful of tests that you've done, especially if the model you tested was doing a lot of I/O; I think NetLogo models are more typically CPU bound.

My impression was that it's quite typical, standard advice on the JVM to have one thread per core as a usual rule of thumb. That's why I made it the default back in, idk, 2010 or whenever it was. And some casual Googling just now seems to confirm that it's still standard advice? I only spent a few minutes on it. But see e.g. https://engineering.zalando.com/posts/2019/04/how-to-set-an-ideal-thread-pool-size.html , which has a formula from no less an authority than Brian Goetz, and the formula never gives an answer less than 1.0 (thread/core), and sometimes give a larger answer (whereas you're suggesting to go down to 0.5).

If NetLogo is going to have a different default and different advice to users, I think you probably ought to be pretty sure that you're right.

If you tell users to use half as many threads, what is the worst possible outcome? The worst possible outcome is that they end up leaving half of their available compute power idle and their experiment will take twice as long for no good reason.

Whereas if you tell users to use one thread per core, what's the potential downside? I'm rather skeptical that you'd be risking anything like a 2x slowdown. Operating systems and the JVM are supposed to function well even when a bunch of threads are contending for the CPU. So although one thread per core might not be _optimal_, I would be surprised if it degraded performance more than a small amount.

I should add that once nuance I'm unclear on is whether physical cores vs. virtual cores matter. In ~2010, the last time I researched this, virtual cores weren't a thing yet as far as I can recall, or not in consumer-level hardware anyway.

One further issue I'd like to raise is the question of memory usage. Every additional model run that's happening in parallel will use additional heap. So if a test you did _did_ show a bad slowdown at one thread per core, my _first_ hypothesis wouldn't be that one thread per core is too many. My _first_ hypothesis would be that I didn't give the JVM enough heap.

brandesNW:
@SethTisue - we will consider your reservations seriously

brandesNW:
This PR was the basis for *New method for default number of parallel runs in BehaviorSpace #2160*  which replaced it.

PR #2163, opened 2023-09-14
Basic multi-variable let and set

LaCuneta:
Example lets:

```netlogo
let [x y z] [10 15 20]
show x ; prints 10
show y ; prints 15
show z ; prints 20
```

Example sets:

```netlogo
ask turtles [
  set [color size shape] [red 2 "arrow"]
  show color ; prints 15
  show size ; prints 2
  show shape ; prints "arrow"
]
```

That only uses agent variables, but you can freely mix agent variables, local variables, code globals, and interface globals.  

All compile time and runtime checks are the same way as with the existing non-multi-variable usage.

### TBD

This does not include:

- Argument destructuring for code procedures or anonymous procedures: `to report [[a b] c] report a + b + c end`.
- "Recursive" destructuring, things like `let [[a b] [x y]] [[10 20] [30 40]]`.

### DIFF
--- a/autogen/docs/dictionary.html.mustache
+++ b/autogen/docs/dictionary.html.mustache
@@ -4910,10 +4910,21 @@ end
       <p>
         Example:
       </p>
-      <pre>
+<pre>
 let prey one-of sheep-here
 if prey != nobody
   [ ask prey [ die ] ]
+</pre>
+      <p>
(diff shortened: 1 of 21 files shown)


brandesNW:
This is a great feature.
When there more elements in the list than needed they are silently ignored.
Is this the desired behavior?

PR #2168, opened 2023-09-15
BehaviorSpace Run Dialog Bug Fix

ERSUCC:
Fixed an issue with the run dialog where the time would keep counting longer than it should with no indication of what was going on behind the scenes. It now displays more informative progress notifications to the user for each step of the process. Also fixed bad code that was slowing down pause/resume functionality for experiments with more data.


brandesNW:
Hi - this looks good to me.

brandesNW (approved):

LaCuneta (approved):
LGTM :+1:

PR #2169, opened 2023-09-15
BehaviorSpace Edit Dialog Bug Fix

ERSUCC:
Disabled main window buttons when edit dialog is open to preserve non-modality of edit dialog while disallowing multiple edit dialogs and running experiments with the edit dialog open.

### DIFF
--- a/netlogo-gui/src/main/lab/gui/ManagerDialog.scala
+++ b/netlogo-gui/src/main/lab/gui/ManagerDialog.scala
@@ -38,6 +38,7 @@ private class ManagerDialog(manager:       LabManager,
   private val closeAction = action(I18N.gui("close"), { () => manager.close() })
   private val abortAction = action(I18N.gui("abort"), { () => abort() })
   private val runAction = action(I18N.gui("run"), { () => run() })
+  private var blockActions = false
   /// initialization
   init()
   private def init() {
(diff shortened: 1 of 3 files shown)


brandesNW:
Thanks for the PR
Unfortunately, as written it has a major flaw. Although the buttons (other than import and export) are disabled when the Experiment window is opened (via *New* or *Edit*) if the user clicks on another experiment the buttons become enabled and another experiment can be opened. Clicking the *Run* button can then freeze NetLogo.
I did some reading on the use of the Consumer and the Runnable interfaces and am not sure whether the change is appropriate. Clearly with the current approach the ability to select experiments must also be disabled. I also think the *Import* button should be disabled since we want to allow operations involving only one experiment at a time. *Export* might make sense, although there might be a confusion about what happens when the user has changed the contents of the experiment. Would they expect their changes to be exported? Logically only the previously saved version should be exported. That could be useful if you are making a change and want to back up your previous version. If we allow exporting we need to document this.
What is the effect on other edit dialogs of making finish in EditDialogFactory a  Consumer instead of a Runnable? Are any of them actually run by the *Thread* class or *ExecutorService* or was Runnable just a convenient class to use back in the day?

LaCuneta:
One other scenario that re-enables the buttons is going to **Tools > BehaviorSpace**.  

Here is one idea that might resolve the issues:  

1. Set a flag on the experiments dialog (manager or whatever it's called) when the experiment edit is open. 
2. Centralize the button enable/disable code to a single method, `updateActionAvailability()` that takes into account the flag and the rest of the existing logic when the experiment editor isn't open.
3. Call that method on dialog open, before the experiment editor is shown after the flag is set, and then again after the experiment edit is closed after the flag is unset.  Also call the same method in the other locations the button status is changed (`valueChanged()`).

Does that sound workable?

ERSUCC:
Aaron, the methods with the "finish" parameter were not there originally, I added them for use only with the edit dialog in BehaviorSpace when we switched the dialog to non-modal. It does not affect any prior uses of EditDialogFactory.

Jeremy, I will try that instead. Hopefully that will fix Aaron's other issue as well.

LaCuneta (approved):
LGTM :+1:

PR #2170, opened 2023-09-18
BehaviorSpace Tooltips

ERSUCC:
### Motivation

With the introduction of a variety of new features in 6.4, the edit dialog became too long for some screens, and even on larger screens, it appeared quite busy and hard to read and process. Converting the subtext hints to tooltips that display when the property name is hovered over will decrease the height and increase the readability of the dialog.

### GUI Changes

The subtext hints below certain properties in the edit dialog have been removed, and those properties' names are now underlined to indicate the presence of a tooltip that appears when you hover over it.

### Other Changes

The documentation was changed slightly to indicate the use of underlines to indicate a tooltip as opposed to the original subtext.


brandesNW (changes requested):
Great - all the tooltips are now working.
One bug - after editing when you hit OK you get an error pop-up
*Invalid value for Note: hover over underlined text for additional info about a field*
presumably has to do with the implementation of the top text.

netlogo-gui/src/main/properties/Label.scala
```diff
@@ -0,0 +1,15 @@
...
+
+abstract class Label(accessor: PropertyAccessor[String], useTooltip: Boolean)
+  extends PropertyEditor(accessor, useTooltip)
+{
+  setLayout(new java.awt.BorderLayout(BORDER_PADDING, 0))
+  val label = new javax.swing.JLabel(accessor.displayName)
+  tooltipFont(label)
+  label.setFont(label.getFont.deriveFont(java.awt.Font.BOLD))
+  add(label, java.awt.BorderLayout.CENTER)
+  override def get: Option[String] = None
```
brandesNW, line 13: the bug *Invalid value for Note: hover over underlined text for additional info about a field* can be fixed by changing this to
`
  override def get: Option[String] = { Some(accessor.displayName)}`

brandesNW (commented):
just a small change needed in Label.scala

netlogo-gui/resources/i18n/GUI_Strings_en.properties
brandesNW, line 1: As per the earlier slack discussion I think it would be better not to use the word *run* with two different meanings.
We can use *executed* where appropriate

**Pre experiment commands:**
executed before the experiment begins

**Post run commands:**
executed at the end of each run

**Post experiment commands:**
executed after the experiment is completed

We can add the following tooltips:

**Experiment name**
Experiment names in a model must be unique.

**Setup commands**
executed when the run begins

**Go commands**
once setup commands complete, go commands are executed repeatedly during a run until a stop condition or time limit is reached.

For consistency, 
**Run combinations in sequential order**
could become
**Execute combinations in sequential order**
or 
**Number run combinations in sequential order**
The corresponding line in the doc would need to be changed.

Is this change too pendantic?

autogen/docs/behaviorspace.md.mustache
brandesNW, line 1: I would add the tooltips to the second summary of changes as follows (some periods added)

More about changes to BehaviorSpace in NetLogo 6.4:

- Experiments created in versions prior to 6.4 can still be opened, but experiments
created using new features are not backwards compatible.
- For information on format changes to output files see [**Output File Changes**](#output-file-changes).
- The Experiment dialog now uses tooltips rather than text below the input elements see
[**Creating an experiment setup**](#creating-an-experiment-setup).
- Additional minor changes can be found by searching this page for *(Since 6.4)*.

autogen/docs/behaviorspace.md.mustache
```diff
@@ -109,7 +109,9 @@ or view sliders in the Interface Tab (*Since 6.4*).
 
 The information that can be included in the Experiment dialog is detailed below.
 Note that it is not necessary to specify everything; some parts can be left blank
-or with their default values.
+or with their default values. If a property's name is underlined, you can hover over
+it to read more info about the property and how it is used, so you don't have to
+revisit this page as often.
```
brandesNW, line 114: Please add *(Since 6.4)* and correct the misplaced period below.

revisit this page as often *(Since 6.4)*.

**Experiment name:** Experiments in the same model must have different names. If
you open a model that contains experiments with duplicate names, the conflicting
names will be altered to ensure that all experiment names remain unique *(Since 6.4)*.

LaCuneta (approved):
Nice, thanks for the work on this!  Except for the one code item Aaron noted, this seems fine to me.

I do think we should open an issue to revisit this issue after the release, both getting a more robust way to display the help text (maybe copy/pastable?) as well as a more long-term solution for organizing the experiment edit screen and other complex screens in NetLogo.

PR #2173, opened 2023-09-20
don't compile unnecessarily when there is no separate code tab.  Fixes  #2172

### LINKED ISSUE #2172: Leaving and returning to a "dirty" Code Tab causes unnecessary compilation
when there is no separate code tab if you are editing the Code Tab (so it is "dirty") and click outside the NetLogo application, and then click the NetLogo application compilation will happen. This frequently leads to a compilation error, since the code was in process.

To reproduce:

- Open a model
- Edit some code - stop in a state where it won't compile correctly
- Click outside NetLogo
- Click the Code tab again
- The code will compile and the error will be highlighted, interrupting the flow of coding

If you are editing a separate code tab, click away and then click the main application window compilation will occur.

brandesNW:
This fixes the following problem: when there is no separate code tab if you are editing the Code Tab (so it is "dirty") and click outside the NetLogo application, and then click the NetLogo application compilation will happen. This frequently leads to a compilation error, since the code was in process.

Someone pointed this out. It might have just been in an email. I looked for a corresponding issue, but didn

### DIFF
--- a/netlogo-gui/src/main/app/Tabs.scala
+++ b/netlogo-gui/src/main/app/Tabs.scala
@@ -98,11 +98,19 @@ class Tabs(workspace:           GUIWorkspace,
     tabManager.setAppCodeTabBindings
   }
 
+    // When there is a separate code tab window, there will be a selected tab in
+    // both windows. This means that the selected tab in the non-active window will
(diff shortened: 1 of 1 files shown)


LaCuneta (approved):
LGTM.  Thanks.

PR #2175, opened 2023-09-20
Edit Dialog Restructuring

ERSUCC:
### Motivation

The current format for edit dialogs is somewhat restrictive and could use a bit of generalization and flexibility. Refactoring the base class for edit dialogs will allow for better dialogs in the future.

### Changes

The EditPanel was generalized to allow static, non-editable properties such as text or graphics mixed in with the existing editable properties.


brandesNW:
Thanks, this looks like it will be very helpful in the future.
Currently holding until after the 6.4.0 release.

PR #2176, opened 2023-09-28
BehaviorSpace Plots Bug

ERSUCC:
Fixed a bug where if an experiment is run with "update plots and monitors" unchecked, it disables the plots in the main NetLogo workspace where experiments are run manually.


brandesNW (commented):
code change is good, just need to remove the models

brandesNW (approved):
good work

PR #2177, opened 2023-09-28
BehaviorSpace Edit Bug Fixes

ERSUCC:
Fixed the following two bugs:

- When an experiment is duplicated, the new experiment has its name by default set to the same as the old experiment, so you have to manually change the name in order for the names to not conflict
- Editing an experiment always overwrites the first one in the list, regardless of which experiment was chosen to edit


brandesNW:
There is still a bug.

- Duplicate an experiment
- Make a change
- Hit OK
- Run the experiment
- Hit Abort

The duplicate will overwrite one of the models, and there will be 2 of the duplicates

brandesNW (approved):
thanks, looks good

PR #2182, opened 2023-10-16
Add gogo6 extension

kgmt0:
This is a fork of the `gogo` extension that I modified to support GGB6. There are enough differences between GGB6 and earlier versions that I think warrants creating a separate extension for it.

This commits just adds a submodule pointing to my repo [here](https://github.com/kgmt0/GoGo6-HID-Extension), but I think it would make sense to move it under the NetLogo organization before merging.


LaCuneta:
We decided to go with an updated GoGo extension instead of a separate one for V6 boards.

PR #2186, opened 2023-10-25
BehaviorSpace Version Speed Comparison Tool

ERSUCC:
### Motivation

We received notice that our changes to BehaviorSpace may have had a negative impact on the runtime of larger experiments. Comparing the runtimes of specific experiments between multiple versions of NetLogo can be difficult, so this tool will help smooth out that process.

### Functionality

The tool allows you to specify two versions of NetLogo, along with an experiment and model to run and a number of identical trials to run for each version. It will then run NetLogo_Console with the specified parameters and time each run, reporting an average experiment execution time for each version at the end.

### DIFF
--- a/bin/bs-bench.scala
+++ b/bin/bs-bench.scala
@@ -2,14 +2,16 @@
 exec scala -deprecation -classpath bin -Dfile.encoding=UTF-8 "$0" "$@"
 !#
 
-import sys.process.Process
-
 object Main
 {
-    var path1 = ""
-    var path2 = ""
+    var oldPath = ""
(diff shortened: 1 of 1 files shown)


brandesNW:
great initial version. will need additional capability such as specifying

- --setup-file <path>: read experiment setups from this file instead of the model file
- --threads <number>: use this many threads to do model runs in parallel, or 1 to disable parallel runs
- --update-plots: enable plot updates

will also want to create a table output comparing results

brandesNW (approved):

PR #2187, opened 2023-10-27
BehaviorSpace Benchmark Tool Improvements

ERSUCC:
Added more argument options and an option for output to a csv file. Also added the option to override repetitions for the experiments.


brandesNW (approved):
thanks for the updates.
I am approving this so I can make some updates.
One thing that will need changing is that the model is always required, even if there is a setup file.
We will need to either add an argument or a try catch to deal with the case where the NetLogo with the old path does not take the updatePlots argument.

PR #2191, opened 2023-11-13
For BehaviorSpace Spanish translation

brandesNW:
Add BehaviorSpace Spanish translation. There is a link to it in the NetLogo User Manual.


LaCuneta (changes requested):
The image links appear broken, I think they need to include the `behaviorspace` subfolder, since the default is to assume the image subfolder name matches the page name.  Checking [the generator code](https://github.com/NetLogo/NetLogo/blob/hexy/project/Markdown.scala#L100) I think this could be done by adding `/behaviorspace/` before each image name.

LaCuneta (approved):

PR #2192, opened 2023-11-17
Fix invalid deprecated annotation used

EwoutH:
Languages like Java and Scala allow you to annotate a method as `deprecated` to signify that the method is no longer supported and will be dropped in the near future and that the users must switch to a suitable alternative. However, marking a Scala method with Java's annotation, i.e. `@Deprecated` is not ideal. Rather, it is suggested that you use the Scala's annotation, i.e. `@deprecated` as the former may or may not trigger the deprecated message.


EwoutH:
> Specify both message and version: @deprecated("message", since = "1.0")

What message and which version should we specify?

brandesNW:
This file was last changed 8 years ago and unfortunately no comment was given as to why it was deprecated.
I would go with @deprecated("Do not create additonal uses of this method.", since = "1.0")

EwoutH:
Thanks for the reminder!

Feel free to squash the commits when merging.

PR #2194, opened 2023-11-27
Implement auto-reload functionality

kgmt0:
Add a check box to MainCodeTab for auto reload. When checked, NetLogo will monitor the currently opened nlogo file for modifications, including modifications from outside NetLogo. When a modification is detected, the model is reloaded.

There are still some issues with this implementation:
* If you're on a tab other than `Interface`, NetLogo will sometimes switch back to it on reload.
* Reload is triggered twice if you save a file from NetLogo itself.
* Included nls files are not being monitored for auto-reload yet. However, nlogo files that include nls files seem to work properly based on my brief testing.


LaCuneta (commented):
This is a great first step, and I can't wait to use it!  Thank you!  I didn't leave any code comments given the work left, and I could've tested more, but I thought I'd gathered enough comments for your consideration.

My inclination is that we should just always watch the files and reload when modified.  If we do want to keep it as an option, I'd suggest moving it to the preferences dialog, so it's active for all models when opened.

The [`__includes` files](https://ccl.northwestern.edu/netlogo/docs/programming.html#multiple-source-files) code tabs will definitely need some extra work to be supported, since as written it looks like we're using only the model path for the watcher.  And the UI checkbox is only on the main code tab.

It would be awesome if the watcher code could live outside of the code tab or other UI elements (mostly).

I did a `netlogo/run` in sbt and loaded **Wolf Sheep Predation** from the models library and checked the **Auto reload** checkbox.  I made edits in VSCode, but they were not reflected in the model.  Might be some kind of models library pathing issue, but when I reloaded the model using the **File** menu it was the updated version from VSCode.

I edited a model outside the models library and things worked!  Nice.  I then loaded a different model and the **Auto reload** box stayed checked.  When I saved the previous model in VSCode the new model reloaded and recompiled.  When I saved the new model in VSCode, nothing happened.  I unchecked and re-checked the **Auto reload** box and things started working as expected.

I made edits to the code in NetLogo, but did not save or compile them.  I then made edits in VSCode and saved them (different edits than done in NetLogo).  The model reloaded and recompiled, but that lost the work that was done in NetLogo.  I think the usual way to handle this is to prompt the user about what they want to do in order to avoid losing any unsaved work in progress.

kgmt0:
> It would be awesome if the watcher code could live outside of the code tab or other UI elements (mostly).

Do you have any suggestions on where should put the watcher code?

kgmt0:
@LaCuneta Thank you for your review. I think I've addressed all of your comments except for the issue with the Wolf Sheep Predation model, which I can't reproduce. I think the issue might be related to how the Auto Reload check box worked in my first implementation. Let me know if the bug is still reproducible for you.

netlogo-gui/src/main/app/codetab/FileWatcherThread.scala
```diff
@@ -0,0 +1,45 @@
...
+        val key: WatchKey = watchService.take
+
+        for (event <- key.pollEvents.asScala;
+             dirPath <- keyPathMap.get(key)) {
+          val eventPath: Path = event.context.asInstanceOf[Path]
+          val fullEventPath: Path = dirPath.resolve(eventPath)
+
+          // TODO: Don't reload if the event occurred as a result of NetLogo
+          // itself saving the model.
+          if(paths.contains(fullEventPath)) {
```
LaCuneta, line 31: Prefer space after `if` and before open paren.

LaCuneta (changes requested):
Excellent, this is getting closer.

I confirmed the issue with editing a models library model is fixed.  The file watcher stayed with the most recently opened model as expected.

The **External Change Detected** dialog worked for me, nice.

I opened a model, made edits in VSCode and saw the recompile in NetLogo as expected.  I then made edits in NetLogo with the model still open in VSCode.  When I saved the changes in NetLogo, VSCode updated to match, but then NetLogo popped up its **External Change Detected** dialog.

Edits to an included file in VSCode do cause a recompile in NetLogo, which is good, but then the includes file tab I had open for it is closed.

Edits to an included file in NetLogo with a save causes an auto-recompile that closes the includes file code tab.  I confirmed in 6.4 release that a save recompiles but leaves the includes file code tab open as expected.

When I save changes in VSCode and NetLogo recompiles it steals focus.  That's not the end of the world, but as someone who habitually hits Ctrl-S while working it would cause me to have to change my workflow a bit.  I think ideally NetLogo shouldn't steal focus when it's recompiling or alerting about a change, but I don't think it's a critical issue.

kgmt0:
@LaCuneta I think I managed to address all issues you raised except for the last one. On my computer (Linux with GNOME), NetLogo doesn't steal focus but the window does get raised. From my testing, it looks like Swing/AWT automatically raises windows when they're shown, and I wasn't able to find a way to stop it from doing that.

LaCuneta (changes requested):
This is really close!  Most functionality is working exactly as expected now.  Only two items found:

I add a button in the interface tab that calls a procedure that does not yet exist, do not save, make an edit to the code in an external editor, save it there, and it reloads in NetLogo without a prompt/warning about losing the work-in-progress in NetLogo.  Basically changes in interface or info that are not saved in NetLogo are not used to determine if the reload from disk dialog should fire.

Editing an nls file in NetLogo and an external editor, if I save in NetLogo it updates in VS Code.  If I change in VS Code and save in NetLogo reloads (and now stays open, which is good!), but it does not include the changes I just made.  Even more edits and saves cause more reloads, but all with the code that was loaded in NetLogo.  I know there are existing open issues with includes files having issues when saving, so this may be related to those, but I'm not 100% sure.  If this is basically an existing issue, I'm okay merging this PR without fixing it, but I do think that underlying issue should be prioritized to be fixed before this feature goes out; without it external edits of nls files is not useful.

brandesNW:
Hi Kritphong,
r.e. nls files - The code probably needs to raise an `ExternalFileSavedEvent` when an nls file has been changed externally and has been reloaded.

kgmt0:
> Hi Kritphong, r.e. nls files - The code probably needs to raise an `ExternalFileSavedEvent` when an nls file has been changed externally and has been reloaded.

Hi Aaron,

Thanks for the hint. I'm unable to get the NLS files to reload by raising `ExternalFileSavedEvent` though. I'm not sure if I'm doing this right, but I tried raising it from `Tabs` and it doesn't seem to do anything. I resorted to adding a new `reload` method in `TemporaryCodeTab`, which seems to work well enough.

kgmt0:
@LaCuneta The two issues you found should be fixed now. I also moved everything out of `MainCodeTab` and refactored the code to use `DirtyMonitor` for dirty file detection.

LaCuneta (approved):
Everything works as expected.  Thank you so much for this!

PR #2197, opened 2023-12-06
BehaviorSpace Extension

ERSUCC:
### Motivation

Right now, BehaviorSpace experiments can be run manually through the GUI, but not in NetLogo code. Being able to run experiments from NetLogo code would allow users to run multiple experiments back to back programmatically and analyze the results in the code.

### Features

It is now possible to create an experiment in code and set the various parameters available in the GUI. Multiple experiments can be created in the same run of a model and accessed or run by name. Experiments created in the GUI can also be run by name in code.

### Syntax

See README.md for more details on the use of this extension.


brandesNW:
This is a great start

Here are some suggestions for alternate command Names:

- bspace:set-name                  => rename-expt
- bspace:set-exit-condition    => set-stop-condition
- bspace:set-thread-count      => set-parallel-runs

brandesNW:
**I think we should consider adding commands to return the default and maximum recommended number of runs/threads**

- bspace:get-default-parallel-runs  *name*  or  bspace:get-default-thread-count *name*
- bspace:get-recommended-max-parallel-runs  *name* or  bspace:get-recommended-max-thread-count *name*

**I think we should consider adding Help commands** 
These would open the documentation in a browser

- bspace:goto-behaviorspace-documentation
- bspace:goto-bspace-extension-documentation

brandesNW:
** I think we should consider a small re-ordering of the primitives to correspond to their order in the BehaviorSpace GUI**

## Primitives

### Experiment Management

[`bspace:create-experiment`](#bspacecreate-experiment)
[`bspace:run-experiment`](#bspacerun-experiment)

### Experiment Parameters

[`bspace:set-name`](#bspaceset-name)
[`bspace:set-variables`](#bspaceset-variables)
[`bspace:set-repetitions`](#bspaceset-repetitions)
[`bspace:set-sequential-run-order`](#bspaceset-sequential-run-order)
[`bspace:set-metrics`](#bspaceset-metrics)
[`bspace:set-run-metrics-every-step`](#bspaceset-run-metrics-every-step)
[`bspace:set-run-metrics-condition`](#bspaceset-run-metrics-condition)
[`bspace:set-pre-experiment-commands`](#bspaceset-pre-experiment-commands)
[`bspace:set-setup-commands`](#bspaceset-setup-commands)
[`bspace:set-go-commands`](#bspaceset-go-commands)
[`bspace:set-post-run-commands`](#bspaceset-post-run-commands)
[`bspace:set-post-experiment-commands`](#bspaceset-post-experiment-commands)
[`bspace:set-exit-condition`](#bspaceset-exit-condition)
[`bspace:set-time-limit`](#bspaceset-time-limit)

### Experiment Run Conditions
[`bspace:set-spreadsheet`](#bspaceset-spreadsheet)
[`bspace:set-table`](#bspaceset-table)
[`bspace:set-stats`](#bspaceset-stats)
[`bspace:set-lists`](#bspaceset-list)
[`bspace:set-update-view`](#bspaceset-update-view)
[`bspace:set-update-plots`](#bspaceset-update-plots)
[`bspace:set-thread-count`](#bspaceset-thread-count)

netlogo-core/src/main/api/LabProtocol.scala
brandesNW, line 1: Was this moved/adapted from ProtocolEditable so that values can be set programatically?
We need to keep in mind that one diff between the programmatic version and the GUI version is that the user can start an experiment, or alter an existing one and then quit, which is not done with the programatic version.
We should probably have a bspace:delete-experiment command. Are programmatically create experiments added to the GUI experiment list? Are they checked for name conflicts? Are they saved when the model is saved?
We don't need to have these worked out for the first pass version, but will need to deal with them.

netlogo-gui/src/main/lab/gui/Supervisor.scala
brandesNW, line 1: the new variable *useGUI* deserves a comment. It may also be useful in places to add a comment indicating that code  is there to support the bspace extension

extensions/bspace/src/main/BehaviorSpaceExtension.scala
brandesNW, line 1: Something to keep in mind for the future: we probably don't want to keep the code for all the primitives in one file. If we add output file creating primitives and plot primitives they should certainly be in separate files.  Having multiple files will also be useful if you and Ryan are both working on the extension - there will be fewer opportunities for version conflicts.

brandesNW (changes requested):
There a some comments that require immediate changes, but some are things to think about.
Thanks again

netlogo-core/src/main/api/LabProtocol.scala
ERSUCC: Yes, it was moved from ProtocolEditable, because I needed to reuse the functionality for parsing variables without creating a new ProtocolEditable, because that would open a dialog and this needs to be headless. I will add a delete-experiment command. No, programmatically created experiments are not added to the GUI list, but they are checked for name conflicts (although not the other way around, I will fix that!). They are not saved when the model is saved, but they persist between runs until clear-all is run. They could be saved with the model if desired.

extensions/bspace/src/main/BehaviorSpaceExtension.scala
ERSUCC: I agree, but the problem is that each primitive needs access to the map of programmatically created experiments, which has to be defined in the BehaviorSpaceExtension class as far as I can tell. If you have any solutions, I would love to separate them into different files.

brandesNW: Maybe look at the LevelSpace (ls) extension. It does a lot of bookkeeping with models. Also View2.5D has special objects and maintains a list of views.

extensions/bspace/build.sbt
brandesNW, line 1: no need for *-beta1* in NetLogoVersion

extensions/bspace/README.md
brandesNW, line 1: It would be better if sentences were not unnecessarily split up.

extensions/bspace/README.md
```diff
@@ -0,0 +1,252 @@
...
+## Primitives
+
+### Experiment Management
+
+[`bspace:create-experiment`](#bspacecreate-experiment)
+[`bspace:run-experiment`](#bspacerun-experiment)
+
+### Experiment Modification
+
+[`bspace:set-name`](#bspaceset-name)
```
brandesNW, line 25: needs to be updated to *rename-experiment*

extensions/bspace/README.md
```diff
@@ -0,0 +1,252 @@
...
+[`bspace:set-setup-commands`](#bspaceset-setup-commands)
+[`bspace:set-go-commands`](#bspaceset-go-commands)
+[`bspace:set-post-run-commands`](#bspaceset-post-run-commands)
+[`bspace:set-post-experiment-commands`](#bspaceset-post-experiment-commands)
+[`bspace:set-repetitions`](#bspaceset-repetitions)
+[`bspace:set-sequential-run-order`](#bspaceset-sequential-run-order)
+[`bspace:set-run-metrics-every-step`](#bspaceset-run-metrics-every-step)
+[`bspace:set-run-metrics-condition`](#bspaceset-run-metrics-condition)
+[`bspace:set-time-limit`](#bspaceset-time-limit)
+[`bspace:set-exit-condition`](#bspaceset-exit-condition)
```
brandesNW, line 36: now *rename-experiment*

extensions/bspace/README.md
```diff
@@ -0,0 +1,252 @@
...
+[`bspace:set-metrics`](#bspaceset-metrics)
+[`bspace:set-variables`](#bspaceset-variables)
+[`bspace:set-thread-count`](#bspaceset-thread-count)
```
brandesNW, line 39: now *set-parallel-runs*

extensions/bspace/README.md
```diff
@@ -0,0 +1,252 @@
...
+### `bspace:run-experiment`
+
+```NetLogo
+bspace:run-experiment *name*
+```
+
+Run the experiment with the specified name. An error will be thrown if an experiment does not exist with the provided
+name.
+
+### `bspace:set-name`
```
brandesNW, line 65: all occurrences of renamed primitives need to be updated

extensions/bspace/README.md
```diff
@@ -0,0 +1,252 @@
...
+[`bspace:set-metrics`](#bspaceset-metrics)
+[`bspace:set-variables`](#bspaceset-variables)
+[`bspace:set-thread-count`](#bspaceset-thread-count)
+[`bspace:set-table`](#bspaceset-table)
+[`bspace:set-spreadsheet`](#bspaceset-spreadsheet)
+[`bspace:set-stats`](#bspaceset-stats)
+[`bspace:set-lists`](#bspaceset-list)
+[`bspace:set-update-view`](#bspaceset-update-view)
+[`bspace:set-update-plots`](#bspaceset-update-plots)
+
```
brandesNW, line 46: documentation needs to be added for

- goto-behaviorspace-documentation
- goto-bspace-extension-documentation
- get-default-parallel-runs
- get-recommended-max-parallel-runs

extensions/bspace/README.md
```diff
@@ -0,0 +1,252 @@
...
+bspace:set-pre-experiment-commands *name* *commands*
+```
+
+Set the pre-experiment commands for the specified experiment. An error will be thrown if an experiment does not exist
+with the provided name.
+
+### `bspace:set-setup-commands`
+
+```NetLogo
+bspace:set-setup-commands *name* *commands*
```
brandesNW, line 85: we should either give an example, or specify explicitly. In this case it is a string such as "setup" or “set random-seed 44 setup”

extensions/bspace/README.md
ERSUCC: I split up sentences in order to keep the 120 character line limit. Does that not apply to README files?

extensions/bspace/src/main/Information.scala
brandesNW, line 1: might be better to have 2 files with more descriptive names, e.g. GetBspaceExtensionDoc.scala and getParallelRunInformation.scala

extensions/gogo
brandesNW, line 1: this shouldn't be in the PR. not sure why it is, I think think you have the latest gogo code. Maybe it has to do with the initial PR.

ERSUCC: I noticed that, and the last commit I made should have removed it. Let me know if it still shows up for you.

extensions/bspace/src/main/Information.scala
ERSUCC: I'm not sure about having a whole other file just for one specific command, I think it's better to group by semantic category. Presumably there will be more get- reporters in the future, so I could separate those two and name the file something along those lines.

extensions/bspace/README.md
brandesNW: I think 120 limit is good. The problem was having two consecutive very short lines that are part of one sentence.

brandesNW: Really the doc is more important, and there that won't be problematic.

extensions/bspace/src/main/Information.scala
brandesNW: separating the get reporters would work. The name *Information.scala* is very generic. Ideally file names are somewhat informative.

brandesNW:
Updates to README are great.

The bspace.html file should go in the NetLogo/netlogo-gui/docs directory.
For the standard look and feel of extension doc see for example - NetLogo/netlogo-gui/docs/palette.html. That's probably not required for an unbundled extension, but it looks nice. At any rate things like the `extensions` statement should line up properly.

The Standard GUI bspace experiment update dialog was a good addition.

NetLogo hung when I tried to delete an experiment that didn't exist.
When I used the run-beta-experiment twice in a row I got the message "No experiment exists with the name beta3". 
If the experiment doesn't exist `get-return-value` might fail. We might need additional management to support the existence of multiple experiments.

LaCuneta (changes requested):
Very interesting extension, and nice job getting it all working.  I played around with it a bit and didn't run into any major issues.  I left a couple notes on items in the code.  Let me know if you have questions on anything, or we can meet to discuss.

Having the `behaviorSpaceReturnValues` live on the workspace feels odd to me.  Is there anywhere else it can be placed?  At a quick glance it looks like the `Supervisor` is handling the experiment running for the extension.  Would adding the fields there to read once the runs are complete make sense?  Also, see my below questions/comments on the feature in general.

Obviously the extension code itself will need to be removed from the PR to merge.  You can do that as an update to this PR or open a new one, whichever you find easier.

The below items are extension-specific, so they don't need to be addressed before we can merge the NetLogo code changes.

I see a lot of the prims take an experiment name to operate on.  It would be fitting with NetLogo style to have a `set-current-experiment` prim that would set the active experiment to work on, to be recorded as some global state in the extension, and then the prims wouldn't need to be "noisy" with the experiment name repeated.

The README seems to use the term `return` for the value that is produced by a reporter.  In NetLogo parlance we say `report` instead.  "Reports the recommended number of threads...", etc.

I think I understand the motivation behind the return reporters, to get data out of an experiment run, right?  But it will only return the values from the last run on the main workspace, is that correct?  I have a harder time seeing the value in that.  If I'm interested in a value from an experiment, wouldn't I be interested in it across many/all of the repetitions of it?  And also, if it only gives a value for the last run of the last experiment, couldn't I get the same info from the primary model itself?

Adding some language tests for the extension would be appreciated.  We've got [some docs on how to set that up using the Extension Plugin](https://github.com/NetLogo/NetLogo-Extension-Plugin#language-tests).  If added, then test runs can be enabled on GitHub Actions, too.

It would be really nice to be able to write actual code for the lists, reporters, and commands for the experiments instead of strings, so it'd be compile-time checked.  But I totally understand that the technical challenges would be large and it'd require a ton of refactoring.

netlogo-core/src/main/lab/Worker.scala
```diff
@@ -100,6 +100,12 @@ class Worker(val protocol: LabProtocol, val supervisorWriting: () => Unit = () =
   class Procedures(workspace: Workspace) {
     val preExperimentProcedure = workspace.compileCommands(protocol.preExperimentCommands)
     val setupProcedure = workspace.compileCommands(protocol.setupCommands)
+    for (command <- setupProcedure.code) {
```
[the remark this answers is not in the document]
ERSUCC: That makes sense. I don't think the run number would be 100% correct, and I just realized this one isn't 100% correct either. I can make a variable to keep track of the stack of experiments that were run and put it in the Workspace or Supervisor files, I'll experiment with that.

netlogo-gui/src/test/lab/gui/ProtocolEditableTests.scala
```diff
@@ -31,17 +31,4 @@ class ProtocolEditableTests extends AnyFunSuite {
       assert(protocol == editedProtocol)
     }
   }
-
-  test("protocol error on overflow") {
```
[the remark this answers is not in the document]
ERSUCC: After I changed some other files, this test started failing, but upon review I realized it was actually not written correctly, so for now I just removed it to make the tests pass. There are tests that it could be replaced with in the future, but they wouldn't be really equivalent.

ERSUCC:
> At a quick glance it looks like the Supervisor is handling the experiment running for the extension. Would adding the fields there to read once the runs are complete make sense?

They could be added there, but only as static fields. The only place they could be instance-specific would be the Workspace or World files. If you think it's fine to have them as static fields, I will move them there, along with any other current and future BehaviorSpace extension fields like return values.

> If I'm interested in a value from an experiment, wouldn't I be interested in it across many/all of the repetitions of it?

You make a good point. Do you think any sort of return values would be beneficial? If so, what format would you want? I can change it to return anything in theory.

brandesNW:
Thoughts on return values.
I agree that getting back return values from only one repetition is not so helpful, and that users should be able to collect run data perhaps by either run number or by parameter values plus repetition.

My initial motivation for the idea was a simple case in which there was something like a presentation I made once using the Fire Model. Suppose your goal is to find a threshold density at which the fire always burns to the end of the forest. If you set a density as the input parameter and get back a boolean you could use that to increase or decrease the density and do a binary search to find a threshold. 

Of course whether or not the fires burns through is stochastic, so you would actually want to run the experiment multiple times and collect data (either individually or by summation of 0/1) to find out whether the fire burns through using a statistical threshold. In this case you do want values for each run.

Similarly there may be cases in which you would want to retain data for multiple input parameters. Sensitivity analysis might be an example, I'm not familiar enough with the details to say for sure.

I agree that the return data should not be attached to the WorkSpace, since Workspaces are reused.

Should we store return data for only the most recent experiment, or for multiple experiments unless it is explicitly freed?

LaCuneta (commented):
Thanks for the updates!

One idea I had for the return values thing:  I remember discussing in a meeting having an extension primitive that could parse a generated output file to structured data in order to process/visualize it via NetLogo code.  If we added that functionality we could let the user define whatever values they wanted via the metrics and they can read it in after the runs are complete.  That avoids trying to anticipate when the user would want the data - per repetition or experiment-wide or even per measurement - they can decide that with their metrics and calculate whatever else they want.  This would also mean we wouldn't need to add any special API on the NetLogo side, it could be handled totally in the extension.

I see the one Depend failure as well.  Looks like `org.nlogo.workspace` depends on `org.nlogo.nvm`, which we definitely don't want `org.nlogo.api` depending on.  Let me know if you have trouble getting that untangled.

brandesNW:
The build failure seems to be related to a dependency problem.
See https://github.com/NetLogo/NetLogo/actions/runs/7310513550/job/19919196027?pr=2197#step:6:247

> check [api] independentOf [api-]
>   Unexpected dependencies found:
>   org.nlogo.api.Workspace
>     -> org.nlogo.workspace.WorkspaceFactory
> check [nvm] independentOf [nvm-]
>   Unexpected dependencies found:
>   org.nlogo.nvm.AnonymousProcedure
>     -> org.nlogo.api.Context
>       -> org.nlogo.api.Workspace
>         -> org.nlogo.workspace.WorkspaceFactory
> ...

It might be worth going back to the first push that failed.

LaCuneta (approved):
Looks good to me :+1:

PR #2200, opened 2023-12-15
Update CITATION.cff for 6.4.0

kgmt0:
6.4.0 has been out for a while now, but I noticed that `CITATION.cff` still says 6.3.0. This updates the version and release date in that file.


LaCuneta (approved):

LaCuneta:
Thanks for this!  I also added this to our build steps so hopefully it's not missed next time.

PR #2201, opened 2023-12-18
BehaviorSpace Spreadsheet Empty Data Fix

ryanp8:
Currently there is a bug where the spreadsheet is unable to create summary data on runs that do not have collected metrics because the specified conditions were not met. This bug occurs because data from the final step is not always collected. 

Now, data for the final step is collected for each run, regardless of whether the metric condition is satisfied. This behavior is also in line with what is in the documentation. I also added an additional check to make sure that the final step data is not counted twice if the condition does also happen to be satisfied.


brandesNW:
Great, this fixed the issue.

PR #2213, opened 2024-02-02
Separate code tab with compilation error. Fix 2104, Fix 2174

brandesNW:
Fixes the problem that separating the code tab when it contains a compilation error caused a runtime exception.

### DIFF
--- a/netlogo-gui/src/main/app/AppTabManager.scala
+++ b/netlogo-gui/src/main/app/AppTabManager.scala
@@ -277,12 +277,12 @@ class AppTabManager(val appTabsPanel:          Tabs,
       codeTabsPanelOption = Some(codeTabsPanel)
       codeTabsPanel.setTabManager(this)
 
-      // If the Selected Component in the App Tabs Panel was the CodeTab
-      // set it to be the Interface Tab, otherwise moving the CodeTab will
-      // cause swing to make the Info Tab the Selected Component
-      if (appTabsPanel.getSelectedComponent == appTabsPanel.mainCodeTab) {
-        appTabsPanel.setSelectedComponent(appTabsPanel.interfaceTab)
-      }
+
+      // The Selected Component in the App Tabs Panel will later need to be set
+      // to the Interface Tab if the the CodeTab is currently the Selected Component
+      // because Swing will make the Info Tab the Selected Componen after the Code Tab
+      // moves (since it will have the highest index).
+      val selectInterfaceTab = if (appTabsPanel.getSelectedComponent == appTabsPanel.mainCodeTab)  true else false
 
       // Move tabs from appTabsPanel to codeTabsPanel.
       // Iterate starting at last tab so that indexing remains valid when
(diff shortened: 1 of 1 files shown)


netlogo-gui/src/main/app/AppTabManager.scala
```diff
@@ -277,12 +277,12 @@ class AppTabManager(val appTabsPanel:          Tabs,
...
-      // cause swing to make the Info Tab the Selected Component
-      if (appTabsPanel.getSelectedComponent == appTabsPanel.mainCodeTab) {
-        appTabsPanel.setSelectedComponent(appTabsPanel.interfaceTab)
-      }
+
+      // The Selected Component in the App Tabs Panel will later need to be set
+      // to the Interface Tab if the the CodeTab is currently the Selected Component
+      // because Swing will make the Info Tab the Selected Componen after the Code Tab
+      // moves (since it will have the highest index).
+      val selectInterfaceTab = if (appTabsPanel.getSelectedComponent == appTabsPanel.mainCodeTab)  true else false
```
LaCuneta, line 285: I'm not sure the `if (...) true else false` is adding much here, I think you can just use the expression.

LaCuneta (changes requested):
I tested the scenarios from the issues, and both worked as expected with the new code.  I left one small comment in the code, after that change I think this is fine to merge.

netlogo-gui/src/main/app/AppTabManager.scala
brandesNW: I thought that initially, but evidently an if without a value and no else returns unit. So Scala is more verbose than NetLogo in this case.

With a value it returns AnyVal
https://www.baeldung.com/scala/conditional-expressions
The return type of an if expression is the closest supertype of the two branch expressions. 

An if expression without an else expression also returns a value. When there is no else expression, the return type is `Any`.

We can easily verify that in the Scala REPL:

> scala> val result = if (false) "failed"
> result: Any = ()

This explains 
https://stackoverflow.com/questions/45619279/what-is-the-return-type-of-a-scala-if-statement-without-else

`val someVal:String = if (someX == someY) n
`
The compiler error will be the following, which is indicating that you are returning a Unit when you are supposed to return String:

> error: type mismatch;
>  found   : Unit
>  required: String

That's because what you have is equivalent to this:

[val someVal: String = if (foo == bar) "Hello World" else ()](url)
() is the only value of [Unit](http://www.scala-lang.org/api/2.12.1/scala/Unit.html), which is not a valid value for a String type.

LaCuneta: I meant just switching to `val selectInterfaceTab = (appTabsPanel.getSelectedComponent == appTabsPanel.mainCodeTab)`, just the expression, no `if` at all

brandesNW: got it

LaCuneta (approved):
:+1: Good to merge once tests pass

PR #2215, opened 2024-02-08
Detached code tab loses focus during text search Fixes #1996

### LINKED ISSUE #1996: Detached code tab loses focus during text search
The symptoms differ between macOS and Windows. Not yet verified on Linux.

When the code tab is detached, text search (cmd-F on macOS, ctr-F on Windows and Linux) focus moves to the search box, and the code tab loses focus, as it should. On macOS the code tab goes behind the interface window if they overlap, which can obscure the text being searched. When the search box is exited, the interface window is selected and typing goes to the Command Center.

The steps to replicate are:
- Open any model
- Detach code tab such that the code tab is in front of the interface tab
- In the code tab window, press (cmd-F on macOS, ctr-F on Windows and Linux)
     - _On macOS the code tab window becomes hidden behind the interface window_
- Close the search box
- Type something
    - _Text will appear in the Command Center_

brandesNW:
See [Issue 1996](https://github.com/NetLogo/NetLogo/issues/1996) For details of the bug. The core of the issue is that if you open the Find Dialog while you are in a separate code tab and then close the dialog focus goes to the Command Center instead of back to the code tab.  In MacOS the separate code tab moves to the back if it overlaps with the main GUI window. Note neither the bug or the fix has been tested in Linux.

The underlying problem is that there is only one FindDialog and its parent is the GUI AppFrame.  If the FindDialog is opened from a separate code tab, when the dialog closes focus goes to its parent GUI AppFrame. Calls to requestFocusInWindow fail because the component you want focus set on must be in the window in focus.  

This code explicitly restores focus to the CodeTabContainer when appropriate.

The current code has a multiple dependency problems, for example.

> check [app.common] independentOf [app.common-]
>   Unexpected dependencies found:
>   org.nlogo.app.common.FindDialog$FindAction$1
>     -> org.nlogo.app.App
>     -> org.nlogo.app.AppTabManager
>     -> org.nlogo.app.AbstractTabsPanel
>     -> org.nlogo.app.CodeTabsPanel
>     -> org.nlogo.app.CodeTabContainer

The FindDialog is created by org.nlogo.app.App and is used by org.nlogo.app.infotab.InfoTab and elsewhere in the org.nlogo.app package.
Using classes and methods from the org.nlogo.app package causes the dependency issue. I think there is not an actual problem because the FindDialog is only created/initialized once, and is hidden between appearances.

### DIFF
--- a/netlogo-gui/src/main/app/common/FindDialog.java
+++ b/netlogo-gui/src/main/app/common/FindDialog.java
@@ -4,6 +4,13 @@ package org.nlogo.app.common;
 
 import java.awt.Frame;
 import java.awt.event.ActionEvent;
+import java.awt.event.WindowAdapter;
+import java.awt.event.WindowEvent;
+import java.awt.event.WindowListener;
+import java.awt.FocusTraversalPolicy;
+import java.awt.KeyboardFocusManager;
+import java.awt.Component;
+import java.awt.Container;
 import javax.swing.Action;
 import javax.swing.JDialog;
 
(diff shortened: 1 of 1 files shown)


brandesNW:
Closing because this PR creates a dependency error. A fix using a different approach should be a new PR

PR #2216, opened 2024-02-10
Fix auto-reload not respecting user prefs

kgmt0:
Auto-reload is getting enabled regardless of user prefs due to the `LoadModelEvent` handler in `Tabs` starting up watcher thread without checking user prefs first. This PR makes it so that the handler checks user prefs before starting the thread. I also included a commit that makes it possible to enable/disable auto-reload without restarting NetLogo.


LaCuneta (approved):
LGTM, thanks, and sorry I missed it in testing.

PR #2223, opened 2024-03-03
added _anyturtleson optimization

kodingkoning:


LaCuneta (changes requested):
Thank you for this!  The rewriter and the logic look correct, which is great.

I see you were already getting the dependency issues cleaned up.  I think the last bit is to change package used inside the files to `org.nlogo.prim` (without the `.etc`)

We have a recently started putting primitives in the `netlogo-core` folder which is shared by headless and GUI, to avoid duplication of files and code.  You should be able to move your optimized prim to `netlogo-core/src/main/prim` (no need for separate headless/GUI files).  Since you're moving the `turtles-on` prim also, it might be a good time to put that in the same location.

We do prefer all new code to be written in Scala.  It should be a pretty clear translation from Java, you can check other optimizations for reference - `_anywith.scala`, `_anyotherwith.scala`, maybe a few others.

When used with a turtle agentset that isn't empty, then it seems like it should just return true to me (no need to check the turtle's patch).  Does that seem right to you?  I don't know why someone would do `any? turtles-on turtles` or similar, but if they do we can optimize it for them.

I had a thought when looking at the code, "what would happen if we used a agentset of links?"  Interestingly, in 6.4.0, this causes an internal exception,  `java.lang.ClassCastException: class org.nlogo.agent.Link cannot be cast to class org.nlogo.agent.Patch (org.nlogo.agent.Link and org.nlogo.agent.Patch are in unnamed module of loader 'app')`.  I'd consider this a bug, but not with the code you wrote, obviously.  If you want to address this, it'd just be a matter of finding an I18N error message that would fit and raising it, similar to the `thatAgentIsDead` error in the code.  You could also add it to the existing `turtles-on` primitive.  Because this is an existing bug, if you'd prefer not to address this, that's fine I'll open an issue about it.

I checked for existing language tests that hit this code path `any? turtles-on ...` and didn't see any.  Because we now have this "special" code, I think having tests added would be a good idea. The file `test/commands/TurtlesOn.txt` seems like a good place to add them in, and some notes on language tests are in the wiki:  https://github.com/NetLogo/NetLogo/wiki/Language-tests  Just a few tests of positive/negative scenarios should be sufficient to guard regressions as we make changes.

If you have any questions on anything, just let me know.  Thanks again!

kodingkoning:
I was getting a variety of errors while trying to move the files around, so I went back to the hexy branch to try to pin the issues down. If I:

```
git mv ./netlogo-gui/src/main/prim/etc/_turtleson.java ./netlogo-core/src/main/prim/etc/
git rm ./netlogo-headless/src/main/prim/etc/_turtleson.java
```

then all tests still pass. However, if I then:

```
git mv netlogo-core/src/main/prim/etc/_turtleson.java netlogo-core/src/main/prim/_turtleson.java
```

as well as changing the package in that file to `package org.nlogo.prim;`, and changing `etc._turtleson` to `_turtleson` in `netlogo-gui/resources/system/tokens.txt` and `shared/resources/main/system/tokens-core.txt`, then I get errors. Specifically:

```
[info] TestAllSyntaxes:
[info] - commands
[info] - reporters *** FAILED ***
[info]   java.lang.ClassNotFoundException: org.nlogo.core.prim._turtleson
[info]   at java.base/jdk.internal.loader.BuiltinClassLoader.loadClass(BuiltinClassLoader.java:641)
[info]   at java.base/jdk.internal.loader.ClassLoaders$AppClassLoader.loadClass(ClassLoaders.java:188)
[info]   at java.base/java.lang.ClassLoader.loadClass(ClassLoader.java:520)
[info]   at java.base/java.lang.Class.forName0(Native Method)
[info]   at java.base/java.lang.Class.forName(Class.java:375)
[info]   at org.nlogo.core.TokenMapper.instantiate(TokenMapper.scala:34)
[info]   at org.nlogo.core.TokenMapper.$anonfun$getReporter$1(TokenMapper.scala:12)
[info]   at scala.Option.map(Option.scala:230)
[info]   at org.nlogo.core.TokenMapper.getReporter(TokenMapper.scala:12)
[info]   at org.nlogo.api.DelegatingMapper.$anonfun$getReporter$2(NetLogoLegacyDialect.scala:69)
[info]   ...
[info] Run completed in 3 seconds, 73 milliseconds.
[info] Total number of tests run: 2
[info] Suites: completed 1, aborted 0
[info] Tests: succeeded 1, failed 1, canceled 0, ignored 0, pending 0
[info] *** 1 TEST FAILED ***
[error] Failed tests:
[error] 	org.nlogo.compile.TestAllSyntaxes
[error] (netlogo / Test / testOnly) sbt.TestsFailedException: Tests unsuccessful
[error] Total time: 12 s, completed Mar 5, 2024, 6:08:44 PM
```

and

```
[info] TestAllTokens:
[info] - all listed primitives exist for NetLogoLegacyDialectTokenMapper$ *** FAILED ***
[info]   java.lang.ClassNotFoundException: org.nlogo.core.prim._turtleson
[info]   at java.base/jdk.internal.loader.BuiltinClassLoader.loadClass(BuiltinClassLoader.java:641)
[info]   at java.base/jdk.internal.loader.ClassLoaders$AppClassLoader.loadClass(ClassLoaders.java:188)
[info]   at java.base/java.lang.ClassLoader.loadClass(ClassLoader.java:520)
[info]   at java.base/java.lang.Class.forName0(Native Method)
[info]   at java.base/java.lang.Class.forName(Class.java:375)
[info]   at org.nlogo.core.TokenMapper.instantiate(TokenMapper.scala:34)
[info]   at org.nlogo.core.TokenMapper.$anonfun$getReporter$1(TokenMapper.scala:12)
[info]   at scala.Option.map(Option.scala:230)
[info]   at org.nlogo.core.TokenMapper.getReporter(TokenMapper.scala:12)
[info]   at org.nlogo.api.DelegatingMapper.$anonfun$getReporter$2(NetLogoLegacyDialect.scala:69)
[info]   ...
[info] - all listed primitives exist for ThreeDTokenMapper$ *** FAILED ***
[info]   java.lang.ClassNotFoundException: org.nlogo.core.prim._turtleson
[info]   at java.base/jdk.internal.loader.BuiltinClassLoader.loadClass(BuiltinClassLoader.java:641)
[info]   at java.base/jdk.internal.loader.ClassLoaders$AppClassLoader.loadClass(ClassLoaders.java:188)
[info]   at java.base/java.lang.ClassLoader.loadClass(ClassLoader.java:520)
[info]   at java.base/java.lang.Class.forName0(Native Method)
[info]   at java.base/java.lang.Class.forName(Class.java:375)
[info]   at org.nlogo.core.TokenMapper.instantiate(TokenMapper.scala:34)
[info]   at org.nlogo.core.TokenMapper.$anonfun$getReporter$1(TokenMapper.scala:12)
[info]   at scala.Option.map(Option.scala:230)
[info]   at org.nlogo.core.TokenMapper.getReporter(TokenMapper.scala:12)
[info]   at org.nlogo.api.DelegatingMapper.$anonfun$getReporter$2(NetLogoLegacyDialect.scala:69)
[info]   ...
[info] Run completed in 2 seconds, 357 milliseconds.
[info] Total number of tests run: 2
[info] Suites: completed 1, aborted 0
[info] Tests: succeeded 0, failed 2, canceled 0, ignored 0, pending 0
[info] *** 2 TESTS FAILED ***
[error] Failed tests:
[error] 	org.nlogo.headless.TestAllTokens
[error] (netlogo / Test / testOnly) sbt.TestsFailedException: Tests unsuccessful
[error] Total time: 12 s, completed Mar 5, 2024, 6:09:39 PM
```

Do you have any insight on these errors? I was not finding other references to `etc._turtleson`, except for in the Heatbugs tests (which is not what is causing these specific errors).

LaCuneta:
Yeah, I should've thought of this when I advised the file move for `_turtleson`, but it's matching parser class will also need to move from [the delightfully-named `java.scala` file](https://github.com/NetLogo/NetLogo/blob/hexy/parser-core/src/main/core/prim/etc/java.scala#L541-L546) to [the unhelpfully-named `misc.scala` file](https://github.com/NetLogo/NetLogo/blob/hexy/parser-core/src/main/core/prim/misc.scala).  Sorry about that!  Those files split the package name up, so even looking for `org.nlogo.core.prim.etc` wouldn't have found anything.  Let me know if that doesn't help.

LaCuneta:
This is looking good, thank you for the updates!  I think you may need to run the headless code dumper directly to get Heatbugs properly updated:  `headless/dump bench`.

Also, if you don't have time to add the language tests, that's fine we can merge without and I'll add them in later.

> I checked for existing language tests that hit this code path any? turtles-on ... and didn't see any. Because we now have this "special" code, I think having tests added would be a good idea. The file test/commands/TurtlesOn.txt seems like a good place to add them in, and some notes on language tests are in the wiki: https://github.com/NetLogo/NetLogo/wiki/Language-tests Just a few tests of positive/negative scenarios should be sufficient to guard regressions as we make changes.

kodingkoning:
I'll do the language tests next!

kodingkoning:
If I add the language tests to hexy, 2/3 of these failed tests still fail. The failures on this pull request are on `nogen; netlogo/Test/medium`, `threed; netlogo/Test/medium`, and `nogen; headless/Test/medium`. It is `nogen; headless/Test/medium` that passes on `hexy`.

For those first two, is this a problem with the test I wrote?

For the third one, the failure is with `any? turtles-on`, so it is specific to the optimization. I'm trying to figure out what's causing the problem, since the error is an `ArgumentTypeException`, though `ANY?` should be optimized out and not be passed arguments.

LaCuneta:
Thank you for the language tests!  I think the 3D failure is due to the `patch 0 0` and similar, as in 3D you need a z-coordinate.  You can append `_2D` to the test name and it will be skipped in 3D mode, and that seems fine here.  If you want to add a 3D version of the test (name ending in `_3D`), that'd probably be good as we have had optimization bugs affect 3D but not 2D.

For the other two tests, it is something to do with bytecode generation, but I'm not 100% sure where it's coming from.  I'll dig into it a bit and let you know.

netlogo-core/src/main/prim/_anyturtleson.scala
```diff
@@ -0,0 +1,45 @@
...
+
+package org.nlogo.prim
+
+import org.nlogo.agent.{ AgentSet, Patch, Turtle }
+import org.nlogo.core.{ AgentKind, I18N, Syntax }
+import org.nlogo.nvm.{ ArgumentTypeException, Context, Reporter, RuntimePrimitiveException }
+
+class _anyturtleson extends Reporter {
+  override def report(context: Context): java.lang.Boolean =
+    Boolean.box(report_1(context, argEvalAgentSet(context, 0)))
```
LaCuneta, line 11: So this is where the failure is happening.  The `argEvalAgentSet()` doesn't allow the single patch (or turtle) through.  In the bytecode version (`report_1`), the argument `agentOrSet` is just `AnyRef` so it runs fine.  What we want here, then, is `args(0).report(context)` to get the `AnyRef` value for the `report_1` method, and it can throw the error if it's not a valid type for the `match`.  I should've noticed this on an earlier review, but hey, this is why the language tests are useful!

kodingkoning:
There is still an error, but I pushed it because it is not clear to me yet what it is. In `TurtlesOn_2D`, I get `not equals(): reporter "count turtles-on n-of 2 turtles-on patch 2 3" Expected "[3]", but got "[6]"`. I think it's duplicating all of the turtles on that patch.

LaCuneta:
Yeah, that seems like a bug to me.  I can reproduce the behavior in 6.4.  Those tests don't appear to be necessary for this PR, so if you want to comment them out so we can merge, that's okay with me.  If you're interested in fixing [the bug in `_turtleson`](https://github.com/NetLogo/NetLogo/blob/hexy/netlogo-gui/src/main/prim/etc/_turtleson.java#L55-L59) we can look at that, too.

kodingkoning:
I'm glad to fix that bug. It will need to be changed for ` _turtleson`, `_anyturtleson`, and `_breedon`. I'll do it in this PR unless you think it's cleaner to do it in a separate one.

LaCuneta:
This PR is fine by me, thank you for fixing it!

LaCuneta (approved):
This looks good to me!  Do you have any other changes you're planning to include, or should we merge?

kodingkoning:
No other changes, so we can merge!

LaCuneta:
Thank you so much for your work!

PR #2235, opened 2024-04-05
Improve widget look and feel

ERSUCC:
- Converts interface to Swing FlatLaf
- Improved widget design
- Improves color picker, now uses RGB sliders instead of fixed color patches


SethTisue:
curious to see a screen shot showing the new look

brandesNW:
The vertical slider looks good when created, but something goes wrong when you change the value using the interface. see image
vertical slider bug

brandesNW:
Current version has an RGB color picker. we need to decide whether we are going back to the swatches version. Since the color chooser can only return NetLogo colors a new picker is probably not worth it, unless we want a different color chooser for Alpha, RGB. In that case we probably want an HSB picker option too.

ERSUCC:
I will look into the vertical sliders. There is already an HSB color picker, you should be able to switch modes within the color picker dialog. I would support changing NetLogo colors to Java colors across the board to keep the better color picker, if people don't disagree with that.

brandesNW:
Yes, the color picker in this version has both RGB and HSB. It should also have alpha or transparency.
We need feedback on the color issue but my guess is that we will keep the current one as an option because it has a long history.

--


From: Isaac Brown ***@***.***>
Date: Thursday, June 27, 2024 at 3:47 PM
Cc: Aaron Andre Brandes ***@***.***>, Comment ***@***.***>
Subject: Re: [NetLogo/NetLogo] Improve widget look and feel (PR #2235)

I will look into the vertical sliders. There is already an HSB color picker, you should be able to switch modes within the color picker dialog. I would support changing NetLogo colors to Java colors across the board to keep the better color picker, if people don't disagree with that.

ERSUCC:
Makes sense. Unfortunately I don't think it's compatible as an alternative option, it would have to be one or the other, so I'll just save the change locally and revert it on GitHub.

PR #2238, opened 2024-04-09
Fix auto-reload breaking on load errors

kgmt0:
If NetLogo encounters an error reloading the model, auto-reload can't be triggered again until the model is re-opened. This is because `Tabs` sets a flag to block auto-reloads if one is already in-progress. Since it never gets notified when an error occurs, the flag will remain set indefinitely. Fix this by defining a new event `LoadErrorEvent`, raise it when an error occurs, and unset the flag when the event is raised.

To test:
1. Make sure `Reload model on external changes` is enabled.
2. Open a model in NetLogo.
3. Clear the contents the file. NetLogo should display an error dialog.
4. Dismiss the dialog.
5. Restore the file.

Without the patch, NetLogo won't reload the model at step 5.

### DIFF
--- a/netlogo-gui/src/main/app/FileManager.scala
+++ b/netlogo-gui/src/main/app/FileManager.scala
@@ -20,7 +20,7 @@ import org.nlogo.awt.{ Hierarchy, UserCancelException }
 import org.nlogo.fileformat.{ FailedConversionResult, ModelConversion, SuccessfulConversion }
 import org.nlogo.swing.{ FileDialog, ModalProgressTask, OptionDialog, UserAction }, UserAction.MenuAction
 import org.nlogo.window.{ BackgroundFileController, Events, FileController, ReconfigureWorkspaceUI },
-  Events.{AboutToCloseFilesEvent, AboutToQuitEvent, AboutToSaveModelEvent, LoadModelEvent, ModelSavedEvent, OpenModelEvent }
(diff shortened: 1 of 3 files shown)


netlogo-gui/src/main/app/FileManager.scala
```diff
@@ -314,6 +314,8 @@ class FileManager(workspace: AbstractWorkspaceScala,
           case e: Exception => println("Exception in FileMenu.loadModel: " + e)
           None
         })
+    new LoadErrorEvent().raise(eventRaiser)
```
LaCuneta, line 317: This appears to raise the load error on every external change.  Shouldn't it only get raised on an actual error?  I also noticed that with these changes, when NetLogo reloads on an external changes it switches to the interface tab on a successful recompile.  I suspect that's related to this.  I at least confirmed the behavior on current `hexy` is to stay on the open code tab.

LaCuneta (changes requested):
Nice catch!  Glad to get this bug cleared before release.  I left a comment in the code with the issue I found in testing.

netlogo-gui/src/main/app/FileManager.scala
kgmt0: You're right! Somehow I completely missed that.

LaCuneta (approved):
:+1:

PR #2254, opened 2024-06-07
Input box bug fix

### LINKED ISSUE #1786: Empty input box crashes NetLogo
Hello,
I have been observing this problem from time to time in many version of NetLogo on OSX.
When I create input boxes in the interface and do not immediately give a value to them, at one point the cursor starts to blink and NL just crashes, without any possibility of saving the file. Clicking Halt does not solve the crash and one has to kill NL and loose the unsaved code. 

I have just suffered this problem in NL 6.1 on OSX 10.14.6, but it is definitely not limited to these versions. My students meet this problem quite frequently too during the workshops I give on ABMs with NL. If I can help you to locate this bug in anyway, please do not hesitate to contact me, because it is quite painful.

ERSUCC:
Fixes a bug where number input boxes infinitely regain focus if left empty. See issue #1786 for details. Fixes #1786.

### DIFF
--- a/netlogo-gui/src/main/window/InputBox.scala
+++ b/netlogo-gui/src/main/window/InputBox.scala
@@ -192,9 +192,6 @@ abstract class InputBox(textArea:AbstractEditorArea, editDialogTextArea:Abstract
             try inputText(inputType.readValue(InputBox.this.textArea.getText))
             catch {
               case ex@(_:LogoException|_:CompilerException|_:ValueConstraint.Violation) =>
-                org.nlogo.awt.EventQueue.invokeLater { () =>
-                  InputBox.this.textArea.requestFocus()
-                }
             }
             editing = false
           }


LaCuneta (approved):
I tested it out and this resolved the core problem of fighting input boxes for me.  The code looks good to me.  I was a little baffled on why that code existed in the first place.

Bu then a secondary issue appeared, in that now input boxes can have values that are invalid for their type.  So I now suspect that code was there to try to force the user to fix their bad value entry.  The most basic example is that if they are left blank, the backing variable value is actually `0` if I check it in the command center.  But if I enter an even weirder value, like "a-string", then that incorrect value remains when I click away from the input, even though the backing value of the variable is left unchanged.  This mismatch in what the UI shows versus what's in the model seems like it could cause confusion (and delay).

I can't think of another widget that allows you to enter values of the wrong type as a reference for what should be done here.  Sliders currently allow setting values that are out-of-bounds from code, but there is discussion on if that's actually okay.  Color input widgets reject changes from code that are not color values with an error.  You could just clear the input value to the last known good value when something bad is entered and forget about it.

Anyways, I'd be okay merging this and worrying about the other issue later on, but I'll let the team decide.

Issue #2054, opened 2022-10-07
The preference "Reload the last used model on startup" does not take account of whether that model is appropriate (2D or 3D).


brandesNW:
To reproduce:

1. Open NetLogo
2. Open a model
3. In the NetLogo->Preferences window ensure that the "Reload the last used model on startup" option is selected.
4. Open NetLogo 3D

It will open the previous NetLogo model, and then pop up a Dialog that says "You are attempting to open a 2D model in NetLogo 3D 6.3.0. You might need to make changes before it will work in 3D." You have the option to Cancel or Continue.

A similar problem arises if you open a 3D model in NetLogo 3D and then open NetLogo.

The appropriate behavior is for NetLogo to open the most recent .nlogo file and for NetLogo 3D to open the most recent .nlogo3d file.

LaCuneta:
Fixed in #2255

PR #2255, opened 2024-06-08
Improved logic for load last model on startup

ERSUCC:
It now filters by model type (2D/3D) before attempting to load. See #2054.


LaCuneta (approved):
LGTM :+1:  Thanks for the fix!

Issue #2131, opened 2023-07-26
Using a constant as a global variable name gives unclear closing bracket error


LaCuneta:
This is similar to #1604 , but different enough that I'd expect the fix to be different, too.

You can create a declaration like `globals [ a b c d e f g ]` in a NetLogo model, and the compiler error you will see is "closing bracket expected" with the `e` highlighted.  This, of course, doesn't make any sense and it takes some thinking to realize what's going on.  This only affects constants, so I believe just `e` and `pi`.  Actual reporters like `max-pxcor` give a more reasonable "There is already a primitive called..." error.

ERSUCC:
@LaCuneta See pull request #2259, I made some changes to fix this and some related issues. #1604 still seems to be an issue, so I'll check that out next.

PR #2259, opened 2024-06-13
Made misplaced constant error message more descriptive

ERSUCC:
Issue #2131 describes a strange error message being printed when a constant such as `e` or `pi` is used as a global variable name. After further investigation, I found that the same incorrect error message was being printed in multiple other places, like in the extensions statement or in function input lists. The changes in this pull request should fix all such places, now displaying a more descriptive error message about the name conflicting with a constant.

### DIFF
--- a/parser-core/src/main/parse/StructureChecker.scala
+++ b/parser-core/src/main/parse/StructureChecker.scala
@@ -10,13 +10,33 @@ package org.nlogo.parse
 
 import
   org.nlogo.core,
-    core.{ BreedIdentifierHandler, StructureDeclarations, Token },
+    core.{ BreedIdentifierHandler, StructureDeclarations, Token, TokenType },
       StructureDeclarations.{ Breed, Declaration, Extensions, Identifier, Includes, Procedure, Variables },
     core.Fail._
 
 import SymbolType._
 
 object StructureChecker {
+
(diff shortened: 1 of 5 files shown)


LaCuneta (changes requested):
Nice, I didn't directly test the code yet (I'll load it up soon), but the changes look good to me.  

One thing I know we'd want is some unit tests for these scenarios.  I see a few similar ones in [`StructureParserTests.scala`](https://github.com/NetLogo/NetLogo/blob/hexy/parser-core/src/test/parse/StructureParserTests.scala#L173-L193), so that might be a good spot to add a few more.

ERSUCC:
I added tests for the new error messages. I noticed that none of the StructureParser error messages are in the translation property files, should that be changed or are they supposed to only be in English?

LaCuneta:
@ERSUCC If you want to go ahead and migrate them to the I18N files that'd be great.  I think leaving them in code as English-only was just an oversight.  The compiler/parser seems to use a [simplified namespace specifier in those files](https://github.com/NetLogo/NetLogo/blob/hexy/shared/resources/main/i18n/Errors_en.properties#L195-L202).

LaCuneta (approved):
Tested it out, works great.  Thank you!

Issue #2111, opened 2023-05-17
Disable error-checking on tab switching


arcpaolo:
Whenever the user switches tabs (e.g., from code to info, or when there are include files), NetLogo automatically tries to do error-checking, and moves the focus where errors are found. 

This is extremely annoying when writing code with include files, or when adding interface elements. It is also dangerous because it highlights allegedly incorrect code, which can lead to accidental deletion. Also the insertion point can be quite far from the spot being edited (for instance a long block without a closing bracket will move the focus all the way to the opening bracket).

This feature should be removed or at least create an option to disable it. Checking for errors with the manual checkbox, when trying to run, and when saving should be more than adequate.

Note that a discussion was started on this topic on the NetLogo community mailing list, and several people agreed this is really annoying.

Thanks.

SethTisue:
> a discussion was started on this topic on the NetLogo community mailing list,

link?

arcpaolo:
https://groups.google.com/g/netlogo-users/c/FljzTaDXKNs/m/NxLIFg7NBAAJ

arcpaolo:
And a further related request: error-checking should also be disabled (or at least optional) on save.

I am in the middle of making some substantial changes to my code, where I know that I may have to write dozens of lines of code before it will work (meaning before it will become error-free). Sometimes I want to save the stuff I have typed as I go along. When I do this, it invariably takes me away from where I was typing to some other random place where the error is occurring, often in a different tab altogether.

But even with a single-file program this is super-annoying. For example, I am in the process of making several changes to variable names in a <breeds>-own block: creating some new variables, removing some old ones, renaming some, moving things around to make it more coherent.

If I am in the middle of this and I hit "command-save" (which I am prone to do frequently out of habit), the error checker will immediately drag me to some random part of the code where one (I am guessing the first instance) of the removed/renamed variables happens to be. Now I have no idea where I am, I have to navigate back to the <breeds>own statement and figure out where I was and what I was doing. Even with a single file (which I often have hundreds of lines of code), this is really really really annoying. Enough so that I am taking time in the middle of a gnarly coding project to write these comments here.

PLEASE TURN OFF AUTOMATIC ERROR CHECKING OR AT LEAST MAKE IT CONFIGURABLE!

ERSUCC:
[...]

PR #2261, opened 2024-06-13
Added property for focusing tabs on error

ERSUCC:
Users have reported an annoying bug-like feature where tabs steal focus every time an error occurs during compilation (see #2111). This pull request adds a preference option that if unchecked would prevent compilation errors from stealing focus from the current tab. It defaults to being checked so that it is not a breaking change.


brandesNW (approved):
good to go

Issue #2263, opened 2024-06-14
Surprising behavior when overwriting files


kgmt0:
I noticed a surprising behavior when NetLogo is asked to overwrite an existing file. When the target file doesn't end with the appropriate extension (.nlogo or .nlogo3d), NetLogo will create a new file with the right extension using the target filename as the base. For example, if it's asked to overwrite a file called `foo.txt`, the actual file that gets written will be `foo.txt.nlogo`. When the user selects the file, they will be asked if they want to overwrite the file (at least that's the case on Linux), so writing to a different file is pretty surprising.

Moreover, if `foo.txt.nlogo` already exists, it will get overwritten instead, which could lead to unintentional data loss. NetLogo does show a second warning dialog in this case, but it's pretty easy to miss that it's a different file since only the file extension is different.

I think the right behavior here is to overwrite the file that the user picked. Optionally, a warning dialog saying that the file extension is incorrect could be displayed before actually overwriting the file.

### Steps to Reproduce
1. Create a file not ending in `.nlogo` or .`nlogo3d`, e.g. `foo.txt`.
2. Launch NetLogo and press Ctrl + S to save.
3. Choose the file created in step 1 using the file picker.
4. A warning dialog asking if the file should be overwritten should appear. Press the appropriate button to accept.
5. Observe how the chosen file doesn't get overwritten as claimed by the warning dialog. Instead, a new file is created using the chosen filename as the base.

brandesNW:
Hi Kritphong,
I noticed part of this behavior when trying to save a 2D model opened in 3D.
I think we have good reasons for wanting the models files to end in .nlogo or .nlogo3d.
I hadn’t noticed the overwrite behavior you pointed out.
Since this is an unusual case, and there is an appropriate warning, I don’t think a change is needed, but I am interested whether anyone else thinks it should be done.

  *   Aaron

--
Aaron Brandes, Software Developer
Center for Connected Learning and Computer-Based Modeling


From: kgmt0 ***@***.***>
Date: Friday, June 14, 2024 at 2:55 AM
To: NetLogo/NetLogo ***@***.***>
Cc: Subscribed ***@***.***>
Subject: [NetLogo/NetLogo] Surprising behavior when overwriting files (Issue #2263)

[...]

PR #2266, opened 2024-06-15
Fix file overwrite issue

ERSUCC:
This pull request adds a file type filter to the Save As dialog as a fix for issue #2263.


brandesNW:
Thanks for this work.
What happens now is that the user is not able to select the file, but the name of the file without the suffix appears. So if the user clicks on "myfile.txt"  the word "myfile" appears. Is this what we want, or should nothing appear?

Issue #2253, opened 2024-06-04
Merge `__includes` files on export to NetLogo Web


LaCuneta:
When the **Save As NetLogo Web** feature is used, only the code from the main code tab is included, `__includes [...]` files are ignored.  Until NetLogo Web fully supports `__includes` (or their successor), it would be useful to at least merge the full model code in the NetLogo Web file so it can run without further changes.

An alternative would be to just add a separate feature to **Save Merged Model** or similar to the **File** menu when `__includes [...]` is used, which could then be separately uploaded to NetLogo Web to be run.

ERSUCC:
See pull request #2267 for a potential temporary solution. It works well when given inputs that are expected, but as mentioned in the pull request description, it is not particularly robust and could potentially break the exported model.

PR #2267, opened 2024-06-15
Added included file merging for export as NetLogo Web

ERSUCC:
As a temporary fix until included files work in NetLogo Web, all included files are now merged with the main code file before exporting as NetLogo Web, as requested in issue #2253. A warning dialog is displayed before the merge, and the start of each included file is marked with a comment in the merged main code file. The current code will work for recursive includes, but it will break the exported model if the text `__includes` is present in any unexpected places such as string literals.


brandesNW:
I used the attached files and got the html file which did not work.
Please check out.
[cumulative.zip](https://github.com/user-attachments/files/15959817/cumulative.zip)

brandesNW:
looks good

Issue #2050, opened 2022-09-30
Logging can throw an exception when chosen directory does not exist


LaCuneta:
With the new logging system setup via the NetLogo preferences, you can select a file location to store the logs that does not exist.  When you do so, NetLogo crashes with an unhelpful exception and can leave a "ghost" splash screen process running.  The logging feature should create a directory if it doesn't exist (including any needed parents), and if for some reason the directory can't be created (permissions issues, etc) then it should fail gracefully with a meaningful error message.

ERSUCC:
Fixed in PR #2269.

PR #2269, opened 2024-06-16
Made logging errors more user friendly

ERSUCC:
As requested in issue #2050, the process of selecting a log directory is now smoother. When you first enter a path that doesn't exist and click OK or Apply, it asks you whether you want to create the path or not, only allowing the preferences to save and close if you confirm. It also ensures that the specified path exists when a model loads, displaying a warning dialog if anything goes wrong.


brandesNW:
would “Error writing to the specified log directory. You might not have write permission. Logs will not be written for this NetLogo session.”
be clearer to the user than the current message
"Error writing to the specified log directory. Logs will not be written for this NetLogo session.”

Issue #1932, opened 2021-07-07
unexpected plotting (bug)


alan-isaac:
In the Interface, add a plot widget to a new model.
Enter the following at the command line.

    clear-all
    plot-pen-up
    plotxy 1 1

A point is clearly plotted. According to the documentation, it should not be.

ERSUCC:
Hi @alan-isaac, this bug is now fixed, see pull request #2276. It turned out that plot pens were always drawing regardless of up/down status, not just in the example you gave.

PR #2276, opened 2024-06-22
Fixed plot pen bug

ERSUCC:
Attempted to fix bug in issue #1932.

Note: This pull request would require a change to the format which would no longer be consequential if pull request #2285 is integrated, so it should be kept as a draft until then.


ERSUCC:
Currently failing a few tests, most likely because some models previously expected there to be something plotted (erroneously) and it is no longer plotted after fixing the bug. Those models' tests and checksums will need to be fixed.

Issue #1881, opened 2020-12-07
“__includes”  with no brackets causes exception


brandesNW:
Open a New model
Add the line `__includes`(no brackets)
Check the green compile mark.

resulting info and stack trace:

Did not find expected open bracket for __includes declaration at position 2147483647 in
 at org.nlogo.core.Fail$.exception(Fail.scala:27)
 at org.nlogo.core.Fail$.exception(Fail.scala:25)
 at org.nlogo.core.Fail$.exception(Fail.scala:23)
 at org.nlogo.parse.StructureParser$.findIncludes(StructureParser.scala:155)
 at org.nlogo.parse.FrontEndMain.findIncludes(FrontEnd.scala:58)
 at org.nlogo.parse.FrontEndMain.findIncludes$(FrontEnd.scala:56)
 at org.nlogo.parse.FrontEnd$.findIncludes(FrontEnd.scala:10)
 at org.nlogo.compile.Compiler.findIncludes(Compiler.scala:163)
 at org.nlogo.app.codetab.CodeTab.getIncludesTable(CodeTab.scala:132)
 at org.nlogo.app.codetab.CodeTab$$anon$3.$anonfun$addControls$1(CodeTab.scala:100)
 at org.nlogo.app.codetab.IncludedFilesMenu.updateVisibility(IncludedFilesMenu.scala:33)
 at org.nlogo.app.codetab.IncludedFilesMenu.handle(IncludedFilesMenu.scala:30)
 at org.nlogo.window.Events$CompiledEvent.beHandledBy(Events.java:223)
 at org.nlogo.window.Event.doRaise(Event.java:198)
 at org.nlogo.window.Event.raise(Event.java:122)
 at org.nlogo.window.CompilerManager$.$anonfun$$lessinit$greater$default$4$1(CompilerManager.scala:20)
 at org.nlogo.window.CompilerManager$.$anonfun$$lessinit$greater$default$4$1$adapted(CompilerManager.scala:20)
 at org.nlogo.window.CompilerManager.raiseEvent(CompilerManager.scala:34)
 at org.nlogo.window.CompilerManager.compileProcedures(CompilerManager.scala:230)
 at org.nlogo.window.CompilerManager.compileAll(CompilerManager.scala:173)
 at org.nlogo.window.CompilerManager.handle(CompilerManager.scala:137)
 at org.nlogo.window.Events$CompileAllEvent.beHandledBy(Events.java:185)
[...]

PR #2277, opened 2024-06-22
Fixed uncaught exception when brackets missing after `__includes`

ERSUCC:
Fixed bug detailed in issue #1881.


brandesNW:
This fixes the exception problem. However it is not possible to save the file. In other cases with compilation errors the file can still be changed.

brandesNW:
Nice

Issue #1832, opened 2020-04-07
Bug on closing NetLogo Window?


cstaelin:
In NetLogo 6.1.1 running under Windows 10, I believe there is a bug in both NetLogo and NetLogo3D that occurs when there are included .nls files AND the NetLogo window is closed without first saving the model.  If a change has been made to an .nls file and the window is closed, “x’ed” out of, NetLogo asks if you wish to save the altered .nls file(s), and proceeds to do so if you so indicate.  However, if you have changed a variable on the Interface tab of the model, but not the code on the primary code tab, then the .nlogo  or .nlogo3D file is not saved and any changes to the variables on the interface tab are lost.  Apparently changing a variable on the Interface tab does not prompt NetLogo to flag the .nlogo  (.nlogo3D) file as having been updated, and thus needful of being saved.

I've attached the files that I've tested this out on.  If you change something in `extra-code` and the value of `chooser-test`, then close the window, extra-code.nls is saved, by not SaveTest.nlogo.

[SaveTest.zip](https://github.com/NetLogo/NetLogo/files/4441685/SaveTest.zip)

brandesNW:
I have verified that this bug occurs in NetLogo and NetLogo3D.

The obvious workaround is to save the model.

brandesNW:
Hi Charles,
I have verified the bug, and noted it in issue 1832.
Thanks for pointing this out.

Aaron

--
Aaron Brandes, Software Developer
Center for Connected Learning and Computer-Based Modeling

brandesNW:
This note breaks out the bug into three manifestations, which I have verified.
Aaron
[IncludedFileBug.zip](https://github.com/NetLogo/NetLogo/files/4633972/IncludedFileBug.zip)


I'm sorry, my first message mixed two different issues. Here I separate
them.


-------- Forwarded Message --------
Subject: Unexpected behavior: included files
Date: Thu, 30 Apr 2020 19:30:45 -0700
From: Steve Railsback <Steve@LangRailsback.com>
To:

Hi--

I have been having trouble understanding why changes to an included code
file seem to get reversed, and think I understand what is going on. Here
are two behaviors that seem dubious.

(I am using NetLogo 6.0.4, but did not see anything about these issues
in the release notes for subsequent versions.)

1) When I open an included file, make a change in it, then go to the
Interface or code tab, then go back to the included file's code tab,
NetLogo lets me close the included file without asking whether I want to
save its changes. So it's easy to close the included file without saving
[...]

PR #2278, opened 2024-06-22
Fixed widget edits not causing save prompt

ERSUCC:
Fixed issue #1832. Editing the value of a widget without using the Edit dialog now causes the model to be dirty and prompts for save on closing.


brandesNW:
Note, this fixes the issue raised by cstaelin. The issue mentioned in the email is now covered by issue https://github.com/NetLogo/NetLogo/issues/2283

Issue #1734, opened 2019-04-09
__includes with full path fails on Mac


alan-isaac:
I do not have access to a Mac, so I will do my best to report a problem encountered by my students, who use 6.0.4.  This only affected Mac users.  My students store some .nls files I give them in a specific location.  If they try to use `__includes` with a full path, starting with a tilde for their home directory, NetLogo fails to find the files.  (I checked that the paths they were using were correct.)  Is this a known issue?

charlesres:
@alan-isaac, thanks for submitting a report. With paths to .nls files, paths that are accepted are absolute paths ("/home/user/netlogo.nls") or relative paths ("./netlogo.nls"). I don't believe NetLogo expands tildas to the home directory, unlike bash and other terminals, which could be the issue you are having. A workaround would be to provide the path as mentioned above.

SethTisue:
that sounds plausible. `~` is Unix shell syntax, it doesn't exist at either the JVM level or the OS level

qiemem:
@charlesres you are correct; this is a common issue in Java path handling. Given the ubiquity of using `~` in *nix, it might be worth handling.

Here's a variety of possible solutions: https://stackoverflow.com/questions/7163364/how-to-handle-in-file-paths

brandesNW:
fixed by https://github.com/NetLogo/NetLogo/pull/2282

PR #2282, opened 2024-06-26
Added tilde parsing for included file paths

ERSUCC:
Adds the enhancement requested in issue #1734. Should work on any platform.

### DIFF
--- a/parser-core/src/main/parse/StructureParser.scala
+++ b/parser-core/src/main/parse/StructureParser.scala
@@ -41,7 +41,7 @@ object StructureParser {
           firstResults
         else {
           Iterator.iterate(firstResults) { results =>
-            val suppliedPath = results.includes.head.value.asInstanceOf[String]
+            val suppliedPath = resolveIncludePath(results.includes.head.value.asInstanceOf[String])
             cAssert(suppliedPath.endsWith(".nls"), IncludeFilesEndInNLS, results.includes.head)
             includeFile(compilationEnvironment, suppliedPath) match {
               case Some((path, fileContents)) =>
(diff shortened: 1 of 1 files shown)


parser-core/src/main/parse/StructureParser.scala
```diff
@@ -159,11 +159,14 @@ object StructureParser {
...
           .collect {
-            case s: String => s
+            case s: String => resolveIncludePath(s)
           }.toSeq
     }
   }
 
+  def resolveIncludePath(path: String) =
+    path.replaceFirst("^~", System.getProperty("user.home").replace("\\", "/"))
+
```
brandesNW, line 169: On MacOS I was able to successfully use
__includes [ "~/ModelsNetLogo/save-nls/set-parameters.nls" ]
which is what the enhancement request asked for.

I thought that the last replacement would make the following work as well:
__includes [ "~\ModelsNetLogo\save-nls\set-parameters.nls" ]

However it gave the error "Illegal character after backslash

as did
__includes [ "\Users\aab1638\ModelsNetLogo\save-nls\set-parameters.nls" ]

brandesNW (changes requested):
Please respond to backslash issue.
Does it make sense to allow tilde substitution for Windows
From what I understand tilde is used in PowerShell, but not  at the command line.
What testing have you done in Windows?

parser-core/src/main/parse/StructureParser.scala
ERSUCC: The NetLogo/Java format expects the user to use forward slashes, since backslashes are reserved for escaping characters such as double quotes. The backslash replacement code I added was only for the `user.home` property, since on Windows it returns a string with backslashes, but it needs to be forward slashes for NetLogo and Java. I did this instead of restricting the platform, because it seemed like an unnecessary restriction to me, but if you think it shouldn't be allowed on Windows, then I can fix the code to do that instead.

brandesNW:
concerns were addressed

PR #2284, opened 2024-06-26
Update when a changed include file is closed with no save

ERSUCC:
EDIT: Integrated with pull request #2286


brandesNW:
a separate code tab can no longer be edited.

PR #2334, opened 2024-12-09
XML File Format

TheBizzle:
This is a replacement of #2285 so that @ERSUCC and I can collaboratively push to this branch in this repository.

Some things that should be discussed and addressed:

- [x] Move this out of `parser-core`?
    * This is *a* parser, but it's not a part of a compiler; it should go into `netlogo-core` (or where ever the old `ModelReader` was) or become its own subproject
- [x] Unit tests
    * We should have some tests for this, even in just a basic sense
      * Empty model
      * Filled-out model
      * NetLogo 3D
      * Roundtrip loading, then saving, and make sure that the contents are (semantically) the same
- [x] Docs for the new file format
  * [This page](https://github.com/NetLogo/NetLogo/wiki/File-%28.nlogo%29-and-Widget-Format) has been an invaluable resource for understanding the old file and widget format; we should have an equivalent for the new formats
  * The working copy of the XML format specification can be found [here](https://github.com/NetLogo/NetLogo/wiki/XML-File-Format).
- [x] Remove Image widget
    * This widget was introduced for testing the resource bundles, but I don't really know of a use case for it, and it wasn't approved for inclusion in NetLogo
    * Relevant files for removing this are: 
      * `interfacepanel.ImageWidget`
      * `interfacepanel.CoreImage`
      * `window.ImageWidget`
      * `window.InterfaceFactory`
      * `window.Widget`
      * `core.Widget`
      * `WidgetXMLLoader`
      * Possible the `image` line in to `window.Properties` (not sure what that line does)
- [x] Create a `resource` extension and sample model to demo the functionality
  * An initial draft of the extension can be found [here](https://github.com/NetLogo/Resource-Extension).


parser-core/src/main/core/WidgetXMLLoader.scala
```diff
@@ -0,0 +1,379 @@
...
+
+        val attributes =
+          Map( "left"      -> input.left.toString
+             , "top"       -> input.top.toString
+             , "right"     -> input.right.toString
+             , "bottom"    -> input.bottom.toString
+             , "multiline" -> input.multiline.toString
+             , "type"      -> typeName
+             ) ++
+          ifDefined(input)("variable", _.variable)
```
TheBizzle, line 357: How can this field (`variable`) ever be undefined?  I get that they might leave the textbox blank, but you can't have a valid (i.e. functioning) Input widget without it being tied to a global variable.  The same goes for the Switch, Slider, and Chooser widget-writers in this file.

netlogo-gui/src/main/properties/ResourcePathEditor.scala
```diff
@@ -0,0 +1,89 @@
...
+  private class ResourceMenu extends JPanel {
+
+    private val label = new JLabel(emptyName)
+
+    private var resourceOpt: Option[ExternalResource.Location] = None
+
+    setBorder(BorderFactory.createLineBorder(Color.BLACK))
+
+    add(label)
+    add(new JLabel("v")) // placeholder until gui redesign is merged
```
TheBizzle, line 27: :eyes:

The comment should go, and/or the functionality should be changed to something that we're happy (enough) with.

netlogo-core/src/main/fileformat/NLogoFormat.scala
```diff
@@ -130,6 +132,16 @@ trait AbstractNLogoFormat[A <: ModelFormat[Array[String], A]] extends ModelForma
     }
   }
 
+  def readExperiments(source: String, editNames: Boolean, existingNames: Set[String]): Try[(Seq[LabProtocol], Set[String])] =
+    Failure(new Exception) // fix after compile
```
TheBizzle, line 136: :eyes:

netlogo-core/src/main/api/ModelLoader.scala
```diff
@@ -1,14 +1,35 @@
...
 import scala.reflect.ClassTag
+import scala.util.{ Failure, Try }
+
+trait AbstractModelLoader {
+  def readModel(uri: URI): Try[Model]
+  def readModel(source: String, extension: String): Try[Model]
+  def save(model: Model, uri: URI): Try[URI]
+  def sourceString(model: Model, extension: String): Try[String]
+  def emptyModel(extension: String): Model
+  // these next two allow ManagerDialog to use the correct format for experiment loading/saving (IB 8/17/24)
```
TheBizzle, line 20: Please sign your comments "Isaac B" instead of "IB".  Our policy changed on this (particularly after Jeremy and I had conflicting initials).

netlogo-core/src/main/fileformat/NLogoXMLLoader.scala
```diff
@@ -0,0 +1,328 @@
...
+              model.map(_.copy(turtleShapes = el.getChildren("shape").map(ShapeXMLLoader.readShape)))
+            case (model, el @ XMLElement("linkShapes", _, _, _)) =>
+              model.map(_.copy(linkShapes = el.getChildren("shape").map(ShapeXMLLoader.readLinkShape)))
+            case (model, XMLElement("previewCommands", _, commands, _)) =>
+              val section = new Section("org.nlogo.modelsection.previewcommands", PreviewCommands.Custom(commands))
+              model.map((m) => m.copy(optionalSections = m.optionalSections :+ section))
+
+            // case "systemDynamics" =>
+            //   optionalSections = optionalSections :+
+            //     new Section("org.nlogo.modelsection.systemdynamics.gui", SDMXMLLoader.readDrawing(element))
```
TheBizzle, line 132: Is it intentional that this never got implemented?  If so: We should probably remove these lines.  If not: We should implement it.

netlogo-core/src/main/fileformat/NLogoXMLLoader.scala
```diff
@@ -0,0 +1,328 @@
...
+
+    for (section <- model.optionalSections) {
+      section.key match {
+        case "org.nlogo.modelsection.previewcommands" =>
+          writer.writeStartElement("previewCommands")
+          writeCDataEscaped(writer, section.get.get.asInstanceOf[PreviewCommands].source)
+          writer.writeEndElement()
+
+        // case "org.nlogo.modelsection.systemdynamics.gui" =>
+        //   writeXMLElement(writer, SDMXMLLoader.writeDrawing(section.get.get.asInstanceOf[AnyRef]))
```
TheBizzle, line 203: Same as above, regarding "implement it or delete".

parser-core/src/main/core/WidgetXMLLoader.scala
ERSUCC: I saw that it was an Option so I assumed that it would be safer to check if it's None so the parser doesn't crash randomly. If it's definitely the case that it would never be None, this can certainly be removed, I just wasn't aware that was an invariant.

netlogo-gui/src/main/properties/ResourcePathEditor.scala
ERSUCC: There is the perfect class for this in the GUI PR, and I would really like to use it here. Is it ok to make a note of that change elsewhere and remove the comment for now?

parser-core/src/main/core/WidgetXMLLoader.scala
ERSUCC: Actually, upon further investigation, I discovered that the variable could actually be None, in the case of a HubNet widget. It seems like you can place a widget in the HubNet client editor without assigning a variable to it, but correct me if I'm wrong about that.

netlogo-core/src/main/fileformat/NLogoXMLLoader.scala
```diff
@@ -0,0 +1,318 @@
...
+          import Model.{ defaultCode, defaultShapes, defaultLinkShapes }
+
+          val version = element("version")
+          val snapToGrid = element("snapToGrid", "false").toBoolean
+
+          val settings  = new Section("org.nlogo.modelsection.modelsettings", ModelSettings(snapToGrid))
+
+          val model = Model(defaultCode, List(), defaultInfo, version, defaultShapes, defaultLinkShapes, List(settings), Seq(), Seq())
+
+          element.children.foldLeft(Try(model)) {
```
ERSUCC, line 110: @TheBizzle The way you refactored this makes it fail every time, because you can't instantiate the Model class without a View widget. I will revert this to what I had before so that it works as expected, unless you have another solution.

ERSUCC: Nevermind, I realized it's probably fine to just instantiate it with a default View widget and assume it will be replaced when the widgets element gets parsed.

netlogo-core/src/main/fileformat/NLogoXMLLoader.scala
TheBizzle: I thought we discussed this and arrived at the conclusion that the serialization/deserialization code could live in the SDM package, and then just be exposed through an API.  What came of that idea?

parser-core/src/main/core/WidgetXMLLoader.scala
TheBizzle: RIP :skull_and_crossbones: 

Good catch.

netlogo-core/src/main/fileformat/NLogoXMLLoader.scala
ERSUCC: I remember there being issues, but that was a long time ago. I'll revisit it.

netlogo-core/src/main/api/XMLUtils.scala
```diff
@@ -0,0 +1,136 @@
...
+
+package org.nlogo.core
+
+import java.io.{ StringReader, Writer }
+import javax.xml.stream.{ XMLInputFactory, XMLOutputFactory, XMLStreamConstants, XMLStreamException }
+
+import scala.util.{ Failure, Try }
+import scala.util.matching.Regex
+
+object XMLUtils {
```
TheBizzle, line 11: A few small things here....

The file is `XMLUtils.scala`, and then it has `object XMLUtils` and `class XMLWriter`.  All of those terms should probably be the same.

If you have the `CDATA_ESCAPE` thing in the companion object for the class (i.e. `object XMLWriter` and `class XMLWriter`), you get the added benefit of not needing to prefix it as `XMLUtils.CDATA_ESCAPE` (so you can just reference it as `CDATA_ESCAPE`).  A `val` inside of a companion object is equivalent to having a `static` member on a class in Java.

Lastly (and very minorly), all-caps is unidiomatic for Scala constants.  Go with `CDataEscape` or `CDATAEscape` instead.

TheBizzle:
The wiki page for the file format looks super-good now, thank you! :star_struck:

netlogo-core/src/main/api/XMLUtils.scala
ERSUCC: If you didn't already see this, XMLUtils.scala contains both the XMLWriter and XMLReader class, both of which access the escape variable, so I didn't want it to only live in an object definition for XMLWriter. If you have a suggestion for a better name, I'm open to changing it.

ERSUCC:
I moved everything out of `parser-core`, except `ExternalResource.scala`, which is used in `Model.scala` (in `parser-core`). If I put `ExternalResource.scala` in `netlogo-core` (even without changing its package) it doesn't compile.

TheBizzle:
@ERSUCC, I came upon a bit of a bug with the resource manager window.  I tried to resize it (since I found the window a bit small), but increasing the size of the window just... increases the size of the window, without increasing any font sizes or anything.  Shrinking the size of the window also just hides content, rather than scaling it down.

On the positive side of things, I've published v1.0.7 of the Import-A extension, which now allows us to see the real use case of the resource manager in action, and your example model works.

I also added the `resource:list` primitive to the extension, which will list the names of all resources in the bundle.

ERSUCC:
> @ERSUCC, I came upon a bit of a bug with the resource manager window. I tried to resize it (since I found the window a bit small), but increasing the size of the window just... increases the size of the window, without increasing any font sizes or anything. Shrinking the size of the window also just hides content, rather than scaling it down.
Good to know, but I would rather not spend too much time on GUI issues right now, since it will all have to change after integrating the new GUI anyway. I'll make a note to revisit this then and ensure it scales properly.

TheBizzle:
@ERSUCC, regarding testing of this PR, it looks to me like we do not, in fact, have the kinds of tests of the new format that I'm looking for.

  * `netlogo-core/src/main/fileformat/ThreeDViewReader.scala`
    * This file is definitely operating on the old format

  * `netlogo-core/src/test/fileformat/NLogoLabFormatTest.scala`
    * Relies on `test/fileformat/BehaviorSpaceExample.xml`
    * Relies on `test/fileformat/BehaviorSpaceMultiples.xml`
    * This one might be fine if the XML format for BehaviorSpace is still the same as it used to be, in the new file format
    
  * `netlogo-gui/src/test/workspace/OpenModelTests.scala`
    * Relies on `test/fileformat/non-nlogo-extension.txt`
    * Relies on `test/fileformat/WidgetSection.txt`
    * These test files are using the old serialization, so I'm not sure how this is working
    * It might be good to preserve that functionality, but should also test with the new format

  * `netlogo-core/src/test/fileformat/HubNetWidgetReadersTests.scala`
    * Clearly uses the old format

  * `netlogo-core/src/test/fileformat/NLogoFormatTests.scala`
    * Again, we might want to preserve these tests for the old format, but we should have something like this for `.nlogox`, too
    * Much of this is pretty clearly just using the old format (especially in `NLogoFormatIOTest`)
    * I've checked, and the temp file output in the roundtripping test for Ants Benchmark is definitely outputting the old file format

ERSUCC:
Noted, I'll add some tests tomorrow.

TheBizzle:
It's not showing up properly here, but this has been merged now.  Big props to @ERSUCC for all of his work on this! 🎉

PR #2339, opened 2025-02-05
Fixed tilt-up/tilt-down floating point error

ERSUCC:
Reasonably often, a pull request will fail due to a floating point error when the `tilt-up` or `tilt-down` primitives are called with 0 as an input. This pull request adds a check in those primitives to ensure that nothing changes if the input is 0.


ERSUCC:
Closing, decided to not be an issue.

Issue #2329, opened 2024-11-10
Incorrect error message in command center


ERSUCC:
I thought that maybe I could define a global in the command center, so I tried to run `globals [ test ]`. I would have expected it to either define a global called `test` or throw an error such as "The globals command cannot be used in the command center". What actually happened was an error saying "END expected". If I never used `to`, why would `end` be expected? This should be resolved. For reference, it seems to also report the same nonsensical error for `extensions` and `__includes`.

SethTisue:
iirc the REPL wraps user-entered code in a `to...end`

you'll get the same error message if you put this in the Code tab

```
to foo
globals [test]
end
```

brandesNW:
Thanks for the clarification Seth

PR #2343, opened 2025-02-15
Improved error message for incorrect keyword usage

ERSUCC:
Inspired by issue #2329, this PR adds a more helpful error message when a keyword such as `to` or `globals` is used in the wrong context. The improved error message is shown in the command center and in the code tab.


brandesNW:
Code looks good. This needs to be merged with hexy. rather than rebase (because there have been so many changes to hexy) it might be better to create a new branch off of hexy and then copy or merge the two files.

Issue #2307, opened 2024-08-30
Inconsistent behavior when double-clicking to select a word in code editors


TheBizzle:
### Steps to reproduce

  * Write `apples-oranges-grapes` in the "Code" tab
  * Double-click on `oranges`
  * Create a plot and write `apples-oranges-grapes` in the single-line pen code editor
  * Double-click on `oranges`

### What should happen

  * Both editors make the same selection

### What actually happens

  * The "Code" tab editor will select only `oranges` (my preferred behavior)
  * The plot code editor will select `apples-oranges-grapes`

### Notes

Similarly, doing `Ctrl+Shift+Left` arrow will select all the way to the beginning of `apples-oranges-grapes`, instead of the beginning of `oranges`.  And the same for `Ctrl+Backspace`.  So the two editors seem to have different ideas of how word boundaries are defined.

Issue #2304, opened 2024-08-30
Ctrl+Del deletes too much


TheBizzle:
### Steps to reproduce

  * Be on Windows/Linux ("Delete" means "Backspace" on Mac keyboards, last I checked)
  * Create a line of code: `let apples 3`
  * Put the cursor between `let` and `apples`
  * Press `Ctrl+Del`

### What should happen

  * `apples` is deleted

### What actually happens

  * `apples 3` is deleted

### Notes

It seems to delete to the end of the line, which is not the normal behavior for `Ctrl+Del` in Windows editors, and is inconsistent with our own behavior for `Ctrl+Backspace` deleting a single word to the left, and `Ctrl+Shift+RightArrow` selecting only the word to the right.

ERSUCC:
I attempted a fix, see PR #2318. Let me know if it still doesn't work as expected; there may be some bugs at the moment since I had to implement it myself because the editor component we currently use doesn't provide functionality for deleting the next word.

TheBizzle:
Unfortunately, this fix doesn't change the behavior at all.

The current editor has the right behavior (in the Code tab) for *selecting* words to the right, so it's unexpected to me that it wouldn't have ability to delete words to the right.

Also interesting: The code editor for *plot pens* does the correct thing (i.e. deleting the word to the right, rather than deleting the entire line).  So I feel like NetLogo's codebase has the technology to do this... and we're just simply doing the wrong thing in these specific circumstances, for some reason.

ERSUCC:
It is indeed strange that the current editor doesn't provide this behavior, but I have triple checked the docs and cannot find it anywhere. Intriguing that it works correctly for plot pens, I'll investigate that.

ERSUCC:
I discovered that it works correctly not just in the plot pens editor, but in any code editor except the Code Tab, because the Code Tab is the only one that uses a third party text editor component which overrides the default key mappings. Unfortunately, it seems that the third party overrides are not actually added to the list of key mappings that can be edited, they seem to be dealt with internally, so I can't find a way to remove the one for Ctrl+Del.

ERSUCC:
Ok, I believe I actually fixed it this time, see the above PR. Let me know if it still doesn't work.

PR #2344, opened 2025-02-16
Standardized text shortcut behavior

ERSUCC:
This PR addresses issues #2307 and #2304. The following shortcuts should now work as intended in the code tab and in other text editors.

- Previous word
- Select previous word
- Next word
- Select next word
- Select word
- Delete previous word
- Delete next word


TheBizzle:
This was merged, thank you! :bow: :tada:

PR #2346, opened 2025-02-16
Added check for missing report in to-report

ERSUCC:
Previously, procedures defined with `to-report` would compile successfully without a `report`, but a runtime error would be thrown once the procedure finished. This PR adds a compile-time check to ensure that a `report` is present somewhere in the procedure. It will also compile successfully if `run` or `runresult` is used in the procedure, because those primitives can cause the procedure to report a value without explicitly calling `report`.


TheBizzle:
Seems pretty good.  But I'd also consider the following to be a reporter definition that shouldn't throw a compilation error:

```netlogo
to-report apples
  error "TODO"
end
```

Could you expand this to cover that case?

ERSUCC:
Good thinking. I added that case.

Issue #2308, opened 2024-08-30
Destructuring bind should work infinitely-many levels deep


TheBizzle:
### Steps to reproduce

  * Write the code `[[x y] z a] [[5 54] 23 12]`

### What should happen

  * You can destructure nested lists as you would expect
    * `x`: 5
    * `y`: 54
    * `z`: 23
    * `a`: 12

### What actually happens

> ERROR: Nothing named Z has been defined.

ERSUCC:
What's the use case for this? Why can't you just do `let [ x y z a ] [ 5 54 23 12 ]`?

TheBizzle:
In real code, the function producing the list and the function destructuring the list are two separate things.  I could have flattened it (and that's probably what I ended up doing—I've forgotten, at this point), but it's annoying because NetLogo is really not helping me to model my data the way that I want to.  I want my function to return multiple values, but NetLogo doesn't provide tuples, so I have to put them into a list.  I want a bunch of related data (e.g. coordinates) to be bundled together into an object or a tuple, but NetLogo doesn't have those, so they're modeled as another list inside of the first list.

This seems to me like a reasonable thing to do want to do:

```netlogo
let [[corner1-x corner1-y corner2-x corner2-y] player-name score] get-current-player-info
```

It's just very arbitrary on our part that the pattern matching should work only one-level deep, so as to require:

```netlogo
let [bounding-box player-name score] get-current-player-info
let [corner1-x corner1-y corner2-x corner2-y] bounding-box
```

or that I need to model my data in a particular way in Function B, knowing that I'm going to want to destructure it in Function A.

ERSUCC:
Ah ok, I didn't know you could do that. It should definitely be allowed, I'll see if I can fix it.

TheBizzle:
Fixed in  #2348

PR #2348, opened 2025-02-18
Infinitely nesting destructuring binds

ERSUCC:
This PR addresses issue #2308 for both `let` and `set`.


TheBizzle:
Very cool, thank you!

Issue #2355, opened 2025-03-09
NetLogo hangs moving through a corner with wrapped coordinates and pen-down


dkondor:
This seems to be the duplicate of #926, but occurs with version 6.4.0 (64-bit, Linux).

Steps to reproduce:
 1. create a world which wraps both horizontally and vertically
 2. set origin to "corner"
 3. `pen-down`
 4. try to move a turtle from `patch 0 0` to `patch max-pxcor max-pycor`
 5. NetLogo hangs, need to force quit ("Tools -> Halt" does not work)

Some observations so far:
 - this only occurs with pen down
 - only occurs if the origin is in a corner (setting it to center fixes the issue)
 - #2293 might be the same issue

Attachments: 
corner_issue1.nlogo -- minimal model demonstrating the issue
corner_issue2.nlogo -- same model with origin set to center (no issue)

[corner_issue.zip](https://github.com/user-attachments/files/19149867/corner_issue.zip)

SethTisue:
nice catch!

ERSUCC:
I have confirmed that this issue is still present in the latest development branch.

ERSUCC:
@dkondor I fixed the bug you reported, and it will hopefully make its way into the next release. Let us know if you encounter another similar bug that isn't fixed by this change.

dkondor:
Thank you for looking into this so quickly! I'll try to test and report back any further issues (although I might not have the time to compile from source, so I might only do so with the next release).

ERSUCC:
We'll be releasing an official Beta version of the next release pretty soon, so you should be able to test it then without building from source.

TheBizzle:
Fixed by [13fa9f2](https://github.com/NetLogo/NetLogo/pull/2358/commits/13fa9f215cc4380bd6a670b43c3dcf4eb7d6ca5c).

PR #2358, opened 2025-03-12
Fixed infinite hang in turtle trail drawer

ERSUCC:
Fixed the bug detailed in issue #2355.


SethTisue:
Could there be a test case?

ERSUCC:
Sure, although I'm not aware of the root cause of the issue, so it would only cover the 4 specific cases that have been identified so far.

PR #2072, opened 2022-12-21
CI: Add Windows and macOS build jobs, use setup-java

EwoutH:
Some upgrades to the CI!

- Adds Windows and macOS jobs to the build matrix
- Uses [actions/setup-java](https://github.com/actions/setup-java) instead of [olafurpg/setup-scala](https://github.com/olafurpg/setup-scala) to setup Java and sbt
- Clone the submodules directly in [actions/checkout](https://github.com/actions/checkout) instead of a separate step
- Always use the latest [Liberica 17](https://bell-sw.com/pages/downloads/#/java-17-lts) version

This is a first step in #2071.

**TODO (help welcome):**
- [ ] Make sure the test jobs run as soon as the Ubuntu build job is finished, and don't wait until the Windows and macOS jobs are also finished (to speed up workflow). [More details here](https://stackoverflow.com/questions/74881221/ci-github-actions-require-only-one-specific-job-in-matrix-to-finish-for-othe).

**Optional, can also be done in other PR:**
- [ ] Figure out useful artifacts to upload for each OS (`.exe` for Windows, `.dmg` for macOS? Are those already created, if so where?).
- [ ] Reduce the size of the cache that get's uploaded from the Ubuntu build job for the tests jobs (currently it's the whole directory, that probably can be way less).

PR #2370, opened 2025-03-26
Updated GitHub actions

ERSUCC:
This pull request updates and expands on PR #2072. The main actions file as well as the actions file in each submodule are now updated to use the latest versions of everything. In addition, the JDK has been switched from Liberica to Eclipse Temurin, due to packaging issues on Mac with Liberica.

Note: a few of the submodules have existing PRs open for GUI-related issues, so for this PR, the new `actions` branch in each of those submodules was created based on the most recent commit in the corresponding PR. It might be good to merge those PRs before this one and then rebase the corresponding `actions` branches, so that the submodule references in `hexy` don't get messed up.

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -21,7 +21,7 @@ jobs:
 
       - uses: actions/setup-java@v4
         with:
-          distribution: temurin
+          distribution: 'temurin'
           java-version: '17'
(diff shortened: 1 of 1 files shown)


ERSUCC:
This PR is now bundled with #2544.

Issue #2377, opened 2025-04-04
Support running the models test suite on individual models


jzkelter:
Currently we have support for running all the model tests on all the models with `test` from sbt and for running specific tests on all the models with `testOnly *NameOfTest`. However, we have no way of running all the tests on a specific model. It would be very useful to be able to do this, because running all the tests takes a few minutes.

ERSUCC:
This was resolved by PR [#414](https://github.com/NetLogo/models/pull/414) of the models repo.

Issue #414, opened 2013-08-16
5.1 regression: error handling at EOF in StructureParser


SethTisue:
these tests fail on master due to the StructureParser rewrite:

``` scala
  test("missing end 1") {
    val e = intercept[CompilerException] {
      compile("to foo to bar")
    }
    assertResult("END expected")(e.getMessage.takeWhile(_ != ','))
  }
  test("missing end 2") {
    val e = intercept[CompilerException] {
      compile("to foo fd 1")
    }
    assertResult("END expected")(e.getMessage.takeWhile(_ != ','))
  }
  test("missing end 3") {
    val e = intercept[CompilerException] {
      compile("to foo")
    }
    assertResult("END expected")(e.getMessage.takeWhile(_ != ','))
  }
```

I should add passing tests like these on 5.0.x, then merge the tests to master and fix the regression there

currently on master in some cases you get a NoSuchElementException (instead of a CompilerException) and in others you get a misleading message like e.g. "expected identifier". StructureCombinators needs more fallback `fail()` calls

PR #2382, opened 2025-04-14
Per-model testing for the models library

ERSUCC:
Addresses issue #2377. The following syntax can be used to filter models by name, ignoring case: `testOnly * -- -Dmodel=name`. If the name has spaces, wrap the argument in double quotes, as follows: `testOnly * -- "-Dmodel=name with spaces"`. Due to the restrictive nature of SBT's testing system, it doesn't seem that there's a more elegant syntax, unfortunately.


ERSUCC:
These changes have already been merged, by PR [#414](https://github.com/NetLogo/models/pull/414) of the models repo.

Issue #1654, opened 2018-07-17
Provide some way to namespace `.nls` file imports


Idloj:
_From @TheBizzle on June 14, 2018 15:21_

_Copied from original issue: Idloj/NetLogo#6_

ERSUCC:
This will likely be resolved by PR #2386. Mentioning it here so this issue can be closed if that is the case.

PR #2386, opened 2025-04-19
Implement the new module system

kgmt0:
This is still work-in-progress, but you can use it to load NLS files now. The `alias` option that allows you to control how names are prefixed is also working.


ERSUCC:
Do you have documentation or a sample model to demonstrate the new syntax?

ERSUCC:
@kgmt0 I played around with it for a bit, and so far I like the way it works and it seems very promising! Here are some issues I noticed:

1. If you import a library and then just use its name without the colon and a command after it, the error message is "Nothing named xyz has been defined." I would expect it to be something along the lines of "xyz is a library, not a command."
2. Imported libraries don't show up under the Included Files dropdown. This is probably up for debate but personally I would like it if they showed up there so I could easily open them.
3. `alias` is used like a keyword as far as I can tell, but the compiler does not identify it as a keyword. This means that it isn't green and bolded like `__includes` or `libraries`, and you can use `alias` as an alias.
4. Including a library in itself causes infinite recursion and the app freezes. This should be caught before it happens and a relevant error should be thrown.
5. Included libraries are not private to the file in which they are included, which seems undesirable to me. For example, if I include library `test` in the Code tab, and I include library `test2` in library `test`, I can now access library `test2` from the Code tab.
6. No error is thrown if you include the same library twice. Similarly, no error is thrown if you include a library once, and then include a different library that is aliased to the same name as the first library. This latter case is problematic, because both libraries are then included, and any command name conflicts become ambiguous. After messing around with it, I discovered that in that case, the version of the command from the first included library is used, and all other commands with the same name are silently ignored.

Let me know if I need to clarify anything I said.

kgmt0:
Thanks for testing!

>     1. If you import a library and then just use its name without the colon and a command after it, the error message is "Nothing named xyz has been defined." I would expect it to be something along the lines of "xyz is a library, not a command."

Hmm. I didn't consider having a special error message for that because libraries are not (currently) a value, but I can see how users would find it helpful. Would love to hear what other people think as well.

>     2. Imported libraries don't show up under the Included Files dropdown. This is probably up for debate but personally I would like it if they showed up there so I could easily open them.

I agree. I plan to implement this but haven't got to it yet.

>     3. `alias` is used like a keyword as far as I can tell, but the compiler does not identify it as a keyword. This means that it isn't green and bolded like `__includes` or `libraries`, and you can use `alias` as an alias.

It's a bit different from other keywords because it only has a special meaning within the context of the libraries syntax. There's no problem if the user wants to name something "alias", or even use the word "alias" as an alias.

>     4. Including a library in itself causes infinite recursion and the app freezes. This should be caught before it happens and a relevant error should be thrown.

Good catch! I'll fix that.

>     5. Included libraries are not private to the file in which they are included, which seems undesirable to me. For example, if I include library `test` in the Code tab, and I include library `test2` in library `test`, I can now access library `test2` from the Code tab.

There's a (not yet implemented) `define-libraries` syntax that you'll be able to use to list exactly the symbols you want to export. Once that's implemented, I plan to make symbols private by default.

>     6. No error is thrown if you include the same library twice. Similarly, no error is thrown if you include a library once, and then include a different library that is aliased to the same name as the first library. This latter case is problematic, because both libraries are then included, and any command name conflicts become ambiguous. After messing around with it, I discovered that in that case, the version of the command from the first included library is used, and all other commands with the same name are silently ignored.

Good point. I'll add some checks for that.

kgmt0:
> Do you have documentation or a sample model to demonstrate the new syntax?

I sent you a link to the design document on Slack.

ERSUCC:
Related issue: #1654

netlogo-core/src/main/compile/api/Backifier.scala
```diff
@@ -7,6 +7,6 @@ import org.nlogo.{ core, nvm }
 import scala.collection.immutable.ListMap
 
 trait Backifier {
-  def apply(procedures: ListMap[String, nvm.Procedure], c: core.Command): nvm.Command
-  def apply(procedures: ListMap[String, nvm.Procedure], r: core.Reporter): nvm.Reporter
+  def apply(procedures: ListMap[Tuple2[String, Option[String]], nvm.Procedure], c: core.Command): nvm.Command
```
ERSUCC, line 10: Is there a reason why you use `Tuple2[T, U]` instead of `(T, U)`?

shared/resources/main/i18n/Errors_en.properties
```diff
@@ -201,6 +201,11 @@ compiler.MultiAssign.emptyVarList = The list of variables names given to {0} mus
 compiler.MultiAssign.tooFewValues = The list of values for {0} must be at least as long as the list of names.  We need {1} value(s) but only got {2} from the list {3}.
 compiler.Backifier.replaced = {0} is no longer a primitive, use {1} instead.
 compiler.StructureParser.includeNotFound = Could not find {0}
+compiler.StructureParser.importNotFound = Could not find {0}
+compiler.StructureParser.importMultipleImports = Attempted to import a module multiple times
+compiler.StructureParser.importLoop = Found a loop in the module import chain
```
ERSUCC, line 206: Since `loop` is already a thing in NetLogo, it might be better to just reuse the `importMultipleImports` property here. Maybe change the wording slightly, something like "Module {0} has already been imported".

netlogo-core/src/main/compile/api/Backifier.scala
kgmt0: Not really. I saw `Tuple2` being used somewhere else so I thought I would do the same thing.

ERSUCC: I defer to @TheBizzle for the final call, but my understanding is that the parenthesis syntax is more idiomatic, and it's what the repo currently uses the most. Not a huge deal, but it may be less familiar to people than the parenthesis syntax.

TheBizzle: Yeah, syntactic sugar preferred.

TheBizzle:
Awesome work, TYVM! :tada:  :sunglasses:  :+1:

Issue #1823, opened 2020-03-10
move NetLogo to new Vector implementation in 2.13.2?


SethTisue:
https://github.com/scala/scala/pull/8534 will likely land soon and be included in Scala 2.13.2

NetLogo's `LogoList` is just a wrapper around `scala.collection.immutable.Vector[AnyRef]`, so the performance differences should carry over directly to NetLogo

some of the improvements give huge speedups on very long lists. but perhaps in NetLogo such huge lists aren't that common

but also, another change in the new Vector is that small sizes are special-cased. I would expect even ordinary NetLogo models to benefit from that — if they use lists at all, I mean. they can expect to benefit in speed, but at least as important, they will use benefit in lower memory usage. users running with large number of agents, where each agent uses small lists, could see a noticeable drop in RAM usage

and it would be interesting to run the NetLogo benchmark suite before-and-after this change

but.... huge caveat.... you'd have to move to Scala 2.13 first. curious, is any work on that planned or in progress?

SethTisue:
Scala 2.13.2 is now available https://github.com/scala/scala/releases/tag/v2.13.2

SethTisue:
Scala 2.13.3 is out, so the new `Vector` is now pretty well battle-tested

EwoutH:
@SethTisue Did you already try running [scala-collection-compat](https://github.com/scala/scala-collection-compat) over the repository? If so, what where the results?

SethTisue:
@EwoutH I think you might be misunderstanding the purpose of scala-collection-compat. it's useful for using certain 2.13 features from 2.12, but we aren't staying on 2.12 and we don't need those features. I don't see that scala-collection-compat is useful for upgrading to 2.13

brandesNW:
Hi Seth,
What steps need to be taken to upgrade to scala 2.13.2 ?
What version of scala do you think we should aim for (if we have the resources) for the next release (or the one after).
- Aaron
--
Aaron Brandes, Software Developer
Center for Connected Learning and Computer-Based Modeling


From: Seth Tisue ***@***.***>
Date: Thursday, November 17, 2022 at 10:14 AM
To: NetLogo/NetLogo ***@***.***>
Cc: Subscribed ***@***.***>
Subject: Re: [NetLogo/NetLogo] move NetLogo to new Vector implementation in 2.13.2? (#1823)

[...]

PR #2394, opened 2025-04-26
Update to Scala 2.13.16

ERSUCC:
This PR updates the main repo and all bundled extensions to Scala 2.13.16.

TODO:
- [x] Update NetLogo Language Library to Scala 2.13 so that the Python extension can be updated
- [ ] Update models repo to Scala 2.13
       - Will need to look into the use of Scala Chart, it does not yet provide a version for Scala 2.13
- [ ] Notify maintainers of non-bundled extensions to update to be compatible with the main repo changes
- [ ] Investigate possible code enhancements and optimizations available with Scala 2.13


SethTisue:
It seems to me that this is essential work to ensure the longterm future of the NetLogo project.

SethTisue:
it so happens that I'm something of an expert on the differences between Scala 2.12 and 2.13 :-)

so, if you run into trouble, don't hesitate to reach out. the bible is of course https://github.com/scala/scala/releases/tag/v2.13.0

a related issue is https://github.com/NetLogo/NetLogo/issues/1823 , about the new `Vector` implementation; I would hope NetLogo users would notice lower memory usage and perhaps some extra speed from the switch

(but note that since I'm no longer directly involved with NetLogo, I know nothing about when this might be mergeable, for example because it affects any plugins, controlling API apps, etc which are themselves written in Scala)

SethTisue:
ecosystem-wise, if plugins have to rebuilt anyway, you might think about whether Scala 3 is an even more desirable target

but regardless of what decision you make there, getting onto 2.13 is the right way to begin

ERSUCC:
> a related issue is #1823 , about the new `Vector` implementation; I would hope NetLogo users would notice lower memory usage and perhaps some extra speed from the switch

Is the suggestion that `LogoList` be replaced with `Vector` rather than being a wrapper around it?

SethTisue:
> Is the suggestion that LogoList be replaced with Vector rather than being a wrapper around it?

Well, I don't remember anything about what the wrapper was for. But offhand, I wouldn't think you'd need to change anything, I would think you'd get the benefits simply by upgrading to 2.13.

TheBizzle:
Thanks, @SethTisue!  Scala 3 is definitely on our radar, but I doubt it will happen for this release.  This was more of a, "We've been on Scala 2.12 for as long as I've worked on NetLogo.  It would be great if we could get even just a Scala 2.13 upgrade in, before we have to get this release out."

But getting onto an entirely modern version of Scala was top-of-mind for me when I suggested this project to @ERSUCC.  I'm very grateful that we managed to get this in, on relatively short notice.  As for Scala 3... and 3.6... we'll get there—hopefully sooner, rather than later.

@ERSUCC, please move the remaining tasks from this ticket to individual issues.  Merging will now commence. :drum:

SethTisue:
🎉 🎉 🎉

Issue #2418, opened 2025-05-19
Move color theme selection to "Preferences" dialog


TheBizzle:
It should be a separate tab.  The "Tools" menu item should be removed.

PR #2425, opened 2025-05-23
Preferences dialog redesign

ERSUCC:
This PR splits the Preferences/Settings dialog into multiple tabs to reduce visual clutter and make it easier to find preferences. The Color Themes dialog is now a tab in the preferences dialog, rather than a separate option accessible from the Tools menu. This resolves issue #2418.


TheBizzle:
The inputs in the "Logging" tab are still misaligned for me:

image

TheBizzle (changes requested):
Most of this is great, but appearance (particularly: alignment) should be cleaned up in "Logging" tab.

PR #2456, opened 2025-06-25
I18N GUI Strings: add closing HTML tag

deining:
This PR adds a missing closing `html` tag to an English translation term.


ERSUCC:
@deining Good catch, thank you!

PR #2459, opened 2025-06-27
Errors_en.properties: minor fixes

deining:
This PR corrects minor issues inside `Errors_en.properties`.


ERSUCC:
@deining Looks like you'll need to change the wording of the expected error messages in the tests that reference the changed properties. Let us know if you have any difficulty figuring out what needs to be changed.

deining:
> @deining Looks like you'll need to change the wording of the expected error messages in the tests that reference the changed properties. Let us know if you have any difficulty figuring out what needs to be changed.

Alle checkers are green now :smile:.

Issue #995, opened 2016-03-25
Allow multiple NetLogo windows to be opened from the Dock in Mac OS X


mrerrormessage:
`com.apple.eawt` has added various hooks which make it possible to interact with the dock in Mac OS X. We should integrate NetLogo to be able to open and manage multiple windows from the Dock.

PR #2528, opened 2025-08-24
Improvements to App initialization

ERSUCC:
This PR reworks the initialization of the `App` class to make the code more readable and more robust. These changes also have the benefit of allowing a potential fix for issue #995, which was previously not possible. Below is a list of changes and why I made them:

### Remove all usages of `PicoContainer`

- Pico is a very mystical and complicated way of initializing the app. It is a hinderance to the readability of the code, because it results in constructors either being called implicitly or being called with dummy parameters that get filled in, so it makes it really tough to figure out which objects depend on each other and where things can go wrong.
- It results in more runtime errors during initialization, because it's easy to mix up the implicit initialization order or forget to add something to `Pico`. Also, constructors are often invoked using a `String` name instead of actual code, which can't be caught at compile time. Both of these can be super annoying, because you may not realize that a class isn't being initialized properly until you open a certain dialog or perform a certain action at runtime.
- I've heard that `Pico` has the benefit of reducing dependency requirements, but this really isn't the case. Sure, you don't need to reference a package or a class by name if you're instantiating it implicitly or with a `String`, but the conceptual dependency still exists and the calling code will not work if the dependency changes or gets removed. It's better to be explicit about dependencies in the code so the compiler can do its job correctly.

### Move initialization to the `App` class

- Previously, a lot of initialization happened in `main`, and the corresponding classes were then loaded in from `Pico` inside an initialization method in `App`. This created a high risk of dangling `null` pointers, because all the members of `App` were declared as `var`s with the value `null` and then assigned actual objects later on.
- With the changes in this PR, all initialization now happens inside the `App` class, and members are declared as `val` wherever possible.

### Rework the parsing of command line arguments

- Command line arguments were previously declared as `var`s in the companion object of `App` and initialized to `null`, assigned as necessary during parsing of command line arguments, and then read in from the `App` class during its initialization. Similar to the initialization of `App`, this created a high risk of dangling `null` pointers.
- The parsing process was also not particularly idiomatic or readable, and didn't have consistent error handling.
- With the changes in this PR, command line arguments are added to one class, and the arguments are parsed in a more straightforward and readable manner. Also, all relevant errors are now caught and throw custom error messages that are internationalized.


TheBizzle:
Man, I hated this thing.  I mean... I've had to use Spring before, so I know that there's worse, but... I'm fairly sure that Bob Dylan wrote the final section of "Masters of War" about PicoContainer:

> I think you will find
> When your death takes its toll
> All the money you made
> Will never buy back your soul
>
> And I hope that you die
> And your death will come soon
> I'll follow your casket
> On a pale afternoon
>
> I'll watch while you're lowered
> Down to your deathbed
> And I'll stand over your grave
> Till I'm sure that you're dead

*Pretty* sure those are the words of a PicoContainer ~~user~~ survivor.

PR #2544, opened 2025-09-09
Update code and GitHub actions workflows for Java 21

ERSUCC:
This pull request updates the repo to work with Java 21. It also updates and unifies the GitHub actions workflow files for the main repo and each submodule.

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -46,25 +46,25 @@ jobs:
       fail-fast: false
       matrix:
         test: [
-          "parserJS/test; parserJVM/test",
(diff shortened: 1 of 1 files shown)


SethTisue:
fwiw, JDK 25 went GA a few days ago

ERSUCC:
That's exciting news! Unfortunately, we can't switch to Java 25 yet, because 32-bit architectures are no longer supported. I don't think we're quite ready to stop providing 32-bit versions of NetLogo.

SethTisue:
oh, interesting, did not know that!

PR #2555, opened 2025-09-16
Start watching files on LoadEndEvent

kgmt0:
Previously, we start watching files on LoadModelEvent. That seemed to be too early, resulting in NLS files sometimes being missing from the list of monitored files.


TheBizzle:
Rebased and merged as commit 197ab6c4585806ae6e4c1ca6074ec709e20df8a5.  Thank you! :bowing_man:

PR #2557, opened 2025-09-17
Fix duplicate dirty file dialog when auto-reloading

kgmt0:
The dialog can be displayed multiple times when there are multiple file events happening in a row. Fix this by stopping event processing as soon as a dirty file is detected.


TheBizzle:
Rebased and merged as commit 98018fbeff96605ca2ff83f0dfddd334d9a9aedd.  Thank you! :bowing_man:

PR #2565, opened 2025-10-07
Run BehaviorSpace experiments in a separate process

### LINKED ISSUE #259: Can't minimize BehaviorSpace window while running (Windows 7)
When I click on the '_' (minimize) button for NetLogo I just hear a beep and nothing happens. Presumably because the dialog "Running Experiment: ..." is open. I can't move the window either (only the experiment dialog is moveable). But I can put other windows on top.

5.0.3
Windows 7

### LINKED ISSUE #2412: Huge memory leak when using parallel runs in BehaviorSpace
While investigating an unrelated issue, I discovered that the workspaces created for parallel runs in BehaviorSpace are never garbage collected. This occurs both in the GUI version of BehaviorSpace and the (in progress) BehaviorSpace extension in NetLogo 7. I also confirmed that the issue occurs in NetLogo 6.4. Here is one way to reproduce the issue from the BehaviorSpace GUI:

1. Open your system profiler, eg. Task Manager, Activity Monitor, etc.
2. Open NetLogo 7 and ensure that it is visible in your profiler
3. Open a model with lots of patches in it, for example Fire
4. Create a new experiment
5. Set the repetitions to 100
6. Set the time limit to 1
7. Run the experiment and set the threads to 100
8. Repeat step 7 until you begin to fear for your computer's life

Note that the memory usage compounds the more times you repeat step 7, and it does not decrease when you load a new model or switch to another existing model. After discovering this bug, I used JProfiler and was able to confirm that the issue is indeed the headless workspaces. For larger models, this could be a fatal flaw, and it should definitely be fixed before the next release if we plan to advertise GUI changes to BehaviorSpace and/or the BehaviorSpace extension.

ERSUCC:
This PR enables BehaviorSpace experiments to be run in a separate process, which fixes the GC problems detailed in issue #2412. It also adds an option to the Run Options dialog that allows the user to specify a memory limit for the experiment process.

Fixes #2412
Fixes #259


SethTisue:
very cool

PR #2605, opened 2025-12-19
Get rid of manual preview images

ERSUCC:
Currently, almost 100 models in the library have manually generated preview images. This makes the release process difficult, leading to out-of-date preview images, and it also prevents automated testing of these models. This PR aims to resolve that issue, replacing as many manual preview images as possible with automated preview commands.


SethTisue:
I always felt a little guilty about leaving this mess behind, thanks for attempting to clean it up 👏

PR #2624, opened 2026-01-31
Add support for custom color themes

### LINKED ISSUE #374: Widgets Colors Should Be Customizable
Luis F. Schettino writes:

> I have been a user of NetLogo for many years and have enjoyed (as have my students, who use the simulations I make for them) the simplicity and clarity of the NetLogo interface. However, I find that as the number of controls on the interface increases, just the sight of 20 or 30 elements among monitors, sliders, buttons and plots, tends to discourage exploration rather than encourage it. I was wondering if you think it would be useful to apply a bit of human factors analysis to the interface. The basic suggestion is to add the option of choosing the color and perhaps even the shape of buttons and sliders. It is much easier for the user to map a color-coded set of controls to their functions.

Being able to choose colors for widgets sound like a good idea for me.

### LINKED ISSUE #1601: Allow turning off current-line highlight
The bar indicating the current line of the source code in recent version of NetLogo seems like it's probably a nice feature for some users, but it's annoying to me.  I keep thinking that I have highlighted the text, despite the fact that actual highlighting is a different color.  I would be happy with a simple blinking block cursor, or any kind of traditional cursor.  A bar across the text at the current line seems pretty non-standard--most programmer's editors don't traditionally do this, and NetLogo didn't use to do this.  (For all I know the bar is the latest thing, with various new editors using it, but it's still not traditional.  Anyone who's been coding for a while will have years of no-bar intuitions built up.)

How about a simple toggle setting to turn it off?  (Or if it there is sufficient benefit to allowing users to change the color (cf. https://github.com/NetLogo/NetLogo/pull/1442/commits/df93e60dd9825c21b2a4c20162c75ab755f16665), one could turn it off by setting the bar to the background color.)

I understand that the NetLogo editor doesn't, and shouldn't, try to do everything and please everyone, and that it may have special orientation to new users, but still, this is a pretty odd feature compared to what many of us are used to.

### LINKED ISSUE #2389: Give NetLogo code colors that are less dreary
People say that NetLogo looks ugly and outdated.  Well, on top of the many things that we've been doing to make it look better, we should probably rethink the colors of text in the Code tab.  Our colors are sad.  Our dark-red/crimson/brown thing is bleh.  Our green is 🤮 .  We can do better.

Here's a picture of something that I (an aesthetically-blind person) came up with to replace it:

Image

The RGB values for these changes are:

  * Literal: 6, 161, 241
  * Reporter: 111,  51, 136
  * Command: 13,  71, 224
  * Keyword:   7, 144,  15
  * Comment: 120, 120, 120

These are not the ideal colors.  But I find them so much less dreary than we current have.  Others will disagree with me.  That's fine.  But some design work should be done on this, and we should come up with a better theme.

ERSUCC:
This PR adds a theme manager and a custom theme editor, allowing users to create variants of the default themes we provide. The theme editor allows the modification of all properties in `InterfaceColors`.

Fixes #374
Fixes #2389
Fixes #1601


ERSUCC:
Due to lack of interest, I'm closing this PR for now. If we ever decide to come back to it there will be so many merge conflicts that I might as well just start over instead of fixing this PR.

PR #2629, opened 2026-02-08
Implement remote commands

kgmt0:
When enabled, this creates a Unix socket (or a named pipe on Windows) that listens to commands. An external editor can use this socket to send commands for NetLogo to run in the form of JSON objects that look like this: `{"type": "nl-run-code", "code": "show 123"}`.

This still needs to be tested on Windows and OSX. I've only tested it on Linux. On Linux (and probably OSX too), an easy way to test this is to use OpenBSD Netcat (GNU Netcat doesn't support Unix sockets). You can open a connection like this `nc -U /path/to/socket/netlogo-1234` and enter the JSON Object you wan to send. On Linux and OSX, the socket will be put in `XDG_RUNTIME_DIR` if it is set, otherwise it will be put in  `~/.netlogo/7.0`. You can also see the exact path in the console output.

### DIFF
--- a/build.sbt
+++ b/build.sbt
@@ -264,6 +264,7 @@ lazy val netlogo = project.in(file("netlogo-gui")).
       "org.scala-lang" %% "scala3-compiler" % "3.7.0",
       "it.unimi.dsi" % "fastutil" % "8.5.16",
       "net.java.dev.jna" % "jna-platform" % "5.17.0",
+      "com.kohlschutter.junixsocket" % "junixsocket-core" % "2.10.1",
       "com.thesamet.scalapb" %% "scalapb-runtime" % scalapb.compiler.Version.scalapbVersion % "protobuf",
       "com.softwaremill.sttp.client4" %% "core" % "4.0.15",
       "com.softwaremill.sttp.client4" %% "upickle" % "4.0.15",
--- a/netlogo-gui/resources/i18n/GUI_Strings_en.properties
+++ b/netlogo-gui/resources/i18n/GUI_Strings_en.properties
@@ -658,6 +658,7 @@ tools.preferences.uiLanguage = User Interface language
 tools.preferences.detectLanguage = <detect language>
 tools.preferences.loadLastOnStartup = Reload the last used model on startup
 tools.preferences.reloadOnExternalChanges = Reload model on external changes
+tools.preferences.enableRemoteCommands = Enable remote commands
 tools.preferences.boldWidgetText = Bold widget text
 tools.preferences.jumpOnClick = <html>Jump sliders to click location<br>(instead of incrementing)</html>
 tools.preferences.sendAnalytics = Send anonymous usage statistics
--- a/netlogo-gui/src/main/app/TabManager.scala
+++ b/netlogo-gui/src/main/app/TabManager.scala
@@ -337,10 +337,23 @@ class TabManager(val workspace: GUIWorkspace, val interfaceTab: InterfaceTab,
     getExternalFileTabs.foreach(_.setIncludedFilesShown(visible))
   }
 
+  def remoteCommandsEnabled: Boolean = interfaceTab.commandCenter.commandServer.running
+  def remoteCommandsEnabled_=(b: Boolean): Unit = {
+    if (b) {
+      interfaceTab.commandCenter.commandServer.start()
+    } else {
+      interfaceTab.commandCenter.commandServer.stop()
+    }
+  }
+
+  remoteCommandsEnabled = getRemoteCommands
+
   def watchingFiles: Boolean = watcherThread != null
   def watchingFiles_=(value: Boolean): Unit = setWatchingFiles(value)
 
   watchingFiles = getAutoReload
+
+  def getRemoteCommands: Boolean = NetLogoPreferences.get("enableRemoteCommands", "false").toBoolean
 
   def getAutoReload: Boolean = NetLogoPreferences.get("reloadOnExternalChanges", "false").toBoolean
 
--- a/netlogo-gui/src/main/app/common/CommandLine.scala
+++ b/netlogo-gui/src/main/app/common/CommandLine.scala
@@ -166,17 +166,22 @@ class CommandLine(commandCenter: CommandCenterInterface,
     executeCurrentBuffer()
   }
 
-  private def executeCurrentBuffer(): Unit = {
-    var inner = getText
-    if (inner.trim.equals("")) {
-      setText("")
-      return
+  private def preprocess(code: String): String = {
+    if (code.trim.equals("")) {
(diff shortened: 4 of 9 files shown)


ERSUCC:
This is a very cool idea! Other than a few unnecessary semicolons, the code looks fine to me, and your preference implementation follows the standards correctly. However, I did notice quite a few issues while testing on Windows, which I have detailed below:

1. You added the `CommandServer` instance as a GUI component at the end the command line, which is very ugly (at least in dark mode), and I don't think it needs to exist. If there's some reason why there needs to be a GUI component for it, then it should be hidden unless the preference is enabled. As far as I can tell, you can simply remove line 128 of `CommandCenter.scala` to solve the problem.

2. Disabling the preference or launching NetLogo with the preference already enabled results in an `IOException` with the following stack trace:
```
java.io.IOException: GetOverlappedResult() failed for connect operation: 6
 at org.scalasbt.ipcsocket.Win32NamedPipeServerSocket.accept(Win32NamedPipeServerSocket.java:186)
 at org.nlogo.app.common.CommandServerSocket.accept(CommandServer.scala:125)
 at org.nlogo.app.common.CommandThread.run(CommandServer.scala:59)
```
3. The name of the chosen pipe is printed to standard output, which won't be seen for anyone using the actual release. This should either be indicated in an extra line under the preference when it's enabled (like when "\<detect language\>" is chosen), or in a dialog that appears after enabling the preference or launching the app.

4. The Windows pipe created by the library you're using is opened in byte mode rather than message mode, despite the fact that your code is clearly intending to send messages. This is slightly less convenient for the client, but not the end of the world if it can't be configured to open in message mode instead.

Ignoring these issues, I was in fact able to connect to the pipe and execute a command in NetLogo, as well as receive the status response. However, if I try to execute two commands in a row, it gives me the following exception:
```
java.io.UncheckedIOException: java.io.IOException: ReadFile() failed: 109
 at java.base/java.io.BufferedReader$1.hasNext(BufferedReader.java:574)
 at java.base/java.util.Spliterators$IteratorSpliterator.tryAdvance(Spliterators.java:1855)
 at java.base/java.util.Spliterators$1Adapter.hasNext(Spliterators.java:681)
 at scala.collection.convert.JavaCollectionWrappers$JIteratorWrapper.hasNext(JavaCollectionWrappers.scala:46)
 at scala.collection.IterableOnceOps.foreach(IterableOnce.scala:619)
 at scala.collection.IterableOnceOps.foreach$(IterableOnce.scala:617)
 at scala.collection.AbstractIterator.foreach(Iterator.scala:1306)
 at org.nlogo.app.common.CommandThread.processRequests(CommandServer.scala:82)
 at org.nlogo.app.common.CommandThread.run(CommandServer.scala:64)
```
Disabling and re-enabling the preference allows me to send another command on the same pipe, even after receiving this error. This very well could be user error, so let me know what you find.

Everything else I tried worked great! I was able to execute a variety of standard primitives, with no functional issues that I could notice.

Hopefully this is a helpful report. Let me know I need to clarify anything, or if you would like to see the code I used to connect to the pipe.

P.S.: In retrospect, although I find this extremely cool and potentially very useful to me for my side projects, it's not clear to me how the average NetLogo user would be intended to access this functionality. I've never encountered any mention of named pipes outside of the Win32 API docs. Maybe I misunderstood the intended purpose of this remote control feature?

ERSUCC:
Review part 2:

I have now tested on Mac as well, with much better results. When disabling the preference, I get the following exception:
```
java.io.IOException: org.scalasbt.ipcsocket.NativeErrorException: [53] Software caused connection abort
 at org.scalasbt.ipcsocket.UnixDomainServerSocket.accept(UnixDomainServerSocket.java:160)
 at org.nlogo.app.common.CommandServerSocket.accept(CommandServer.scala:125)
 at org.nlogo.app.common.CommandThread.run(CommandServer.scala:59)
```
But aside from that and the GUI issue I mentioned, everything seems to be working smoothly. I was able to use `nc` as indicated, and it worked quite well.

It might be worth noting that on Mac, the socket does not actually get created at `~/.netlogo/7.0` as you indicated. But there are no functional problems with where it does end up, so I doubt you'll need to do anything to change that.

kgmt0:
Thanks for testing! That's a lot more breakage than I expected. I might have to spin up a VM myself to debug.

kgmt0:
> P.S.: In retrospect, although I find this extremely cool and potentially very useful to me for my side projects, it's not clear to me how the average NetLogo user would be intended to access this functionality. I've never encountered any mention of named pipes outside of the Win32 API docs. Maybe I misunderstood the intended purpose of this remote control feature?

The idea is that this will be used by an editor plugin so that you can do things like running `setup` and `go` without leaving your editor. The plugin would be responsible for discovering these sockets and sending the commands on behalf of the user. In the case that there are multiple sockets, it could use some heuristics to decide which socket to use or just directly ask the user (preferably presenting them as something more user-friendly than just the socket paths). Ideally, the user wouldn't need to know that these sockets exist at all.

ERSUCC:
@kgmt0 Unfortunately, this new library is even worse on Windows than the last one. I don't get any exception dialogs when toggling the preference or starting up NetLogo, but I am no longer able to connect to the pipe. If I use your solution as is, it says it can't access the file, presumably because pipes are supposed to be created at `\\.\pipe\...` instead of in an arbitrary location. But if I change the code to create the pipe in the correct location, it fails to create one at all. Am I supposed to use a different method of connecting to the socket with this new library?

kgmt0:
> Am I supposed to use a different method of connecting to the socket with this new library?

Yes, it turns out [Windows supports Unix sockets now](https://devblogs.microsoft.com/commandline/af_unix-comes-to-windows/), so I just used Unix sockets on all OSes. But since there's no longer any need for platform-specific code, I'm thinking I might as well just use the Java standard library and create the socket myself. I'll update the PR if that goes well.

ERSUCC:
Ah ok, fantastic! Thanks for the clarification.

kgmt0:
It looks like switching to just using the Java standard library would be more work than I thought. For some reason, Unix sockets only work with `ServerSocketChannel`, and the API is quite a bit lower level than `ServerSocket` that I'm currently using, so I think it's worth using a library here.

TheBizzle (approved):
@kgmt0, if you rebase this onto the latest `wip-7.1`, it should be good to merge.

kgmt0:
`test-extensions` is still failing, but I don't think it's related to this PR.

TheBizzle:
We shouldn't merge stuff before it gets a :heavy_check_mark: from the CI.  But my mistake for the miscommunication.  I've let @ERSUCC know to fix the things that are causing tests to fail on `wip-7.1`, so we can see a :heavy_check_mark: .

PR #2633, opened 2026-02-16
Surround code and info sections of nlogox files with newlines

kgmt0:
This ensures that the first and the last line of the code and info section are on their own lines, which makes it easier to edit with an external editor. When reading the file, the surrounding newlines are stripped. If the newlines are missing, the file still parses as before.

As an example, this is what the code section currently looks like. Note how the XML tag shifts the first line to the right and breaks the indentation of the NetLogo code. Because the XML tags are on the same line as the code, common navigation keys like `Home` and `End` also become much less useful.
```xml
  <code><![CDATA[to test-code
  print "hello from the code"
end]]></code>
```

With this change, the same code section looks a lot closer to how it does in NetLogo code editor.
```xml
  <code><![CDATA[
to test-code
  print "hello from the code"
end
]]></code>
```
`RoundTripTests` are currently failing because the models need to be resaved.


ERSUCC:
Unless I'm misunderstanding something, this will remove intentionally added newlines when a 7.0 model is opened for the first time in 7.1. It might be worth checking the `version` attribute of the `model` tag before applying the new formatting behavior.

kgmt0:
> Unless I'm misunderstanding something, this will remove intentionally added newlines when a 7.0 model is opened for the first time in 7.1. It might be worth checking the `version` attribute of the `model` tag before applying the new formatting behavior.

Good point. I added a check for that, but now you'll have to either modify my patch or bump the version number to test.

TheBizzle (changes requested):
Make a branch of the `models` repo and then open SBT in this repo and run `resaveModels` to apply this change to all of the models.  Then, it should be good to merge.

PR #2644, opened 2026-03-06
Allow full keyboard operation of the GUI

### LINKED ISSUE #2531: Buttons do not indicate focus for keyboard access
## Steps to Reproduce

- Open the Models Library, the search input box should be highlighted/blinking cursor
- Hit tab to move focus to the next control, observe the Clear button is not highlighted
- Continuing hitting tab; the tree view and text input seem to give an indication they are focused, the buttons do not.
- Note that the buttons are gaining focus, as hitting enter after tabbing will cause them to activate, it's just not visually apparent.

## What Happens in 6.4

In NetLogo 6.4 focused button elements do gain a highlight indicator/border.

## Notes

While the steps to reproduce reference the Models Library, it appears to affect all buttons, as on the model save/discard dialog, shape editors, etc.  Tested on Ubuntu 24, 64-bit, light or dark mode.

### LINKED ISSUE #2575: NetLogo cannot be reliably navigated without a mouse
In the interest of making NetLogo more accessible to more users, it would be a huge bonus if the application could be more reliably navigated with a keyboard or dictation software instead of a mouse. Currently, most interface items are either not focusable or do not indicate when they are focused, making NetLogo pretty much unusable for users who cannot or prefer not to use a mouse.

ERSUCC:
This PR adds the necessary functionality to allow full keyboard operation of the GUI for focus traversal and basic model authoring.

Fixes #2575
Fixes #2531


alan-isaac:
This seems great but I cannot tell: will it now be possible to use the keyboard to switch focus from the command line to the widget window, thereby gaining access to the keyboard shortcuts (e.g., of buttons)? 🤞

ERSUCC:
@alan-isaac Yes, every GUI element, including the widget panel and the widgets inside of it, is accessible by keyboard! There isn't a direct shortcut to get to the widget panel from the Command Center, but you can get there by cycling the focus with the Tab key.

alan-isaac:
This is all great, but I suspect the single most needed keyboard action is to toggle between the command line and the widget window. Can a keyboard toggle be added?

PR #2663, opened 2026-04-30
Fix wip 7.1 test failures

kgmt0:
This fix test failures that occurred following the merge of https://github.com/NetLogo/NetLogo/pull/2633.


kgmt0:
The models still need to be resaved, but we need to bump the version number before doing that. 85178a5e7b728b60044a134761374ce3e1a1435e and dffed1bd9f1514ead0fa8a1aca1f6e606bdd469b only add/strip newlines if the version number is 7.1.0 or higher, so just resaving the models now will break some tests.

TheBizzle (changes requested):
This branch will not be merged before 7.1.0 is on the table, anyway. I'd rather not have all of this version-checking code for a situation that we won't ever find ourselves in. Instead, you should increase the version in [`version.txt`](https://github.com/NetLogo/NetLogo/blob/wip-7.1/shared/resources/main/version.txt) to `7.1.0-beta1` and then run `resaveModels` again so they all get saved with the right version and spacing.

kgmt0:
PR for resaved models: https://github.com/NetLogo/models/pull/449

kgmt0:
I think there's something weird going on with GitHub Actions. 68f1a2e shouldn't be able to pass the tests until https://github.com/NetLogo/models/pull/449 is merged, but somehow [it did](https://github.com/NetLogo/NetLogo/actions/runs/25194705744/job/73928780335). Now that it's actually merged, it's back to failing the tests again.

ERSUCC:
@kgmt0 If you look at the tests that are currently failing on `wip-7.1`, they don't seem to relate at all to the models repo, so I think they are unrelated. Also, I don't see why the commit you referenced should have failed the tests, because even though you didn't merge the PR in the models repo, you still pointed this PR to the correct commit in the models repo.

kgmt0:
> If you look at the tests that are currently failing on wip-7.1, they don't seem to relate at all to the models repo, so I think they are unrelated.

That's even more strange. How did the same failures not happen for https://github.com/NetLogo/NetLogo/commit/68f1a2e69fe70cd14e1accb93f97f07826e5cb6b? The exact same [commit](https://github.com/kgmt0/NetLogo/commit/68f1a2e69fe70cd14e1accb93f97f07826e5cb6b) in my fork has [failing tests](https://github.com/kgmt0/NetLogo/actions/runs/25194704998/job/73872580202).

> Also, I don't see why the commit you referenced should have failed the tests, because even though you didn't merge the PR in the models repo, you still pointed this PR to the correct commit in the models repo.

The commit only exists in my own fork right now. I haven't pushed anything to official models repo yet.

ERSUCC:
Hmm, that does sound wrong. GitHub has apparently been having database issues recently, so maybe that's the explanation?

PR #2676, opened 2026-06-15
Infrastructure: Simplify InputBox logic with BoxedValues instead of text + value

AwesomeDudeDK:
Previously there were many unnecessary comparisons with variables text and value when the logic could be handled with BoxedValues. This simplifies the code and reduces the amount of comparisons and work we have to do every time an input box is updated. This PR also fixes the logic of what happens to the existing text when we change the type of an input box. If the text is reinterpretable, we keep it as is, otherwise reset it to the default value.

### DIFF
--- a/netlogo-gui/src/main/window/InputBox.scala
+++ b/netlogo-gui/src/main/window/InputBox.scala
@@ -5,7 +5,6 @@ package org.nlogo.window
 import java.awt.{ Color, Component, Dimension, Font, Frame, Graphics, GridBagConstraints, GridBagLayout, Insets,
                   LinearGradientPaint }
 import java.awt.event.{ ActionEvent, ActionListener, FocusEvent, FocusListener, KeyEvent, WindowAdapter, WindowEvent }
-import java.lang.{ Double => JDouble }
 
 import javax.swing.{ AbstractAction, JButton, JDialog, JLabel, JPanel, ScrollPaneConstants, UIManager }
 import javax.swing.KeyStroke.getKeyStroke
@@ -16,7 +15,7 @@ import org.nlogo.api.Approximate.approximate
 import org.nlogo.api.Color.{ getColor, getColorNameByIndex, modulateDouble }
 import org.nlogo.agent.InputBoxConstraint
 import org.nlogo.awt.{ Hierarchy, Positioning }
-import org.nlogo.core.{ BoxedValue, CompilerException, I18N, InputBox => CoreInputBox, LogoList, NumericInput,
+import org.nlogo.core.{ BoxedValue, CompilerException, I18N, InputBox => CoreInputBox, NumericInput,
                         StringInput, Widget => CoreWidget }
 import org.nlogo.editor.{ AbstractEditorArea, EditorConfiguration }
 import org.nlogo.swing.{ Button, ButtonPanel, DialogButton, OptionPane, RoundedBorderPanel,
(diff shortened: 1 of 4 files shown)


netlogo-gui/src/main/render/TorusRenderer.java
```diff
@@ -206,42 +206,41 @@ public void drawLink(GraphicsInterface g, org.nlogo.api.Link link, LinkDrawer.Li
...
-    double y2 = graphicsYNoOffset(endY, world.fixedPatchSize());
+    double x1 = graphicsXNoOffset(startX, world.patchSize());
+    double y1 = graphicsYNoOffset(startY, world.patchSize());
+    double x2 = graphicsXNoOffset(endX, world.patchSize());
+    double y2 = graphicsYNoOffset(endY, world.patchSize());
 
     double size = penSize / 2;
 
     g.drawLine(x1, y1, x2, y2);
-
```
ERSUCC, line 217: This removal seems unrelated and just makes the diff longer. Probably best to not include it in the PR.

netlogo-gui/src/main/window/InputBox.scala
```diff
@@ -73,7 +72,7 @@ abstract class InputBox(textArea: AbstractEditorArea, editDialogTextArea: Abstra
 
     def initGUI(): Unit = {
       removeAll()
-
+
```
ERSUCC: The standard for this repo is to remove all trailing whitespace at the end of lines and at the end of files. You can configure VS Code to do this when you save a file; go to Settings and search for "trim trailing whitespace".

netlogo-gui/src/main/window/InputBox.scala
```diff
@@ -130,20 +129,19 @@ abstract class InputBox(textArea: AbstractEditorArea, editDialogTextArea: Abstra
...
-  protected var value: Option[AnyRef] = None
-  def valueText = text
-  def valueObject() = value.orNull
-  def valueObject(value: AnyRef): Unit = {valueObject(value, false)}
-  def valueObject(value: Any, raiseEvent: Boolean): Unit = {
-    oldText = text
-    text = Dump.logoObject(toAnyRef(value))
-    this.value = Option(toAnyRef(value))
-    if (text != textArea.getText) textArea.setText(text)
+  // values are now stored as boxedValues
```
ERSUCC: This comment doesn't really tell the reader anything, since it is very clear in the following lines that type is `BoxedValue`. I think it's fine to just remove the old comment and not add anything in its place. Also, when you make comments, sign it with your name so it's clear who made the comment originally if it gets moved around by someone else in the future.

netlogo-gui/src/main/window/InputBox.scala
```diff
@@ -130,20 +129,19 @@ abstract class InputBox(textArea: AbstractEditorArea, editDialogTextArea: Abstra
...
+
+  protected var oldBoxedValue: BoxedValue = StringInput("", StringInput.StringLabel, false)
```
ERSUCC: Define this after `boxedValue` and set it to `boxedValue` so that you don't need to type the same thing twice.

netlogo-gui/src/main/window/InputBox.scala
```diff
@@ -155,22 +153,25 @@ abstract class InputBox(textArea: AbstractEditorArea, editDialogTextArea: Abstra
     }
   }
 
-  protected def inputText(input: Object): Unit = {
-    if (input != null) valueObject(input, true)
+  protected def getInput(bv: BoxedValue): Unit = {
```
ERSUCC: Use `Option[BoxedValue]` instead of `BoxedValue`, then do `bv.foreach` instead of checking if it's `null`. You can create `Option`s when you call `getInput` by wrapping the existing input with `Option(...)`

netlogo-gui/src/main/window/InputBox.scala
```diff
@@ -691,31 +667,34 @@ abstract class InputBox(textArea: AbstractEditorArea, editDialogTextArea: Abstra
...
+    override def defaultValue = StringInput("0", StringInput.ReporterLabel, false)
     override def enableBracketMatcher = true
     @throws(classOf[ValueConstraint.Violation])
     @throws(classOf[CompilerException])
     override def readValue(text: String) = {
-      constraint.assertConstraint(text)
-      Exceptions.ignoring(classOf[CompilerException]) {
-        compiler.checkReporterSyntax(text)
+      if (text.isEmpty) {
+        StringInput("0", StringInput.ReporterLabel, this.multiline)
```
ERSUCC, line 681: You don't need `this.` unless there's a conflicting local variable.

netlogo-gui/src/main/window/InputBoxWidget.scala
```diff
@@ -36,15 +36,35 @@ class InputBoxWidget(textArea: AbstractEditorArea, dialogTextArea: AbstractEdito
     if (!editing) new InterfaceGlobalEvent(this, false, true, false, false).raise(this)
   }
 
-  override def valueObject(value: Any, raiseEvent: Boolean): Unit = {
-    if (! this.value.contains(toAnyRef(value))) {
-      oldText = text
-      text = Dump.logoObject(toAnyRef(value))
-      this.value = Option(toAnyRef(value))
-      if (!text.equals(textArea.getText())) textArea.setText(text)
+  override def valueObject(value: AnyRef): Unit = value match {
```
ERSUCC: I feel like this method could be a lot cleaner if you used `InputType.readValue`, which you already have set up to check all this stuff.

netlogo-gui/src/main/window/InputBoxWidget.scala
```diff
@@ -36,15 +36,35 @@ class InputBoxWidget(textArea: AbstractEditorArea, dialogTextArea: AbstractEdito
...
+  }
+
+  override def valueObject(value: BoxedValue, raiseEvent: Boolean): Unit = {
+    val valueChanged = boxedValue != value
+    if (valueChanged) {
+      oldBoxedValue = boxedValue
+      boxedValue = value
+    }
+    if (boxedValue.defaultString != textArea.getText) textArea.setText(boxedValue.defaultString)
+    if (valueChanged) {
```
ERSUCC: Don't check the same condition twice if it doesn't change in between the checks.

parser-core/src/main/parse/StructureChecker.scala
```diff
@@ -20,16 +20,10 @@ import SymbolType._
...
-  }
-
   def rejectMisplacedConstants(declarations: Seq[Declaration]): Unit = {
     for (declaration <- declarations) {
       declaration match {
-        case Variables(_, names, _, _) =>
+        case Variables(_, names) =>
           for (name <- names) {
             if (name.token.tpe == TokenType.Literal) {
               if (!Constants.get(name.token.text).isEmpty) {
```
ERSUCC, line 35: Isn't this change from a commit a while back? Why is it included in this PR? Make sure you've properly rebased with `bugfix-buffer` and that there aren't any extra changes.

parser-core/src/test/parse/StructureParserTests.scala
```diff
@@ -169,10 +169,10 @@ class StructureParserTests extends AnyFunSuite {
       "Variable name conflicts with a constant.") }
   test("string in globals") {
```
ERSUCC, line 170: Same here, see previous comment.

shared/resources/main/i18n/Errors_en.properties
```diff
@@ -204,8 +204,7 @@ compiler.MultiAssign.tooFewValues = The list of values for {0} must be at least
 compiler.Backifier.replaced = {0} is no longer a primitive, use {1} instead.
 compiler.StructureParser.includeNotFound = Could not find {0}
 compiler.StructureConverter.noBreed = There is no breed "{0}"
-compiler.StructureChecker.variableLiteral = A string/number cannot be used as a global variable name.
-compiler.StructureChecker.declOrder = All declarations must come before procedures.
+compiler.StructureChecker.variableLiteral = You cannot use a string/number as a global variable name.
```
ERSUCC: Also this, see previous two comments.

ERSUCC (changes requested):

PR #2681, opened 2026-06-28
Fix parsing for imports with identifier renaming

kgmt0:
The parser was accepting any identifier instead of just `as` for renaming. For example, `import [foo bar baz] from qaz` would import `foo` from `qaz` renaming it to `baz`. 5783090 fixes that, making it import `foo`, `bar`, and `baz` as it's supposed to.

a35ecd9 makes it so that parens are required around renames, e.g. `import [(foo as bar)] from qaz`. This is to eliminate ambiguity when parsing something like `import [foo as bar] from qaz`. In that case, it isn't clear if it's supposed to just import `foo` renaming it to `bar`, or import `foo`, `as`, and `bar`.

0b9e54c has test cases for imports with identifier lists and renames.

### DIFF
--- a/parser-core/src/main/parse/StructureCombinators.scala
+++ b/parser-core/src/main/parse/StructureCombinators.scala
@@ -117,9 +117,9 @@ extends scala.util.parsing.combinator.Parsers {
     }
 
   def renamableIdentifier: Parser[(String, String)] =
-    identifier ~ opt(exactIdentifier("AS") ~ identifier) ^^ {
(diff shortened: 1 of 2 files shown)


TheBizzle:
GitHub seems to be picking up the GHA config without the recent fix, even though the fix is on your branch.  The tests for this branch's code passed [here](https://github.com/NetLogo/NetLogo/actions/runs/28834178998).

PR #2683, opened 2026-07-01
Fix stack overflow when a module imports itself

kgmt0:
`findAllImportedFiles()` rescursively finds imports on all imported files, so a module that imports itself will cause an infinite recursion. Fix this by keeping track of files that have already been seen in a Set and not recursing if the current file is in that Set.


TheBizzle:
GitHub seems to be picking up the GHA config without the recent fix, even though the fix is on your branch.  The tests for this branch's code passed [here](https://github.com/NetLogo/NetLogo/actions/runs/28834188997).

PR #2684, opened 2026-07-01
Fix ignored imports

kgmt0:
The list of imports were not being preserved across iterations, causing some import statements to be ignored.


TheBizzle:
GitHub seems to be picking up the GHA config without the recent fix, even though the fix is on your branch.  The tests for this branch's code passed [here](https://github.com/NetLogo/NetLogo/actions/runs/28834168306).

PR #2686, opened 2026-07-08
Display an error when there are duplicates in the import list

kgmt0:
This adjusts the parser for the second form of the import syntax so that listing duplicate identifiers will result in an error, e.g. `import [foo foo] from baz`.

### DIFF
--- a/parser-core/src/main/parse/StructureCombinators.scala
+++ b/parser-core/src/main/parse/StructureCombinators.scala
@@ -26,7 +26,7 @@ package org.nlogo.parse
 // and discard input).
 
 import
-  org.nlogo.core.{ Token, TokenType, StructureDeclarations },
+  org.nlogo.core.{ I18N, Token, TokenType, StructureDeclarations },
   StructureDeclarations._
 
 object StructureCombinators {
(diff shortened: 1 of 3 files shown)


parser-core/src/main/parse/StructureCombinators.scala
```diff
@@ -122,11 +122,23 @@ extends scala.util.parsing.combinator.Parsers {
...
 
-  def importedIdentifierList: Parser[Map[String, String]] =
-    openBracket ~> commit(rep(renamableIdentifier) <~ closeBracket) ^^ {
-      case x => x.toMap
+  def importedIdentifierList: Parser[Map[String, String]] = {
+    val parser = openBracket ~> commit(rep(renamableIdentifier) <~ closeBracket) ^^ {
+      case xs => {
+        val originalLength = xs.length
+
+        // Check if there are any duplicate keys/values in the rename map.
```
TheBizzle: Please sign and date the comment or remove it (and make the code clearer, if you feel it warrants a comment).

TheBizzle (changes requested):

parser-core/src/main/parse/StructureCombinators.scala
kgmt0: I rewrote this part to use a locally defined function that checks for duplicates. It's a bit more verbose, but I think it's more readable.

PR #2692, opened 2026-07-27
Prevent imported modules from accessing agent variables in the model

kgmt0:
For example, if the model file contains this code:
```
import test

turtles-own [foo]
```

And `test-nls` contains this code:
```
to test
   ask tutles [show foo]
end
```

Without this PR, the example above will compile because `test.nls` has access to `foo` declared in the model file. This patch makes it so that the module no longer has access, so the example would trigger an error.


TheBizzle (changes requested):
This is good, but, by this logic, shouldn't `create-wolves` also be impermissible in `.nls` files?  Currently, it does not produce errors.

kgmt0:
Breed procedures are no longer allowed.

PR #2693, opened 2026-08-07
Change file extension for modules to .nlm

kgmt0:
Tests are failing because models need to updated. See https://github.com/NetLogo/models/pull/454.

### DIFF
--- a/parser-core/src/test/parse/StructureParserTests.scala
+++ b/parser-core/src/test/parse/StructureParserTests.scala
@@ -151,10 +151,10 @@ class StructureParserTests extends AnyFunSuite {
       |end
     """.stripMargin
 
-    expectParseAllError("import [baz baz] from foo", "Identifier list contains duplicates", src)
-    expectParseAllError("import [baz (qaz as baz)] from foo", "Identifier list contains duplicates", src)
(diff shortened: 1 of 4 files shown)


parser-core/src/test/parse/StructureParserTests.scala
```diff
@@ -335,16 +335,18 @@ class StructureParserTests extends AnyFunSuite {
     expectError("breed [ cats cats ]", "A breed cannot have the same plural and singular name")
   }
 
-  def compileAll(src: String, nlsSrc1: String, nlsSrc2: String = ""): StructureResults = {
+  def compileAll(src: String, suffix: String, fooSrc: String, barSrc: String = ""): StructureResults = {
```
TheBizzle: Can you do this in a way that isn't stringly-typed?

Either `...(..., suffix: SourceType, fooSrc: String, ...): ...` with

```scala
enum SourceType {
  case NLInclude, NLModule
}
```

or

`...(..., fooSrc: NLSource, ...): ...` with

```scala
enum NLSource {
  case NLInclude(src: String)
  case NLModule(src: String)
}
```

would be fine.

TheBizzle (changes requested):

PR #2696, opened 2026-08-22
Display an error when encountering scoped identifiers with consecutive colons

kgmt0:
Example: `foo:::bar`. Previously, this was just treated as `foo:bar`. This PR makes it an error instead.


TheBizzle:
Merged in c30984ac1cee3b4ef2b9bdee5cba87eb734dc649.

PR #2698, opened 2026-08-27
Fix incorrect prefix when importing from a directory with a single module

kgmt0:
Previously, subdirectories were being left out of the resulting prefix. For example, suppose there exists a directory containing a single module file at the path `three/four/five.nlm` that exports a single procedure called `hello`. Importing the directory with `import three` would result in `three:hello` becoming visible in the importing file instead of `three:four:five:hello`.


kgmt0:
Tests are passing in my repo https://github.com/kgmt0/NetLogo/actions/runs/33102593731.

TheBizzle (changes requested):
I don't like the way in which this is using regular expressions.  They're mainly being used to strip/replace prefixes/suffixes of strings, for which I feel they are overkill.

I asked GPT to give me more idiomatic Scala code for what you're doing here, and it produced this (with some adjustments from me):

```scala
import java.util.Locale

val basePath = compilationEnvironment.resolvePath(currentImport.filename.getOrElse(""))
val relativePath = currentPath.stripPrefix(s"$basePath$separator")
val path = relativePath.replace(separator, ":").toUpperCase(Locale.ROOT).stripSuffix(".NLM")
```

I haven't verified that this works, but it looks a lot more palatable to me.

TheBizzle (changes requested):
> Importing the directory with `import three` would result in `three:hello` becoming visible in the importing file instead of `three:four:five:hello`.

My test of these commits still shows the same old behavior.  I can also access it from:

* `import [hello] from three:four:five` (correct)
* `import [hello] from three:four` (incorrect)
* `import [hello] from three` (incorrect)

kgmt0:
> My test of these commits still shows the same old behavior. I can also access it from:
> 
>     * `import [hello] from three:four:five` (correct)
> 
>     * `import [hello] from three:four` (incorrect)
> 
>     * `import [hello] from three` (incorrect)

Should be fixed now. That syntax is only supposed to be used when importing directly from a module file, so that's the error message I'm using.

PR #2700, opened 2026-09-09
Remove ImportOption and ImportAlias in StructureDeclarations

kgmt0:
These are no longer used by anything.


TheBizzle:
Merged in 520ea4816583b390fd06704dce70c3d758327566.

PR #2701, opened 2026-09-09
Reject path separators in module paths

kgmt0:
This makes it so that NetLogo displays an error whenever a path separator is found in a module path. Example: "import foo:bar/baz".


TheBizzle:
Merged in d849c6afba65b8c80fc89a3c3acd887f1d14c717.

PR #2702, opened 2026-09-14
Support extensions in modules

kgmt0:
This allows extensions to be used in modules. There is a check to ensure that all extensions used in imported modules are declared in the model code using the `extensions` primitive.


TheBizzle (changes requested):
I don't have a problem with any of what's already here, but I want us to make this better before merging it.

Currently, `.nlm` file editors have no syntax highlighting or code completion suggestions for extension primitives.  This would be more understandable when the extension hasn't been loaded by the main code tab yet, but even in that case, the `.nlm` editor is unaware of extension primitives.

But we should probably solve this for both cases (i.e. already-loaded and not-yet-loaded) in the same way: read the extension's syntax from a JSON file independently of actually loading it.  @ERSUCC, what ever happened to that functionality?

ERSUCC:
@TheBizzle It may now be broken since it's been so long, so I'll have to check on it and do some testing. But that issue aside, it has been completed for many months. You can find the changes [here](https://github.com/NetLogo/NetLogo-Extension-Plugin/tree/prims).

TheBizzle:
@ERSUCC , once you've verified that it's working, publish a new version of the plugin and get all of the bundled extensions using it.

